PR #62, opened 2025-05-19
Update to use published npm and pypy packages.

chendrixson:


jonwis (approved):

samples/python/summarize/requirements.txt
```diff
@@ -1,2 +1,3 @@
 openai>=1.0.0
 python-dotenv>=0.19.0 
+foundry-local-sdk>=0.0.1
```
jambayk, line 3: 0.3.0

jambayk (approved):

samples/python/summarize/requirements.txt
chendrixson: Nothing published less than that version, so this is probably okay.

PR #101, opened 2025-05-22
User/chendrixson/rust library

chendrixson:
First drop of a Rust crate for the foundry local SDK.  Includes mocked out tests, an integration test, and a simple sample.

### DIFF
--- a/samples/rust/hello-foundry-local/src/main.rs
+++ b/samples/rust/hello-foundry-local/src/main.rs
@@ -11,10 +11,12 @@ async fn main() -> Result<()> {
     println!("Hello Foundry Local!");
     println!("===================");
 
-    // Create a FoundryLocalManager instance with default options
-    // This will start the Foundry service if not running
+    // Create a FoundryLocalManager instance using the builder pattern
     println!("\nInitializing Foundry Local manager...");
-    let mut manager = FoundryLocalManager::new(None, None, None).await?;
+    let mut manager = FoundryLocalManager::builder()
+        .bootstrap(true)  // Start the service if not running
+        .build()
+        .await?;
 
     // List all the models in the catalog
     println!("\nAvailable models in catalog:");
--- a/samples/rust/hello-foundry-local/src/main.rs
+++ b/samples/rust/hello-foundry-local/src/main.rs
@@ -10,52 +10,53 @@ async fn main() -> Result<()> {
 
     println!("Hello Foundry Local!");
     println!("===================");
-    
+
     // Create a FoundryLocalManager instance with default options
     // This will start the Foundry service if not running
     println!("\nInitializing Foundry Local manager...");
     let mut manager = FoundryLocalManager::new(None, None, None).await?;
-    
+
     // List all the models in the catalog
     println!("\nAvailable models in catalog:");
     let models = manager.list_catalog_models().await?;
     for model in models {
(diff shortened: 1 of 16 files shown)


jonwis (approved):

samples/rust/hello-foundry-local/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+        env_logger::Env::default().filter_or(env_logger::DEFAULT_FILTER_ENV, "info"),
+    );
+
+    println!("Hello Foundry Local!");
+    println!("===================");
+
+    // Create a FoundryLocalManager instance with default options
+    // This will start the Foundry service if not running
+    println!("\nInitializing Foundry Local manager...");
+    let mut manager = FoundryLocalManager::new(None, None, None).await?;
```
kennykerr: This is a good example where the builder pattern is preferred to avoid `(None, None, None)` which is not very idiomatic. For examples, see how a [Windows service is constructed here](https://crates.io/crates/windows-services) or how a [registry key is constructed here](https://crates.io/crates/windows-registry).

sdk/rust/src/api.rs
```diff
@@ -0,0 +1,503 @@
...
+    ///
+    /// # Returns
+    ///
+    /// A new FoundryLocalManager instance.
+    pub async fn new(
+        alias_or_model_id: Option<&str>,
+        bootstrap: Option<bool>,
+        timeout_secs: Option<u64>,
+    ) -> Result<Self> {
+        assert_foundry_installed()?;
```
kennykerr: An "assert" in Rust generally implies that the call will panic (terminate the process) if the precondition is not met but this function presumably bubbles up the failure in a recoverable way so perhaps `check_foundry_installed` might be more appropriate.

sdk/rust/src/api.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// * `alias_or_model_id` - Optional alias or model ID to download and load. Only used if bootstrap is true.
+    /// * `bootstrap` - Optional boolean to start the service if it is not running. Default is false.
+    /// * `timeout_secs` - Optional timeout for the HTTP client in seconds. Default is 10 seconds.
+    ///
+    /// # Returns
+    ///
+    /// A new FoundryLocalManager instance.
+    pub async fn new(
+        alias_or_model_id: Option<&str>,
+        bootstrap: Option<bool>,
```
kennykerr: You don't seem to distinguish between `false` and `None` so perhaps `bootstrap: bool` will suffice and be simpler.

sdk/rust/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+//!     println!("{}", result["choices"][0]["message"]["content"]);
+//!     
+//!     Ok(())
+//! }
+//! ```
+
+pub mod api;
+pub mod client;
+pub mod models;
+pub mod service;
```
kennykerr: Unless you specifically need these modules to be public, its generally cleaner to keep them private and just expose, via `pub use`, the specific types at crate scope as you already below.

samples/rust/hello-foundry-local/src/main.rs
chendrixson: Updated to use builder, since the common case will probably be emptier construction.

PR #111, opened 2025-05-27
Adding download progress to the C# SDK

aaronpowell:


sdk/cs/src/FoundryManager.cs
```diff
@@ -204,6 +204,123 @@ public async Task<List<ModelInfo>> ListCachedModelsAsync(CancellationToken ct =
         return modelInfo;
     }
 
+    public async IAsyncEnumerable<ModelDownloadProgress> DownloadModelWithProgressAsync(
```
amarin16: there was some recent refactoring that renamed this file to `FoundryLocalManager.cs`, could you please move these changes on top of the latest `main` branch update?

sdk/cs/src/FoundryManager.cs
```diff
@@ -204,6 +204,123 @@ public async Task<List<ModelInfo>> ListCachedModelsAsync(CancellationToken ct =
...
+    string aliasOrModelId,
```
amarin16: Could you also please leverage the `sdk/cs/.editorconfig` that was recently added in order to get consistent formatting? One way to use it is `dotnet format` from powershell

sdk/cs/src/FoundryManager.cs
amarin16: Could you please add a unit test for this new method inside `sdk/cs/test/FoundryLocal.Tests/FoundryLocalManagerTest.cs`?

aaronpowell: Done

sdk/cs/src/FoundryManager.cs
aaronpowell: I think this should be sorted

sdk/cs/src/FoundryManager.cs
aaronpowell: Done - rebased the changes on top and force updated the branch

PR #122, opened 2025-05-29
Fixup workflow so it runs cargo fmts in the sub directories.

chendrixson:
Chose to add a cargo.toml file in the samples directory and cargo fmt in both places rather than having a top level cargo.toml workspace file.  Better to keep the root cleaner.

### DIFF
--- a/.github/workflows/rustfmt.yml
+++ b/.github/workflows/rustfmt.yml
@@ -1,4 +1,4 @@
-name: rustfmt
+name: Rust-fmt
 
 on:
   pull_request:
@@ -21,5 +21,9 @@ jobs:
         uses: actions/checkout@v4
       - name: Update toolchain
         run: rustup update --no-self-update stable && rustup default stable
-      - name: Check
+      - name: Check SDK
+        working-directory: sdk/rust
+        run: cargo fmt --all -- --check
+      - name: Check Samples
+        working-directory: samples
         run: cargo fmt --all -- --check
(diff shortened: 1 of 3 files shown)


.github/workflows/rustfmt.yml
```diff
@@ -0,0 +1,29 @@
...
+  pull_request:
+    paths:
+      - 'sdk/rust/**'
+      - 'samples/rust/**'
+  push:
+    paths:
+      - 'sdk/rust/**'
+      - 'samples/rust/**'
+    branches:
+      - main
```
snnn: Add your branch name? 

Or, you must resolve the conflicts first. Now your PR has conflicts:
```
This branch has conflicts that must be resolved
```
So no PR pipeline can run.

.github/workflows/rustfmt.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Update toolchain
+        run: rustup update --no-self-update stable && rustup default stable
+      - name: Check SDK
+        working-directory: sdk/rust
+        run: cargo fmt --all -- --check
+      - name: Check Samples
+        working-directory: samples
```
jonwis: Should this say samples/rust ?

chendrixson: Yes indeed!

chendrixson:
Found a fmt issue, so it's working!  Will fix the issues and push an update.

chendrixson:
Everything looking good with the workflow, merging!

PR #158, opened 2025-06-10
update rust sample for first run resilience

cicorias:
- uses a smaller model so download is faster
- adds catalog check if model exists
- adds local cache check and downloads as needed


samuel100:
@cicorias Thanks for the contribution 🙏!

PR #188, opened 2025-07-08
[WIP] Add  upgrade and upgradeable APIs to JS and Python SDKs

amarin16:
This depends on the `/openai/upgradable/` and `/openai/upgrade` endpoints being available in FL.

### DIFF
--- a/sdk/python/foundry_local/api.py
+++ b/sdk/python/foundry_local/api.py
@@ -271,6 +271,62 @@ class FoundryLocalManager:
             )
         return model_info
 
+    def is_model_upgradable(self, alias_or_model_id: str) -> bool:
+        """
+        Check if a newer version of a model is available.
+
+        Args:
+            alias_or_model_id (str): Alias or Model ID.
+
+        Returns:
+            bool: True if a newer version is available, False otherwise.
+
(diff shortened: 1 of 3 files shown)


sdk/python/foundry_local/api.py
```diff
@@ -190,11 +210,67 @@ def get_model_info(self, alias_or_model_id: str, raise_on_not_found: bool = Fals
...
+                            best_version = version
+                            model_info = info
+                    except ValueError:
+                        continue  # Skip if version is not numeric
+
         if model_info is None and raise_on_not_found:
             raise ValueError(f"Model {alias_or_model_id} not found in the catalog.")
         return model_info
 
+    def _get_latest_model_info(self, alias_or_model_id: str, raise_on_not_found: bool = False) -> FoundryModelInfo | None:
```
jambayk, line 238: for this it might be just simpler to do `self.get_model_info(alias_or_model_id.split(":")[0])`? all of the extra logic here leads to the same thing

amarin16: That also works, I've updated.

jambayk (approved):

PR #204, opened 2025-07-29
Add SDK/version as User-Agent in HTTPClients

bmehta001:
- Embeds SDK + version into the default User-Agent header for Http Client for C#/JS/Python/Rust
- Bump patch versions where necessary - not sure it's necessary for Rust
- Pull versions programmatically to avoid hardcoding
- Used genversion library for JavaScript and Reflection for C#
- Lint/add node_modules to .gitignore
- Tested C#, Rust, Python, and JS

**PR Summary**
The changes update the HTTP client implementations in Rust, C#, Python, and JavaScript to dynamically include the SDK version in the User-Agent header. This ensures consistent version reporting while also bumping the overall package versions.

- sdk/rust/src/client.rs: Updated the client builder to use the dynamic version from Cargo.
- sdk/cs/src/FoundryLocalManager.cs: Configured the C# client to extract the assembly version for the User-Agent header.
- sdk/python/foundry_local/client.py and new sdk/python/foundry_local/version.py: Modified the Python client to import and use the SDK version in its header.
- sdk/js/src/client.ts and new sdk/js/src/version.ts: Adjusted the JavaScript client and build script to inject the SDK version into the User-Agent header.


sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -14,7 +14,7 @@
       <ImplicitUsings>enable</ImplicitUsings>
       <Nullable>enable</Nullable>
       <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
-      <Version>0.1.0</Version>
+      <Version>0.1.1</Version>
```
amarin16: I've already updated the minor version in the latest PR, so the versions no longer need to be updated for the next release

sdk/js/src/version.ts
```diff
@@ -0,0 +1,2 @@
+// Generated by genversion.
+export const version = '0.3.2'
```
amarin16: Let's make this 0.4.0

sdk/python/foundry_local/version.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.3.2"
```
amarin16: "0.4.0"

sdk/cs/src/FoundryLocalManager.cs
```diff
@@ -47,6 +48,7 @@ public partial class FoundryLocalManager : IDisposable, IAsyncDisposable
     private List<ModelInfo>? _catalogModels;
     private Dictionary<string, ModelInfo>? _catalogDictionary;
     private readonly Dictionary<ExecutionProvider, int> _priorityMap;
+    private static readonly string AssemblyVersion = typeof(FoundryLocalManager).Assembly.GetName().Version?.ToString() ?? "unknown";
```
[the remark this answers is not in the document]
bmehta001: Unnecessary - we know that Version will be available

sdk/python/foundry_local/version.py
bmehta001: Fixed

sdk/js/src/version.ts
bmehta001: Fixed

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
bmehta001: Thanks, fixed all

bmehta001: And merged main

amarin16 (approved):

PR #238, opened 2025-09-11
Fix Python version

bmehta001:
- Fix Python version by exporting it in __init__ file
- `pip install -e .` now works


jambayk:
I made this chang in #239 to fix the pyproject.toml. Is this still required? I can do `pip install -e .` fine

PR #307, opened 2025-11-16
C# SDK Refresh

samuel100:
Refresh of C# SDK

### DIFF
--- a/samples/cs/GettingStarted/nuget.config
+++ b/samples/cs/GettingStarted/nuget.config
@@ -10,7 +10,7 @@
       <package pattern="*" />
     </packageSource>
     <packageSource key="ORT">
-      <package pattern="Microsoft.ML.OnnxRuntime.Foundry" />
+      <package pattern="*Foundry*" />
     </packageSource>
   </packageSourceMapping>
 </configuration>


samples/cs/GettingStarted/cross-platform/AudioTranscriptionExample/AudioTranscriptionExample.csproj
```diff
@@ -0,0 +1,37 @@
...
+  <!-- Packages -->
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AI.Foundry.Local" />
+  </ItemGroup>
+
+  <!-- ONNX Runtime GPU and CUDA provider (required for Linux)-->
+  <ItemGroup Condition="'$(RuntimeIdentifier)' == 'linux-x64'">
+    <PackageReference Include="Microsoft.ML.OnnxRuntime.Gpu" />
+    <PackageReference Include="Microsoft.ML.OnnxRuntimeGenAI.Cuda" />
+  </ItemGroup>
```
skottmckay, line 25: This should be covered by the last part of ExcludeExtraLibs.props. Can cleanup later though.

Causes this warning as a) Core add the GenAI.Cuda as a dependencies, then b) we add it in the props file to be explicit about the version as we use that in the path to copy. A third addition here really shouldn't be necessary.

/home/s/github/fl.sdk/samples/cs/GettingStarted/cross-platform/FoundryLocalWebServer/FoundryLocalWebServer.csproj : warning NU1504: Duplicate 'PackageReference' items found. Remove the duplicate items or use the Update functionality to ensure a consistent restore behavior. The duplicate 'PackageReference' items are: Microsoft.ML.OnnxRuntimeGenAI.Cuda , Microsoft.ML.OnnxRuntimeGenAI.Cuda 0.11.0.

samples/cs/GettingStarted/ExcludeExtraLibs.props
skottmckay, line 1: Should we rename this at some point given it's now copying .so file on linux?

PR #309, opened 2025-11-17
Linux x64 build fixes

skottmckay:


samuel100 (commented):
I've added the props changes into my branch, but left the csproj files the same (with publish)

Issue #325, opened 2025-11-25
add descriptive/better error message (include URL) if model download fails


bmerkle:
`foundry model download phi-4-mini ` fails in my company environment (as described in https://github.com/microsoft/Foundry-Local/issues/323)

seems that `eastus.api.azureml.ms` is somehow blocked in my company environment.

However following the trace i can not see the destination URL (which is currently blocked in company environment)

Please: could you add a more descriptive error message if model download fails ? (e.g. output the destination URL)

```
PS C:\Users\merklbe1\MyApp> foundry model download phi-3.5-mini
Downloading Phi-3.5-mini-instruct-cuda-gpu:1...
[14:20:52 ERR] Failed to download model Phi-3.5-mini-instruct-cuda-gpu:1, error The SSL connection could not be established, see inner exception.
Failed: Downloading Phi-3.5-mini-instruct-cuda-gpu:1
Exception: The SSL connection could not be established, see inner exception.
[14:20:52 ERR] LogException
Microsoft.AI.Foundry.Local.Common.FLException: The SSL connection could not be established, see inner exception.
   at Microsoft.AI.Foundry.Local.Common.ModelManagement.<DownloadModelAsync>d__9.MoveNext() + 0x2304
--- End of stack trace from previous location ---
   at Microsoft.AI.Foundry.Local.Common.ModelManagement.<DownloadModelAsync>d__9.MoveNext() + 0x25e5
--- End of stack trace from previous location ---
   at Microsoft.AI.Foundry.Local.Commands.ModelDownloadCommand.<<Create>b__1_0>d.MoveNext() + 0x137
--- End of stack trace from previous location ---
   at Microsoft.AI.Foundry.Local.Common.CommandActionFactory.<>c__DisplayClass0_0`1.<<Create>b__0>d.MoveNext() + 0x1e7
--- End of stack trace from previous location ---
   at System.CommandLine.NamingConventionBinder.CommandHandler.<GetExitCodeAsync>d__66.MoveNext() + 0x50
--- End of stack trace from previous location ---
   at System.CommandLine.NamingConventionBinder.ModelBindingCommandHandler.<InvokeAsync>d__11.MoveNext() + 0x61
--- End of stack trace from previous location ---
   at System.CommandLine.Invocation.InvocationPipeline.<InvokeAsync>d__0.MoveNext() + 0x1cd
--- End of stack trace from previous location ---
[...]

PR #326, opened 2025-11-25
fix for #325: add descriptive/better error message (include URL) if model download fails 

bmerkle:
- added handling for failed downloads
- output url which causes the problem
- add unit test this case


bmerkle:
@samuel100 or @MaanavD can you please have a look at this PR ? 
I fixed a real world problem in company IT environments. Thanks so much.

baijumeswani:
The source code has significantly changed since this PR. Closing this. Please re-open if still relevant.

PR #348, opened 2025-12-09
Fix README install instructions

martinpeck:
This PR fixes some formatting issues related to the MacOS install instructions. At the same time, it tidies up the Windows instructions that come just before.

The current README looks like this. Notice that the MacOS instructions are poorly formatted and start with `bash`:

image

The PR fixes this and the result is as follows:

image


baijumeswani:
The README has significantly changed since this PR. Closing this. Please re-open if still relevant.

PR #360, opened 2025-12-19
[Demo] Sample for official OpenAI library integration (via custom transport)

trrwilson:
This PR is a non-exhaustive demonstration of using a custom pipeline transport (concept from `System.ClientModel`) to facilitate with the official OpenAI library from https://github.com/openai/openai-dotnet. It's not intended to merge as-is, but rather provide  a reference for official client integration options and approaches.

Only non-streaming is initially implemented -- but streaming shouldn't be too different, especially with impl sources available.

`Program.cs` is an adapted `HelloFoundryLocalSdk` , but uses the included `FoundryLocalChatClient` for use of the `OpenAI`.

> [!NOTE]
> Integrated into Foundry Local, client retrieval wouldn't require the new public type, as something like the following would "just work" to provide a client:
> ```csharp
> // Internally: return new FoundryLocalChatClient(this)
> ChatClient client = model.GetChatClient();
> ```

The guts of the derived client lie in the `FoundryLocalPipelineTransport` and its nested classes. This transport replaces standard HTTP traffic behavior with use of Foundry Local CoreInterop, instead, stubbing substantial portions of functionality that aren't applicable to the scenario.

As not all CoreInterop capabilities have external visibility, this use relies on Reflection via an encapsulated `FoundryLocaInteropWrapper` type at the end of the custom transport's nested classes.

### DIFF
--- a/samples/cs/GettingStarted/Directory.Packages.props
+++ b/samples/cs/GettingStarted/Directory.Packages.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <OnnxRuntimeGenAIVersion>0.11.2</OnnxRuntimeGenAIVersion>
-    <OnnxRuntimeVersion>1.23.2</OnnxRuntimeVersion>    
+    <OnnxRuntimeVersion>1.23.2</OnnxRuntimeVersion>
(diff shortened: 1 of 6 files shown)


baijumeswani:
The samples code has significantly changed since this PR. Closing this. Please re-open if still relevant.

PR #376, opened 2026-01-14
feat(www): unhide whisper models and display as SDK-only

MaanavD:
- Remove filter that hid automatic-speech-recognition models
- Add whisper model detection by name/alias in addition to taskType
- Group all whisper variants (tiny, base, large, turbo) into single card
- Display SDK-only notice with documentation link for speech-to-text models
- Update deduplication logic to preserve different whisper sizes as variants


metang (approved):
approved

PR #398, opened 2026-01-24
Add multi-language and temperature support in js SDK

selenayang888:
Add auto detection and multi-language with temperature support in js SDK

### DIFF
--- a/sdk_v2/js/src/openai/audioClient.ts
+++ b/sdk_v2/js/src/openai/audioClient.ts
@@ -1,5 +1,38 @@
 import { CoreInterop } from '../detail/coreInterop.js';
 
+export class AudioClientSettings {
+    language?: string;
+    temperature?: number;
+
+    /**
+     * Serializes the settings into an OpenAI-compatible request object.
+     * @internal
+     */
+    _serialize() {
+        // Standard OpenAI properties
+        const result: any = {
+            Language: this.language,
+            Temperature: this.temperature,
+        };
+
+        // Foundry specific metadata properties
+        const metadata: Record<string, string> = {};
+        if (this.language !== undefined) {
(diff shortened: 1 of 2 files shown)


sdk_v2/js/src/openai/audioClient.ts
```diff
@@ -18,10 +18,12 @@ export class AudioClient {
      * @param audioFilePath - Path to the audio file to transcribe.
      * @returns The transcription result.
      */
-    public async transcribe(audioFilePath: string): Promise<any> {
+    public async transcribe(audioFilePath: string, language: string | null = null, temperature: number = 0.0): Promise<any> {
```
prathikr: should we have some sort of audio_inputs dict instead of passing each input separately? can also add some validation for invalid/extra parameters

sdk_v2/js/test/openai/audioClient.test.ts
```diff
@@ -30,7 +30,42 @@ describe('Audio Client Tests', () => {
...
+        if (!model || !cachedVariant) return;
+
+        model.selectVariant(cachedVariant.id);
+        await model.load();
+        
+        try {
+            const audioClient = model.createAudioClient();
+            expect(audioClient).to.not.be.undefined;
+
+            const response = await audioClient.transcribe(AUDIO_FILE_PATH, "en", 0.1);
```
prathikr: If english is the default, we should have a test in another language to actually test out this feature

sdk_v2/js/src/openai/audioClient.ts
prathikr: it can be similar to the chatSettings object in chatClient

prathikr (approved):

PR #401, opened 2026-01-26
[sdk_v2] install nightly packages in CI

prathikr:
allows CIs to test latest Foundry Local Core features

### DIFF
--- a/.github/workflows/build-cs-steps.yml
+++ b/.github/workflows/build-cs-steps.yml
@@ -37,43 +37,14 @@ jobs:
         env:
           NUGET_AUTH_TOKEN: ${{ secrets.AZURE_DEVOPS_PAT }}
 
-      # - name: Resolve Latest Core Version
-      #   shell: pwsh
-      #   run: |
-      #     $UseWinML = "${{ inputs.useWinML }}"
-      #     $PackageId = if ($UseWinML -eq 'true') { "Microsoft.AI.Foundry.Local.Core.WinML" } else { "Microsoft.AI.Foundry.Local.Core" }
-      #     $FeedUrl = "https://pkgs.dev.azure.com/aiinfra/PublicPackages/_packaging/ORT-Nightly/nuget/v3/index.json"
-          
-      #     Write-Host "Resolving latest nightly version for $PackageId from $FeedUrl..."
-          
-      #     try {
-      #         $serviceIndex = Invoke-RestMethod -Uri $FeedUrl
-      #         $baseAddressRes = $serviceIndex.resources | Where-Object { $_.'@type' -like "PackageBaseAddress/3.0.0*" } | Select-Object -First 1
-      #         $baseAddress = $baseAddressRes.'@id'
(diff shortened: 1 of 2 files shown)


.github/workflows/build-cs-steps.yml
```diff
@@ -10,6 +10,10 @@ on:
         required: false
         type: boolean
         default: false
+      useNightly:
+        required: false
+        type: boolean
+        default: true
```
baijumeswani: When would one not use nightly in the CI pipeline?

.github/workflows/build-cs-steps.yml
```diff
@@ -40,13 +44,48 @@ jobs:
         env:
           NUGET_AUTH_TOKEN: ${{ secrets.AZURE_DEVOPS_PAT }}
 
+      - name: Add ORT Feed (Nightly)
+        if: ${{ inputs.useNightly }}
+        run: dotnet nuget add source "https://pkgs.dev.azure.com/aiinfra/PublicPackages/_packaging/ORT/nuget/v3/index.json" -n ORT --configfile ../nuget.config
```
baijumeswani: Should we update nuget.config to always look at the nightly feed? That way, devs who are builidng the package from source can also fetch the nightly package when building.

baijumeswani: Why not use the ORT-Nightly feed?

.github/workflows/build-cs-steps.yml
```diff
@@ -40,13 +44,48 @@ jobs:
...
+
+      - name: Resolve Nightly Version
```
baijumeswani: Why do we need to do this? Could we take a dependency on version `*` to get the absolute latest from that feed?

prathikr: `*` provides latest stable version, not pre-release
I also tried `*-*` but that also returned latest stable

prathikr: The idea of pulling all versions string, sorting the list, and returning element[0] is also the same across js/c# if we do it this way, js definitely doesn't have a way to pull latest pre-release version with a `*` using http

sdk_v2/js/script/install.cjs
```diff
@@ -39,37 +39,49 @@ const REQUIRED_FILES = [
...
     name: 'Microsoft.ML.OnnxRuntime.Foundry', 
-    version: '1.23.2', 
+    version: '1.23.2', // Hardcoded stable version
     files: ['onnxruntime'],
     feed: ORT_FEED
   },
   { 
     name: useWinML ? 'Microsoft.ML.OnnxRuntimeGenAI.WinML' : 'Microsoft.ML.OnnxRuntimeGenAI.Foundry', 
-    version: '0.11.4', 
+    version: '0.11.4', // Hardcoded stable version
```
baijumeswani: should use 0.11.2 since 0.11.4 has known issues.

baijumeswani (approved):

PR #406, opened 2026-01-29
JavaScript SDK_v2 examples

samuel100:
Added examples for the JavaScript SDK_v2.


samples/js/audio-transcription-example/app.js
```diff
@@ -0,0 +1,44 @@
...
+console.log(`\nLoading model ${modelAlias}...`);
+await model.load();
+console.log('✓ Model loaded');
+
+// Create audio client
+console.log('\nCreating audio client...');
+const audioClient = model.createAudioClient();
+console.log('✓ Audio client created');
+
+// Example audio transcription
```
prathikr, line 34: Should we include a streaming example as well?

samuel100: Good idea. I've added into the audio transcription example to show both methods.

PR #426, opened 2026-02-09
[sdk_v2] version fixes for bug bash

prathikr:


sdk_v2/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -98,7 +98,9 @@
     
     <PropertyGroup>
       <!-- default version unless overridden during dotnet build/restore command -->
-      <FoundryLocalCoreVersion Condition="'$(FoundryLocalCoreVersion)' == ''">0.8.1.0</FoundryLocalCoreVersion>
+      <FoundryLocalCoreVersion Condition="'$(FoundryLocalCoreVersion)' == ''">0.9.0.8-dev.20260209T090216.b5352143</FoundryLocalCoreVersion>
+      <FoundryLocalCoreWinMLVersion Condition="'$(FoundryLocalCoreVersion)' != ''">$(FoundryLocalCoreVersion)</FoundryLocalCoreWinMLVersion>
+      <FoundryLocalCoreWinMLVersion Condition="'$(FoundryLocalCoreVersion)' == ''">0.9.0.7-dev.20260209T090407.b5352143</FoundryLocalCoreWinMLVersion>
```
baijumeswani: I had this issue in my original commit. We need to reverse the order here otherwise `FoundryLocalCoreVersion` gets defined in the line above and will never evaluate to "".

baijumeswani (approved):

PR #427, opened 2026-02-09
[sdk_v2] adjust install.cjs to include dxil.dll and dxcompiler.dll into the packages directory

prathikr:
fixes bug when running webgpu models on win-x64


prathikr:
waiting on Baiju to test this manually on win-x64 before merging...

PR #432, opened 2026-02-09
[sdk_v2] print variants in js sdk example

prathikr:
Defaults the example to printing the model variants so we can confirm if EPs have been loaded correctly across platforms


skottmckay (approved):
:shipit:

PR #467, opened 2026-02-27
fix: return promise from Model.download() to enable await with progre…

samuel100:
Model.download() was discarding the Promise returned by ModelVariant.download(), causing await to resolve immediately. This meant load() could run before the download completed, resulting in 'Model path does not exist' errors.


baijumeswani:
@copilot please merge main into this branch

baijumeswani (approved):

PR #471, opened 2026-02-27
Fix the bug: when the file does not exist, it should throw the error

selenayang888:
Fix the bug: when the file does not exist, it should throw the error message to user.


baijumeswani:
@copilot please merge main into this branch.

baijumeswani (approved):

PR #486, opened 2026-03-05
Add live audio transcription streaming support to Foundry Local JS SDK

rui-ren:
Here's the corrected PR description with all names aligned to the actual code:

---

Adds real-time audio streaming support to the Foundry Local JS SDK, enabling live microphone-to-text transcription via ONNX Runtime GenAI ASR.

The existing AudioClient only supports file-based transcription. This PR introduces `LiveAudioTranscriptionSession` that accepts continuous PCM audio chunks (e.g., from a microphone) and returns partial/final transcription results as an async iterable.

### New files
- `src/openai/liveAudioTranscriptionClient.ts` — Streaming session with `start()`, `append()`, `getTranscriptionStream()`, `stop()`, `dispose()`
- `src/openai/liveAudioTranscriptionTypes.ts` — `LiveAudioTranscriptionResponse` and `CoreErrorResponse` interfaces, `tryParseCoreError()` helper
- `src/detail/coreInterop.ts` — Added `executeCommandWithBinary()` method and `StreamingRequestBuffer` struct for binary PCM data transport
- app.js — E2E example with microphone capture (naudiodon2) and synthetic audio fallback
- `test/openai/liveAudioTranscription.test.ts` — Unit tests for types/settings and E2E test with synthetic PCM audio

### Modified files
- `src/imodel.ts` — Added `createLiveTranscriptionSession()` to interface
- `src/model.ts` — Delegates to `selectedVariant.createLiveTranscriptionSession()`
- `src/modelVariant.ts` — Implementation (creates new `LiveAudioTranscriptionSession(modelId, coreInterop)`)
- `src/index.ts` — Exports `LiveAudioTranscriptionSession`, `LiveAudioTranscriptionOptions`, `LiveAudioTranscriptionResponse`, `TranscriptionContentPart`

## API surface

```js
const session = model.createLiveTranscriptionSession();

session.settings.sampleRate = 16000;
session.settings.channels = 1;
session.settings.language = "en";

await session.start();

// Push audio from microphone callback
await session.append(pcmBytes);

// Read results as async iterable
for await (const result of session.getTranscriptionStream()) {
    console.log(result.content[0].text);
}

await session.stop();
```

## Design highlights
- **Internal async push queue** — Bounded `AsyncQueue<T>` serializes audio pushes from any context (safe for mic callbacks) and provides backpressure via FIFO resolver queue. Mirrors C#'s `Channel<T>` pattern.
- **Binary data transport** — `executeCommandWithBinary()` sends PCM bytes alongside JSON params via `StreamingRequestBuffer`, with transcription results parsed from push responses.
- **Settings freeze** — Audio format settings are snapshot-copied and `Object.freeze()`d at `start()`, immutable during the session
- **Buffer copy** — `append()` copies the input `Uint8Array` before queueing, safe when caller reuses buffers
- **Drain-on-stop** — `stop()` completes the push queue, waits for the push loop to drain, parses final transcription from stop response, then completes the output stream
- **Error propagation** — `start()` failures are propagated to `outputQueue` so `getTranscriptionStream()` consumers see the error; `tryParseCoreError()` handles both raw JSON and CoreInterop-prefixed error messages
- **Dispose safety** — `dispose()` wraps `stop()` in try/catch, never throws

## Native core dependency
This PR adds the JS SDK surface. The 3 native commands (`audio_stream_start`, `audio_stream_push`, `audio_stream_stop`) are routed through `execute_command` and the new `execute_command_with_binary` exports. The code compiles with zero TypeScript errors without the native library.

## Testing
- ✅ TypeScript compilation — 0 errors across all source files
- ✅ Unit tests for `parseTranscriptionResult()`, `tryParseCoreError()`, `LiveAudioTranscriptionOptions`
- ✅ E2E test with synthetic PCM audio (skips gracefully if native core unavailable)

## Parity with C# SDK
This implementation mirrors the C# `LiveAudioTranscriptionSession` with identical logic:
- Same session lifecycle: `start` → `append` → `getStream` → `stop`
- Same push loop with error handling and binary data transport
- Same settings freeze and buffer copy semantics
- Same drain-before-stop ordering with final result parsing
- Same E2E test pattern (synthetic 440Hz sine wave, 100ms chunks, ConversationItem-shaped response validation)
- Same renamed types: `LiveAudioTranscription*` (matching C# rename)

---

Changes from the original:
| Old (incorrect) | New (matches code) |
|---|---|
| `LiveAudioTranscriptionClient` | `LiveAudioTranscriptionSession` |
| `LiveAudioTranscriptionSettings` | `LiveAudioTranscriptionOptions` |
| `LiveAudioTranscriptionResult` | `LiveAudioTranscriptionResponse` |
| `createLiveTranscriptionClient()` | `createLiveTranscriptionSession()` |

### DIFF
--- a/sdk/js/src/openai/liveAudioTranscriptionTypes.ts
+++ b/sdk/js/src/openai/liveAudioTranscriptionTypes.ts
@@ -56,7 +56,6 @@ export function parseTranscriptionResult(json: string): LiveAudioTranscriptionRe
 
 /**
  * Structured error response from native core audio streaming commands.
- * @internal
  */
 export interface CoreErrorResponse {
     /** Machine-readable error code. */


sdk/js/src/openai/liveAudioTranscriptionClient.ts
```diff
@@ -0,0 +1,369 @@
+import { CoreInterop } from '../detail/coreInterop.js';
+import { LiveAudioTranscriptionResult, tryParseCoreError } from './liveAudioTranscriptionTypes.js';
+
```
kunal-vaishnavi, line 3: Can you add an E2E example in the samples folder using the JS SDK?

sdk/js/src/openai/liveAudioTranscriptionClient.ts
```diff
@@ -0,0 +1,369 @@
...
+        const copy = new LiveAudioTranscriptionSettings();
+        copy.sampleRate = this.sampleRate;
+        copy.channels = this.channels;
+        copy.bitsPerSample = this.bitsPerSample;
+        copy.language = this.language;
+        copy.pushQueueCapacity = this.pushQueueCapacity;
+        return Object.freeze(copy) as LiveAudioTranscriptionSettings;
+    }
+}
+
```
kunal-vaishnavi, line 32: There are some Copilot-raised issues with this file.

---

## Major issues / likely bugs

### 1) `audio_stream_push` does not appear to send actual audio bytes (only length)
In `pushLoop()`, the command payload includes `SessionHandle` and `AudioDataLength`, but not the `audioData` buffer itself:

```ts
this.coreInterop.executeCommand("audio_stream_push", {
  Params: {
    SessionHandle: this.sessionHandle!,
    AudioDataLength: audioData.length.toString()
  }
});
```

Unless `CoreInterop.executeCommand` implicitly reads from some shared memory or side channel (not shown here), this looks like a functional bug: the core can’t transcribe audio it never receives. At minimum, this deserves a comment explaining the data path; ideally, pass bytes explicitly (e.g., `AudioDataBase64`, `AudioData`, or a binary channel supported by `CoreInterop`).

**Recommendation:** confirm the expected contract for `"audio_stream_push"` and update the call to include the buffer (or document how `CoreInterop` transfers it).

---

### 2) `AsyncQueue` backpressure is not safe for “multiple producers”
The comment says: “multiple producers writing” but the implementation has a single `backpressureResolve` slot:

- If the queue is full, `write()` stores `this.backpressureResolve = resolve`.
- If *two* producers call `write()` while full, the second overwrites the first resolver and the first producer can hang forever.

Also, `tryWrite()` ignores `maxCapacity` entirely and can grow memory unbounded even when a capacity was configured—this defeats the point of `pushQueueCapacity`.

**Recommendation (choose one):**
- Make `AsyncQueue` explicitly **single-producer** and enforce/document it; or
- Fix it for real multi-producer backpressure by maintaining a FIFO of pending resolvers (e.g., `backpressureResolvers: Array<() => void>`), and enforce capacity in `tryWrite` (return `false` when full).

---

### 3) Potential “lost wakeup” / fairness issues in backpressure release
The consumer iterator releases backpressure when `queue.length < maxCapacity`, but only one waiter can be released (`backpressureResolve` is a single slot). Even with a single producer, this can create awkward behavior under bursty load; with multiple producers it’s incorrect as noted above.

---

## Lifecycle / concurrency concerns

### Stop ordering is a bit odd (abort happens after draining)
`stop()` does:
1) `this.pushQueue?.complete()` (drain then end)
2) `await this.pushLoopPromise`
3) `this.sessionAbortController?.abort()`

But `abort()` can’t influence the loop if you’ve already awaited it. If your intent is “stop now, don’t drain”, you’d abort first. If your intent is “drain then stop”, you don’t need abort at all (except maybe to stop other background tasks).

**Recommendation:** decide semantics and align:
- **Drain semantics:** remove abort checks from push loop, or keep them but don’t call `abort()` (or call it earlier only for external cancellation).
- **Immediate stop semantics:** `abort()` first, then `pushQueue.complete()` (optionally with an error), then `await pushLoopPromise`.

### Push calls during/after stop
`pushAudioData()` checks `this.stopped`, so once `stop()` sets `stopped = true`, further pushes throw (good). But if someone calls `pushAudioData()` concurrently right as stop begins, it may pass the check and then block forever if the queue is completed (your `write()` throws if completed, so it should reject—good), but consider race windows if `pushQueue` becomes null in the future refactor.

---

## Error handling / API ergonomics

- `start()` catch calls `this.outputQueue.complete()` but doesn’t pass the error. That means consumers see a normal completion rather than failure if they started reading early.
  - **Recommendation:** `this.outputQueue.complete(err)` so `getTranscriptionStream()` fails loudly.
- `stop()` always `this.outputQueue?.complete()` even if there was a fatal push error already set. Because `AsyncQueue.complete()` no-ops if already completed, this is fine.
- Consider surfacing the native error in a typed way rather than embedding parsed `code` into a string. You already have `tryParseCoreError()`; you could add a custom error class with `code` field.

---

## Performance considerations

- `AsyncQueue` uses `Array.shift()` which is **O(n)** per item due to reindexing. For audio streaming, this could become a hotspot.
  - **Recommendation:** implement the queue as a ring buffer or keep a `headIndex` and occasionally compact.
- `pushAudioData()` always copies the buffer. This is safe but can be expensive at high throughput.
  - **Recommendation:** if callers already supply immutable chunks, consider an opt-in “no copy” mode (documented), or accept `ArrayBuffer` and slice appropriately.

---

## Minor style / maintainability notes

- `LiveAudioTranscriptionSettings.snapshot()` returns `Object.freeze(copy) as LiveAudioTranscriptionSettings`; consider returning `Readonly<LiveAudioTranscriptionSettings>` for better typing.
- `AsyncQueue.error` getter is unused in this file; either use it or remove it to reduce surface area.
- `console.error` / `console.warn` in a library can be noisy. Consider injecting a logger or using an internal debug hook.

---

## Suggested concrete changes (priority order)

1) **Fix/confirm audio bytes transport** in `"audio_stream_push"` (most critical).
2) **Fix `AsyncQueue` multi-producer backpressure** (or document it’s single-producer), and enforce capacity in `tryWrite`.
3) Replace `Array.shift()` with a more efficient queue structure.
4) Propagate startup errors to stream consumers by completing `outputQueue` with the error.
5) Clarify stop semantics (drain vs immediate cancel) and adjust abort usage accordingly.

sdk/js/src/openai/liveAudioTranscriptionClient.ts
```diff
@@ -0,0 +1,369 @@
...
+     */
+    private async pushLoop(): Promise<void> {
+        try {
+            for await (const audioData of this.pushQueue!) {
+                if (this.sessionAbortController?.signal.aborted) {
+                    break;
+                }
+
+                try {
+                    this.coreInterop.executeCommand("audio_stream_push", {
```
kunal-vaishnavi: Shouldn't this be `executeCommandWithBinary` instead? You may need to modify [interop for JS](https://github.com/microsoft/Foundry-Local/blob/main/sdk/js/src/detail/coreInterop.ts) as you did for C# to add it.

sdk/js/src/openai/liveAudioTranscriptionClient.ts
rui-ren: samples/js/live-audio-transcription-example/

done, thank you.

sdk/js/src/openai/liveAudioTranscriptionClient.ts
rui-ren: Done, finished. 


Fixed the one real bug — `start()` now passes the error to `outputQueue.complete(err)` so consumers on `getTranscriptionStream()` will see the failure instead of a silent completion.

samples/js/live-audio-transcription-example/app.js
```diff
@@ -0,0 +1,148 @@
...
+try {
+    const { default: portAudio } = await import('naudiodon2');
+
+    audioInput = portAudio.AudioIO({
+        inOptions: {
+            channelCount: 1,
+            sampleFormat: portAudio.SampleFormat16Bit,
+            sampleRate: 16000,
+            framesPerBuffer: 1600,  // 100ms chunks
+            maxQueue: 15            // buffer during event-loop blocks from sync FFI calls
```
kunal-vaishnavi, line 92: Is this the same size that was used in C#? If not, let's make sure it is to keep parity between the two E2E samples.

samples/js/live-audio-transcription-example/app.js
```diff
@@ -0,0 +1,148 @@
...
+
+let audioInput;
+try {
+    const { default: portAudio } = await import('naudiodon2');
+
+    audioInput = portAudio.AudioIO({
+        inOptions: {
+            channelCount: 1,
+            sampleFormat: portAudio.SampleFormat16Bit,
+            sampleRate: 16000,
```
kunal-vaishnavi: Can we read these values from `client.settings`?

samples/js/live-audio-transcription-example/app.js
```diff
@@ -0,0 +1,148 @@
...
+
+    audioInput.start();
+} catch (err) {
+    console.warn('⚠ Could not initialize microphone (naudiodon2 may not be installed).');
+    console.warn('  Install with: npm install naudiodon2');
+    console.warn('  Falling back to synthetic audio test...');
+    console.warn();
+
+    // Fallback: push 2 seconds of synthetic PCM (440Hz sine wave)
+    const sampleRate = 16000;
```
kunal-vaishnavi:
```suggestion
```

samples/js/live-audio-transcription-example/app.js
```diff
@@ -0,0 +1,148 @@
...
+    const duration = 2;
+    const totalSamples = sampleRate * duration;
```
kunal-vaishnavi, line 127:
```suggestion
    const totalSamples = client.settings.sampleRate * duration;
```

samples/js/live-audio-transcription-example/app.js
```diff
@@ -0,0 +1,148 @@
...
+    const pcmBytes = new Uint8Array(totalSamples * 2);
+    for (let i = 0; i < totalSamples; i++) {
+        const t = i / sampleRate;
```
kunal-vaishnavi, line 130:
```suggestion
        const t = i / client.settings.sampleRate;
```

samples/js/live-audio-transcription-example/app.js
```diff
@@ -0,0 +1,148 @@
...
+        const sample = Math.round(32767 * 0.5 * Math.sin(2 * Math.PI * 440 * t));
+        pcmBytes[i * 2] = sample & 0xFF;
+        pcmBytes[i * 2 + 1] = (sample >> 8) & 0xFF;
+    }
+
+    // Push in 100ms chunks
+    const chunkSize = (sampleRate / 10) * 2;
```
kunal-vaishnavi, line 137:
```suggestion
    const chunkSize = (client.settings.sampleRate / 10) * 2;
```

sdk/cs/src/OpenAI/ChatClient.cs
```diff
@@ -171,7 +171,7 @@ private async IAsyncEnumerable<ChatCompletionCreateResponse> ChatStreamingImplAs
             {
                 var failed = false;
 
-                var response = await _coreInterop.ExecuteCommandWithCallbackAsync(
+                await _coreInterop.ExecuteCommandWithCallbackAsync(
```
kunal-vaishnavi: Now that the C# SDK has been merged into main, let's revert all of the C# SDK changes here (unless there is a necessary reason such as a bug fix).

sdk/cs/test/FoundryLocal.Tests/Microsoft.AI.Foundry.Local.Tests.csproj
```diff
@@ -38,6 +38,10 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(RuntimeIdentifier)' != 'linux-x64' AND '$(UseWinML)' != 'true'">
+    <PackageReference Include="Microsoft.ML.OnnxRuntime.Gpu.Linux" Version="1.24.3" ExcludeAssets="all" PrivateAssets="all" />
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/Foundry-Local/pull/486/changes#r3013839695)

sdk/js/src/detail/coreInterop.ts
```diff
@@ -93,6 +104,19 @@ export class CoreInterop {
...
+    }
+
+    /**
+     * Lazily load the execute_command_with_binary native function.
+     * Only called when audio streaming is used, so older Core builds
+     * without this export don't break non-audio functionality.
+     */
+    private _getExecuteCommandWithBinary(): any {
+        if (!this.execute_command_with_binary) {
+            this.execute_command_with_binary = this.lib.func('void execute_command_with_binary(StreamingRequestBuffer *request, _Inout_ ResponseBuffer *response)');
```
kunal-vaishnavi: Why is this needed as a separate private method via lazy loading? We already update the FL Core version so all builds will export it. We can add this line below the other `this.execute_***` commands.

sdk/js/src/openai/liveAudioTranscriptionClient.ts
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+    /**
+     * Push a chunk of raw PCM audio data to the streaming session.
+     * Can be called from any context. Chunks are internally queued
+     * and serialized to native core one at a time.
+     *
+     * @param pcmData - Raw PCM audio bytes matching the configured format.
+     */
+    public async pushAudioData(pcmData: Uint8Array): Promise<void> {
```
kunal-vaishnavi: Should this be `append` to match the C# SDK API?

sdk/js/test/openai/liveAudioTranscription.test.ts
```diff
@@ -0,0 +1,107 @@
+import { describe, it } from 'mocha';
+import { expect } from 'chai';
+import { parseTranscriptionResult, tryParseCoreError } from '../../src/openai/liveAudioTranscriptionTypes.js';
+import { LiveAudioTranscriptionSettings } from '../../src/openai/liveAudioTranscriptionClient.js';
+
+describe('Live Audio Transcription Types', () => {
+
```
kunal-vaishnavi, line 8: Can we add a full E2E test like this one for C#?

https://github.com/microsoft/Foundry-Local/blob/9e0c70d0fcd8aa6dc2bdcd679dd432bf4ae7a59b/sdk/cs/test/FoundryLocal.Tests/LiveAudioTranscriptionTests.cs#L175-L178

sdk/cs/src/OpenAI/ChatClient.cs
rui-ren: You are right. Thanks

sdk/cs/test/FoundryLocal.Tests/Microsoft.AI.Foundry.Local.Tests.csproj
rui-ren: Thank you!

samples/js/live-audio-transcription-example/app.js
rui-ren: done.

rui-ren: done, thank you!

samples/js/live-audio-transcription-example/app.js
rui-ren: updated to 
```
const sampleRate = client.settings.sampleRate;
```

samples/js/live-audio-transcription-example/app.js
rui-ren: updated

sdk/js/src/openai/liveAudioTranscriptionClient.ts
rui-ren: done,

and update all the API with Documentations

sdk/js/src/detail/coreInterop.ts
rui-ren: done, updated. Thanks

samples/js/live-audio-transcription-example/app.js
rui-ren: resolved.

sdk/js/src/openai/liveAudioTranscriptionClient.ts
```diff
@@ -0,0 +1,369 @@
...
+                } catch (error) {
+                    const errorMsg = error instanceof Error ? error.message : String(error);
+                    const errorInfo = tryParseCoreError(errorMsg);
+
+                    const fatalError = new Error(
+                        `Push failed (code=${errorInfo?.code ?? 'UNKNOWN'}): ${errorMsg}`,
+                        { cause: error }
+                    );
+                    console.error('Terminating push loop due to push failure:', errorMsg);
+                    this.outputQueue?.complete(fatalError);
```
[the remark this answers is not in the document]
rui-ren: removed

sdk/js/src/openai/liveAudioTranscriptionClient.ts
```diff
@@ -0,0 +1,369 @@
...
+                try {
+                    this.coreInterop.executeCommand("audio_stream_push", {
+                        Params: {
+                            SessionHandle: this.sessionHandle!,
+                            AudioDataLength: audioData.length.toString()
+                        }
+                    });
+                } catch (error) {
+                    const errorMsg = error instanceof Error ? error.message : String(error);
+                    const errorInfo = tryParseCoreError(errorMsg);
```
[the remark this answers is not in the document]
rui-ren: done

sdk/js/src/openai/liveAudioTranscriptionClient.ts
```diff
@@ -0,0 +1,369 @@
...
+            this.waitingResolve = null;
+            resolve({ value: item, done: false });
+            return;
+        }
+
+        if (this.queue.length >= this.maxCapacity) {
+            await new Promise<void>((resolve) => {
+                this.backpressureResolve = resolve;
+            });
+        }
```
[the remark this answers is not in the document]
rui-ren: done

sdk/js/src/openai/liveAudioTranscriptionClient.ts
```diff
@@ -0,0 +1,369 @@
...
+            await this.pushLoopPromise;
+        }
+
+        this.sessionAbortController?.abort();
+
+        let stopError: Error | null = null;
+        try {
+            this.coreInterop.executeCommand("audio_stream_stop", {
+                Params: { SessionHandle: this.sessionHandle! }
+            });
```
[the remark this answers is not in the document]
rui-ren: done

sdk/js/src/imodel.ts
```diff
@@ -16,6 +17,18 @@ export interface IModel {
...
+     * The model must be loaded before calling this method.
+     * @returns A LiveAudioTranscriptionClient instance.
+     */
+    createLiveTranscriptionClient(): LiveAudioTranscriptionClient;
+    /**
+     * Creates a LiveAudioTranscriptionClient for real-time audio streaming ASR.
+     * @returns A LiveAudioTranscriptionClient instance.
+     */
+    createLiveTranscriptionClient(): LiveAudioTranscriptionClient;
     /**
```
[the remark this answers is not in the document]
rui-ren: done

sdk/js/src/openai/liveAudioTranscriptionClient.ts
```diff
@@ -0,0 +1,369 @@
...
+    tryWrite(item: T): boolean {
+        if (this.completed) return false;
+
+        if (this.waitingResolve) {
+            const resolve = this.waitingResolve;
+            this.waitingResolve = null;
+            resolve({ value: item, done: false });
+            return true;
+        }
+
```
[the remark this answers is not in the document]
rui-ren: done

samples/js/live-audio-transcription-example/app.js
rui-ren: done

rui-ren:
@copilot+claude-opus-4.6 , I have made changes based on my coworkers' feedback, please review it, and make sure I covered all. Thanks!! You do not need to make separate PR, just resolved those covered, and unresolve those need to be done. Thanks..

sdk/js/script/install.cjs
```diff
@@ -0,0 +1,357 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// Adapted from onnxruntime\js\node\script\install-utils.js
```
kunal-vaishnavi: This file looks to be really similar to `install-utils.js`. Let's modify that file as needed rather than making an almost duplicate copy. What changes are needed in the `sdk/js/script/*` files and why?

rui-ren: Yes, updated it. Thank you so much.

sdk/js/src/openai/liveAudioTranscriptionTypes.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @internal
+ */
+export interface CoreErrorResponse {
+    /** Machine-readable error code. */
+    code: string;
+    /** Human-readable error message. */
+    message: string;
+    /** Whether this error is transient and may succeed on retry. */
+    isTransient: boolean;
+}
```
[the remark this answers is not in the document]
rui-ren: fix!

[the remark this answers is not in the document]
rui-ren: fixed

sdk/js/src/detail/coreInterop.ts
```diff
@@ -93,6 +104,12 @@ export class CoreInterop {
 
         this.execute_command = this.lib.func('void execute_command(RequestBuffer *request, _Inout_ ResponseBuffer *response)');
         this.execute_command_with_callback = this.lib.func('void execute_command_with_callback(RequestBuffer *request, _Inout_ ResponseBuffer *response, CallbackType *callback, void *userData)');
+        try {
```
kunal-vaishnavi: Why do we need try-catch here? We will always have an updated FL Core build since we changed the version.

sdk/js/src/detail/coreInterop.ts
```diff
@@ -129,6 +146,57 @@ export class CoreInterop {
...
     }
 
+    /**
+     * Execute a native command with binary data (e.g., audio PCM bytes).
+     * Uses the execute_command_with_binary native entry point which accepts
+     * both JSON params and raw binary data via StreamingRequestBuffer.
+     */
+    public executeCommandWithBinary(command: string, params: any, binaryData: Uint8Array): string {
+        if (!this.execute_command_with_binary) {
+            throw new Error('execute_command_with_binary is not available in this Core build. Update Foundry Local Core to use audio streaming.');
```
kunal-vaishnavi: Why do we need this? Same comment as [here](https://github.com/microsoft/Foundry-Local/pull/486/changes#r3019105278) that the FL Core build will always be updated.

sdk/js/test/openai/liveAudioTranscription.test.ts
```diff
@@ -0,0 +1,202 @@
...
+                return;
+            }
+
+            const model = await catalog.getModel(NEMOTRON_MODEL_ALIAS);
+            expect(model).to.not.be.undefined;
+            model!.selectVariant(cachedVariant);
+            await model!.load();
+
+            try {
+                const client = model!.createLiveTranscriptionClient();
```
kunal-vaishnavi: In C#, we get an audio client and then a live transcription session from the audio client. Here, it seems to be different. Can we make this match the C# example?

https://github.com/microsoft/Foundry-Local/blob/a74261095b2f71b58933188132f7e6a78f368aec/samples/cs/live-audio-transcription-example/Program.cs#L42-L46

sdk/js/src/modelVariant.ts
```diff
@@ -149,6 +150,14 @@ export class ModelVariant implements IModel {
         return new AudioClient(this._modelInfo.id, this.coreInterop);
     }
 
+    /**
+     * Creates a LiveAudioTranscriptionClient for real-time audio streaming ASR.
+     * @returns A LiveAudioTranscriptionClient instance.
+     */
+    public createLiveTranscriptionClient(): LiveAudioTranscriptionClient {
```
kunal-vaishnavi: Why are there two `createLiveTranscriptionClient` methods: one in `model.ts` and one in `modelVariant.ts`?

sdk/js/src/detail/coreInterop.ts
rui-ren: This one, as the CI pipeline failed with the issue of SDK . That's why we add it here. However, i can remove it.

sdk/js/src/detail/coreInterop.ts
rui-ren: This was suggested by github, I will delete it now, for production code. Thanks.

samples/js/live-audio-transcription-example/README.md
```diff
@@ -0,0 +1,58 @@
...
+await session.append(pcmBytes);
+
+// Read results
+for await (const result of session.getTranscriptionStream()) {
+    console.log(result.content[0].text);       // transcribed text
+    console.log(result.content[0].transcript); // alias (OpenAI compat)
+    console.log(result.is_final);              // true for final results
+}
+
+await client.stop();
```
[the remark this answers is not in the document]
rui-ren: updated.

sdk/js/src/detail/coreInterop.ts
```diff
@@ -93,6 +104,7 @@ export class CoreInterop {
 
         this.execute_command = this.lib.func('void execute_command(RequestBuffer *request, _Inout_ ResponseBuffer *response)');
         this.execute_command_with_callback = this.lib.func('void execute_command_with_callback(RequestBuffer *request, _Inout_ ResponseBuffer *response, CallbackType *callback, void *userData)');
+        this.execute_command_with_binary = this.lib.func('void execute_command_with_binary(StreamingRequestBuffer *request, _Inout_ ResponseBuffer *response)');
```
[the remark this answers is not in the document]
rui-ren: As @kunal suggested, we will always have the latest version in production. 

Ignore this suggestion.

sdk/js/src/openai/liveAudioTranscriptionClient.ts
```diff
@@ -0,0 +1,395 @@
...
+        if (this.queue.length >= this.maxCapacity) {
+            await new Promise<void>((resolve) => {
+                this.backpressureQueue.push(resolve);
+            });
+        }
+
+        this.queue.push(item);
+    }
```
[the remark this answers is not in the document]
rui-ren: resolved.

sdk/js/src/openai/liveAudioTranscriptionClient.ts
```diff
@@ -0,0 +1,395 @@
...
+                    const errorMsg = error instanceof Error ? error.message : String(error);
+                    const errorInfo = tryParseCoreError(errorMsg);
+
+                    const fatalError = new Error(
+                        `Push failed (code=${errorInfo?.code ?? 'UNKNOWN'}): ${errorMsg}`,
+                        { cause: error }
+                    );
+                    this.outputQueue?.complete(fatalError);
+                    return;
+                }
```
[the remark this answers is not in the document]
rui-ren: fixed.

sdk/js/src/openai/liveAudioTranscriptionClient.ts
```diff
@@ -0,0 +1,395 @@
...
+     * Dispose the client and stop any active session.
+     * Safe to call multiple times.
+     */
+    public async dispose(): Promise<void> {
+        try {
+            if (this.started && !this.stopped) {
+                await this.stop();
+            }
+        } catch (error) {
+            console.warn('Error during dispose cleanup:', error instanceof Error ? error.message : String(error));
```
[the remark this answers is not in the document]
rui-ren: fixed.

samples/js/live-audio-transcription-example/app.js
```diff
@@ -0,0 +1,151 @@
...
+            maxQueue: 15            // buffer during event-loop blocks from sync FFI calls
+        }
+    });
+
+    audioInput.on('data', (buffer) => {
+        const pcm = new Uint8Array(buffer);
+        session.append(pcm).catch((err) => {
+            console.error('append error:', err.message);
+        });
+    });
```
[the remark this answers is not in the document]
rui-ren: fixed, thanks.

sdk/js/src/index.ts
```diff
@@ -6,6 +6,8 @@ export { ModelVariant } from './modelVariant.js';
 export type { IModel } from './imodel.js';
 export { ChatClient, ChatClientSettings } from './openai/chatClient.js';
 export { AudioClient, AudioClientSettings } from './openai/audioClient.js';
+export { LiveAudioTranscriptionSession, LiveAudioTranscriptionOptions } from './openai/liveAudioTranscriptionClient.js';
+export type { LiveAudioTranscriptionResponse, TranscriptionContentPart } from './openai/liveAudioTranscriptionTypes.js';
```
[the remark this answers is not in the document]
rui-ren: fixed.

sdk/js/src/openai/liveAudioTranscriptionTypes.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * A transcription result from a real-time audio streaming session.
+ * Shaped like the OpenAI Realtime API's ConversationItem so that
+ * customers access text via result.content[0].text or result.content[0].transcript.
+ */
+export interface LiveAudioTranscriptionResponse {
+    /** Unique identifier for this result (if available). */
+    id?: string | null;
+    /** Whether this is a partial (interim) or final result for this segment. */
+    is_final: boolean;
```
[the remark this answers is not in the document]
rui-ren: fixed.

sdk/js/src/openai/liveAudioTranscriptionTypes.ts
```diff
@@ -0,0 +1,95 @@
...
+    const candidates = [errorString];
+    const prefixIdx = errorString.indexOf('failed: ');
+    if (prefixIdx !== -1) {
+        candidates.push(errorString.substring(prefixIdx + 8));
+    }
+
+    for (const candidate of candidates) {
+        try {
+            const parsed = JSON.parse(candidate);
+            if (typeof parsed.code === 'string' && typeof parsed.isTransient === 'boolean') {
```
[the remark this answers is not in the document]
rui-ren: fixed.

sdk/js/test/openai/liveAudioTranscription.test.ts
```diff
@@ -0,0 +1,203 @@
...
+            }
+
+            const model = await catalog.getModel(NEMOTRON_MODEL_ALIAS);
+            expect(model).to.not.be.undefined;
+            model!.selectVariant(cachedVariant);
+            await model!.load();
+
+            try {
+                const audioClient = model!.createAudioClient();
+                const client = audioClient.createLiveTranscriptionSession();
```
kunal-vaishnavi:
```suggestion
                const session = audioClient.createLiveTranscriptionSession();
```

rui-ren: updated! thanks!

PR #500, opened 2026-03-09
rusk sdk v2 init

samuel100:
# Add Rust SDK v2

Introduces the Foundry Local Rust SDK (`sdk_v2/rust`), updated Rust samples (`samples/rust`), and CI pipeline support.

## SDK (`sdk_v2/rust`)

A safe, async Rust wrapper over the Foundry Local Core native library, providing:

- **`FoundryLocalManager`** — singleton entry point that initializes the native engine, exposes the model catalog, and manages the local web service lifecycle.
- **`Catalog`** — async model discovery with a 6-hour TTL cache. Lookup by alias or variant ID, with helpful error messages listing available alternatives on miss.
- **`Model` / `ModelVariant`** — full model lifecycle: download (with streaming progress), load, unload, cache inspection, and removal.
- **`ChatClient`** — OpenAI-compatible chat completions (non-streaming and streaming via `futures_core::Stream`). Supports temperature, top-p/top-k, response format (text, JSON, JSON schema, Lark grammar), and tool calling.
- **`AudioClient`** — OpenAI-compatible audio transcription (non-streaming and streaming). Accepts `impl AsRef<Path>` for file paths.
- **Error handling** — `thiserror`-based `FoundryLocalError` enum with discriminated variants (`LibraryLoad`, `CommandExecution`, `ModelOperation`, `Validation`, etc.) and a crate-wide `Result<T>` alias.
- **FFI bridge** (`detail/core_interop.rs`) — dynamically loads the native `.dll`/`.so`/`.dylib` via `libloading`, with platform-aware memory deallocation and a trampoline pattern for streaming callbacks. Async methods use `tokio::task::spawn_blocking` to keep the runtime unblocked.
- **Build script** (`build.rs`) — automatically downloads NuGet packages (Core, ORT, GenAI) at compile time, extracts platform-specific native libraries, and caches them across builds. Supports `winml` and `nightly` feature flags.

Re-exports `async-openai` request/response types for convenience so callers don't need a direct dependency.

## Samples (`samples/rust`)

Replaces the old `hello-foundry-local` sample with four focused examples:

- **`native-chat-completions`** — basic sync + streaming chat completion
- **`tool-calling-foundry-local`** — multi-turn tool calling with streaming chunk assembly
- **`audio-transcription-example`** — audio file transcription (sync and streaming)
- **`foundry-local-webserver`** — starts the local web service and makes requests via the OpenAI-compatible HTTP endpoint

Each sample has its own `Cargo.toml`, `README.md`, and is included in the workspace.

## CI (`.github/workflows`)

- **`build-rust-steps.yml`** — reusable composite action for Rust SDK build + test steps.
- **`foundry-local-sdk-build.yml`** — workflow that triggers on `sdk_v2/rust/**` and `samples/rust/**` changes.

## Testing

- Integration tests mirror the JavaScript SDK test suite structure: `manager_test`, `catalog_test`, `model_test`, `model_load_manager_test`, `chat_client_test`, `audio_client_test`.
- All tests are `#[ignore]` by default (require native library at runtime) and can be enabled in CI with `--include-ignored`.
- Shared test utilities (`tests/common/mod.rs`) provide config, model aliases, and helper functions.
- `cargo clippy --all-targets --all-features` passes with zero code warnings.

### DIFF
--- a/sdk_v2/rust/docs/api.md
+++ b/sdk_v2/rust/docs/api.md
@@ -49,11 +49,11 @@ pub struct FoundryLocalManager { /* private fields */ }
 
 | Method | Signature | Description |
 |--------|-----------|-------------|
-| `create` | `fn create(config: FoundryLocalConfig) -> Result<&'static Self>` | Initialise the SDK. First call creates the singleton; subsequent calls return the existing instance (config is ignored after first call). |
+| `create` | `fn create(config: FoundryLocalConfig) -> Result<&'static Self, FoundryLocalError>` | Initialise the SDK. First call creates the singleton; subsequent calls return the existing instance (config is ignored after first call). |
 | `catalog` | `fn catalog(&self) -> &Catalog` | Access the model catalog. |
-| `urls` | `fn urls(&self) -> Result<Vec<String>>` | URLs the local web service is listening on. Empty until `start_web_service` is called. |
-| `start_web_service` | `async fn start_web_service(&self) -> Result<()>` | Start the local web service. Retrieve listening URLs via `urls()`. |
-| `stop_web_service` | `async fn stop_web_service(&self) -> Result<()>` | Stop the local web service. |
+| `urls` | `fn urls(&self) -> Result<Vec<String>, FoundryLocalError>` | URLs the local web service is listening on. Empty until `start_web_service` is called. |
+| `start_web_service` | `async fn start_web_service(&self) -> Result<(), FoundryLocalError>` | Start the local web service. Retrieve listening URLs via `urls()`. |
+| `stop_web_service` | `async fn stop_web_service(&self) -> Result<(), FoundryLocalError>` | Stop the local web service. |
 
 ---
 
@@ -128,12 +128,12 @@ pub struct Catalog { /* private fields */ }
 | Method | Signature | Description |
 |--------|-----------|-------------|
 | `name` | `fn name(&self) -> &str` | Catalog name as reported by the native core. |
-| `update_models` | `async fn update_models(&self) -> Result<()>` | Refresh catalog if cache expired or invalidated. |
-| `get_models` | `async fn get_models(&self) -> Result<Vec<Arc<Model>>>` | Return all known models. |
-| `get_model` | `async fn get_model(&self, alias: &str) -> Result<Arc<Model>>` | Look up a model by alias. |
-| `get_model_variant` | `async fn get_model_variant(&self, id: &str) -> Result<Arc<ModelVariant>>` | Look up a variant by unique id. |
-| `get_cached_models` | `async fn get_cached_models(&self) -> Result<Vec<Arc<ModelVariant>>>` | Return only variants cached on disk. |
-| `get_loaded_models` | `async fn get_loaded_models(&self) -> Result<Vec<Arc<ModelVariant>>>` | Return model variants currently loaded in memory. |
+| `update_models` | `async fn update_models(&self) -> Result<(), FoundryLocalError>` | Refresh catalog if cache expired or invalidated. |
+| `get_models` | `async fn get_models(&self) -> Result<Vec<Arc<Model>>, FoundryLocalError>` | Return all known models. |
+| `get_model` | `async fn get_model(&self, alias: &str) -> Result<Arc<Model>, FoundryLocalError>` | Look up a model by alias. |
+| `get_model_variant` | `async fn get_model_variant(&self, id: &str) -> Result<Arc<ModelVariant>, FoundryLocalError>` | Look up a variant by unique id. |
+| `get_cached_models` | `async fn get_cached_models(&self) -> Result<Vec<Arc<ModelVariant>>, FoundryLocalError>` | Return only variants cached on disk. |
+| `get_loaded_models` | `async fn get_loaded_models(&self) -> Result<Vec<Arc<ModelVariant>>, FoundryLocalError>` | Return model variants currently loaded in memory. |
 
 ---
 
@@ -151,14 +151,14 @@ pub struct Model { /* private fields */ }
 | `id` | `fn id(&self) -> &str` | Unique identifier of the selected variant. |
 | `variants` | `fn variants(&self) -> &[ModelVariant]` | All variants in this model. |
 | `selected_variant` | `fn selected_variant(&self) -> &ModelVariant` | Currently selected variant. |
-| `select_variant` | `fn select_variant(&self, id: &str) -> Result<()>` | Select a variant by id. |
-| `is_cached` | `async fn is_cached(&self) -> Result<bool>` | Whether the selected variant is cached on disk. |
-| `is_loaded` | `async fn is_loaded(&self) -> Result<bool>` | Whether the selected variant is loaded in memory. |
-| `download` | `async fn download<F>(&self, progress: Option<F>) -> Result<()>` | Download the selected variant. `F: FnMut(&str) + Send + 'static` |
-| `path` | `async fn path(&self) -> Result<PathBuf>` | Local file-system path of the selected variant. |
-| `load` | `async fn load(&self) -> Result<()>` | Load the selected variant into memory. |
-| `unload` | `async fn unload(&self) -> Result<String>` | Unload the selected variant from memory. |
-| `remove_from_cache` | `async fn remove_from_cache(&self) -> Result<String>` | Remove the selected variant from the local cache. |
+| `select_variant` | `fn select_variant(&self, id: &str) -> Result<(), FoundryLocalError>` | Select a variant by id. |
+| `is_cached` | `async fn is_cached(&self) -> Result<bool, FoundryLocalError>` | Whether the selected variant is cached on disk. |
+| `is_loaded` | `async fn is_loaded(&self) -> Result<bool, FoundryLocalError>` | Whether the selected variant is loaded in memory. |
+| `download` | `async fn download<F>(&self, progress: Option<F>) -> Result<(), FoundryLocalError>` | Download the selected variant. `F: FnMut(&str) + Send + 'static` |
+| `path` | `async fn path(&self) -> Result<PathBuf, FoundryLocalError>` | Local file-system path of the selected variant. |
+| `load` | `async fn load(&self) -> Result<(), FoundryLocalError>` | Load the selected variant into memory. |
+| `unload` | `async fn unload(&self) -> Result<String, FoundryLocalError>` | Unload the selected variant from memory. |
+| `remove_from_cache` | `async fn remove_from_cache(&self) -> Result<String, FoundryLocalError>` | Remove the selected variant from the local cache. |
 | `create_chat_client` | `fn create_chat_client(&self) -> ChatClient` | Create a ChatClient bound to the selected variant. |
 | `create_audio_client` | `fn create_audio_client(&self) -> AudioClient` | Create an AudioClient bound to the selected variant. |
 
@@ -177,13 +177,13 @@ pub struct ModelVariant { /* private fields */ }
 | `info` | `fn info(&self) -> &ModelInfo` | Full metadata for this variant. |
 | `id` | `fn id(&self) -> &str` | Unique identifier. |
 | `alias` | `fn alias(&self) -> &str` | Alias shared with sibling variants. |
-| `is_cached` | `async fn is_cached(&self) -> Result<bool>` | Whether cached locally. ⚠️ Full IPC per call — prefer `Catalog::get_cached_models()` for batch use. |
-| `is_loaded` | `async fn is_loaded(&self) -> Result<bool>` | Whether currently loaded in memory. |
-| `download` | `async fn download<F>(&self, progress: Option<F>) -> Result<()>` | Download the variant. `F: FnMut(&str) + Send + 'static` |
-| `path` | `async fn path(&self) -> Result<PathBuf>` | Local file-system path. |
-| `load` | `async fn load(&self) -> Result<()>` | Load into memory. |
-| `unload` | `async fn unload(&self) -> Result<String>` | Unload from memory. |
-| `remove_from_cache` | `async fn remove_from_cache(&self) -> Result<String>` | Remove from local cache. |
+| `is_cached` | `async fn is_cached(&self) -> Result<bool, FoundryLocalError>` | Whether cached locally. ⚠️ Full IPC per call — prefer `Catalog::get_cached_models()` for batch use. |
+| `is_loaded` | `async fn is_loaded(&self) -> Result<bool, FoundryLocalError>` | Whether currently loaded in memory. |
+| `download` | `async fn download<F>(&self, progress: Option<F>) -> Result<(), FoundryLocalError>` | Download the variant. `F: FnMut(&str) + Send + 'static` |
+| `path` | `async fn path(&self) -> Result<PathBuf, FoundryLocalError>` | Local file-system path. |
+| `load` | `async fn load(&self) -> Result<(), FoundryLocalError>` | Load into memory. |
+| `unload` | `async fn unload(&self) -> Result<String, FoundryLocalError>` | Unload from memory. |
+| `remove_from_cache` | `async fn remove_from_cache(&self) -> Result<String, FoundryLocalError>` | Remove from local cache. |
 | `create_chat_client` | `fn create_chat_client(&self) -> ChatClient` | Create a ChatClient bound to this variant. |
 | `create_audio_client` | `fn create_audio_client(&self) -> AudioClient` | Create an AudioClient bound to this variant. |
 
@@ -218,8 +218,8 @@ pub struct ChatClient { /* private fields */ }
 
 | Method | Signature | Description |
 |--------|-----------|-------------|
-| `complete_chat` | `async fn complete_chat(&self, messages: &[ChatCompletionRequestMessage], tools: Option<&[ChatCompletionTools]>) -> Result<CreateChatCompletionResponse>` | Non-streaming chat completion. |
-| `complete_streaming_chat` | `async fn complete_streaming_chat(&self, messages: &[ChatCompletionRequestMessage], tools: Option<&[ChatCompletionTools]>) -> Result<ChatCompletionStream>` | Streaming chat completion. |
+| `complete_chat` | `async fn complete_chat(&self, messages: &[ChatCompletionRequestMessage], tools: Option<&[ChatCompletionTools]>) -> Result<CreateChatCompletionResponse, FoundryLocalError>` | Non-streaming chat completion. |
+| `complete_streaming_chat` | `async fn complete_streaming_chat(&self, messages: &[ChatCompletionRequestMessage], tools: Option<&[ChatCompletionTools]>) -> Result<ChatCompletionStream, FoundryLocalError>` | Streaming chat completion. |
 
 **Example:**
 ```rust
@@ -259,8 +259,8 @@ pub struct AudioClient { /* private fields */ }
 
 | Method | Signature | Description |
 |--------|-----------|-------------|
-| `transcribe` | `async fn transcribe(&self, audio_file_path: impl AsRef<Path>) -> Result<AudioTranscriptionResponse>` | Transcribe an audio file. |
-| `transcribe_streaming` | `async fn transcribe_streaming(&self, audio_file_path: impl AsRef<Path>) -> Result<AudioTranscriptionStream>` | Streaming transcription. |
+| `transcribe` | `async fn transcribe(&self, audio_file_path: impl AsRef<Path>) -> Result<AudioTranscriptionResponse, FoundryLocalError>` | Transcribe an audio file. |
+| `transcribe_streaming` | `async fn transcribe_streaming(&self, audio_file_path: impl AsRef<Path>) -> Result<AudioTranscriptionStream, FoundryLocalError>` | Streaming transcription. |
 
 **Example:**
 ```rust
@@ -345,7 +345,7 @@ pub struct JsonStream<T> { /* private fields */ }
 
 impl<T> Unpin for JsonStream<T> {}
 impl<T: DeserializeOwned> Stream for JsonStream<T> {
-    type Item = Result<T>;
+    type Item = Result<T, FoundryLocalError>;
 }
 ```
 


sdk_v2/rust/src/catalog.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Look up a model by its alias.
+    pub async fn get_model(&self, alias: &str) -> Result<Model> {
+        if alias.trim().is_empty() {
+            return Err(FoundryLocalError::Validation {
+                reason: "Model alias must be a non-empty string".into(),
+            });
+        }
+        self.update_models().await?;
+        let map = self.models_by_alias.lock().unwrap();
```
nenad1002: Unwraps in SDKs are dangeorus, and it should be used rarely, if it's going to crash the error should be returned as you already return Result.

Something like this is probably better, although we would have to discuss this.

```
let map = self.models_by_alias.lock().map_err(|_| FoundryLocalError::Internal {
    reason: "Unexpected error occured in the SDK, please retry".into(),
})?;

File: sdk_v2/rust/src/catalog.rs [hunk 2]
```diff
@@ -0,0 +1,187 @@
...
+
+/// How long the catalog cache remains valid before a refresh.
+const CACHE_TTL: Duration = Duration::from_secs(6 * 60 * 60); // 6 hours
+
+/// The model catalog provides discovery and lookup for all available models.
+pub struct Catalog {
+    core: Arc<CoreInterop>,
+    model_load_manager: Arc<ModelLoadManager>,
+    name: String,
+    models_by_alias: Mutex<HashMap<String, Model>>,
```
nenad1002: This is very inefficient, especially cause you keep cloning in method below. I don't think Clone is right thing for Model to do, but fixing that will be pretty hard, what we can do for a quick fix is the following. Wrap Model and ModelVariant to arc like: Arc<Model>, so models_by_alias: Mutex<HashMap<String, Arc<Model>>>.

Arc is a reference (like a pointer in C++), it will allow you to access without cloning full object, cloning Arc is cheap, making all model management methods faster/more efficient.

sdk_v2/rust/src/detail/core_interop.rs
```diff
@@ -0,0 +1,470 @@
...
+
+// ── FFI types ────────────────────────────────────────────────────────────────
+
+/// Request buffer passed to the native library.
+#[repr(C)]
+struct RequestBuffer {
+    command: *const i8,
+    command_length: i32,
+    data: *const i8,
+    data_length: i32,
```
nenad1002, line 28: consider making it u32, length cannot be negative

sdk_v2/rust/src/detail/core_interop.rs
```diff
@@ -0,0 +1,470 @@
...
+                let _ = tx.send(chunk.to_owned());
+            })
+        });
+
+        Ok((rx, handle))
+    }
+
+    /// Read the response buffer, free the native memory, and return the data
+    /// string or raise an error.
+    fn process_response(response: &ResponseBuffer) -> Result<String> {
```
nenad1002: Whenever you have cheap objects like ResponseBuffer (but many times even not cheap) consider not using ref, but just move it with (response: ResponseBuffer), this way you force caller not to call the method twice which would be wrong here. In C++ this would be copied, but Rust by default moves, invalidating the input data (in this case response).

sdk_v2/rust/src/detail/core_interop.rs
```diff
@@ -0,0 +1,470 @@
...
+                let slice =
+                    std::slice::from_raw_parts(response.error, response.error_length as usize);
+                String::from_utf8_lossy(slice).into_owned()
+            })
+        } else {
+            None
+        };
+
+        let data_str = if !response.data.is_null() && response.data_length > 0 {
+            Some(unsafe {
```
nenad1002: Okay for now, but this unsafe code needs to live separately, so its always easily findable, this is often a source of bugs.

sdk_v2/rust/src/detail/model_load_manager.rs
```diff
@@ -0,0 +1,73 @@
...
+impl ModelLoadManager {
+    pub(crate) fn new(core: Arc<CoreInterop>, external_service_url: Option<String>) -> Self {
+        Self {
+            core,
+            external_service_url,
+        }
+    }
+
+    /// Load a model by its identifier.
+    pub async fn load(&self, model_id: &str) -> Result<String> {
```
nenad1002: Shouldnt return anything

sdk_v2/rust/GENERATE-DOCS.md
```diff
@@ -0,0 +1,41 @@
...
+| Module / Type | Description |
+|---|---|
+| `FoundryLocalManager` | Singleton entry point — SDK initialisation, web service lifecycle |
+| `FoundryLocalConfig` | Configuration (app name, log level, service endpoint) |
+| `Catalog` | Model discovery and lookup |
+| `Model` | Grouped model (alias → best variant) |
+| `ModelVariant` | Single variant — download, load, unload |
+| `ChatClient` | OpenAI-compatible chat completions (sync + streaming) |
+| `AudioClient` | OpenAI-compatible audio transcription (sync + streaming) |
+| `CreateChatCompletionResponse` | Typed chat completion response (from `async-openai`) |
```
nenad1002, line 29: Do we want to use type from async-openai?

sdk_v2/rust/src/model.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// human-readable progress strings as they arrive from the native core.
+    pub async fn download<F>(&self, progress: Option<F>) -> Result<()>
+    where
+        F: FnMut(&str) + Send + 'static,
+    {
+        self.selected_variant().download(progress).await
+    }
+
+    /// Return the local file-system path of the selected variant.
+    pub async fn path(&self) -> Result<String> {
```
nenad1002: should return PathBuf probably

sdk_v2/rust/src/foundry_local_manager.rs
```diff
@@ -0,0 +1,109 @@
...
+
+/// Global singleton holder.
+static INSTANCE: OnceLock<FoundryLocalManager> = OnceLock::new();
+
+/// Primary entry point for interacting with Foundry Local.
+///
+/// Created once via [`FoundryLocalManager::create`]; subsequent calls return
+/// the existing instance.
+pub struct FoundryLocalManager {
+    _config: Configuration,
```
nenad1002: If you marked it as _ if its never used, just remove it, same goes for model_load_manager below.
if it's used shouldn't be marked with underscore in public API.

sdk_v2/rust/src/foundry_local_manager.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    /// URLs that the local web service is listening on.
+    ///
+    /// Empty until [`Self::start_web_service`] has been called.
+    pub fn urls(&self) -> Vec<String> {
+        self.urls.lock().unwrap().clone()
+    }
+
+    /// Start the local web service and return the listening URLs.
+    pub async fn start_web_service(&self) -> Result<Vec<String>> {
```
nenad1002: Public API should return something like Result<Vec<String>, FoundryLocalError> and not just swallow errors.

The same goes for some other methods, if it can have an issue, we do not want to swallow it and pretend it didn't happen, we need to return it to the user.

sdk_v2/rust/tests/integration.rs
```diff
@@ -0,0 +1,1081 @@
...
+//!
+//! All test modules are compiled into one binary so the native core is only
+//! initialised once (via the `OnceLock` singleton in `FoundryLocalManager`).
+//! Running them as separate binaries causes "already initialized" errors
+//! because the .NET native runtime retains state across process-level
+//! library loads.
+
+mod common;
+
+mod manager_tests {
```
prathikr: can these be split up into separate test files similar to js/cs?

.github/workflows/build-rust-steps.yml
```diff
@@ -0,0 +1,110 @@
...
+    inputs:
+      platform:
+        required: false
+        type: string
+        default: 'ubuntu' # or 'windows' or 'macos'
+      useWinML:
+        required: false
+        type: boolean
+        default: false
+      run-integration-tests:
```
prathikr, line 14: tests should run on all platforms

.github/workflows/build-rust-steps.yml
```diff
@@ -0,0 +1,110 @@
...
+jobs:
+  build:
+    runs-on: ${{ inputs.platform }}-latest
+
+    defaults:
+      run:
+        working-directory: sdk_v2/rust
+
+    env:
+      CARGO_FEATURES: ${{ inputs.useWinML && '--features winml' || '' }}
```
prathikr, line 31: evaluates to true/false and ''? the after || should be false to be more readable

.github/workflows/build-rust-steps.yml
```diff
@@ -0,0 +1,110 @@
...
+            Write-Error "Git checkout failed."
+            exit 1
+          }
+          Write-Host "`nDirectory contents:"
+          Get-ChildItem -Recurse -Depth 2 | ForEach-Object { Write-Host "  $($_.FullName)" }
+
+      - name: Check formatting
+        run: cargo fmt --all -- --check
+
+      - name: Run clippy
```
prathikr, line 84: what is clippy for? would be helpful to include a comment since its new for rust and not in js/cs

.github/workflows/build-rust-steps.yml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: Run unit tests
+        run: cargo test --lib ${{ env.CARGO_FEATURES }}
+
+      - name: Run integration tests
+        if: ${{ inputs.run-integration-tests }}
+        run: cargo test --test '*' ${{ env.CARGO_FEATURES }} -- --test-threads=1 --nocapture
+
+      - name: Package crate
+        run: cargo package ${{ env.CARGO_FEATURES }} --allow-dirty
```
prathikr, line 99: what is allow dirty? implies some sort of non-compliant build which is OK for public pipelines but would be good to document for clarity

.github/workflows/foundry-local-sdk-build.yml
```diff
@@ -56,4 +56,30 @@ jobs:
...
+      run-integration-tests: true
+    secrets: inherit
+  build-rust-windows-WinML:
+    uses: ./.github/workflows/build-rust-steps.yml
+    with:
+      platform: 'windows'
+      useWinML: true
+      run-integration-tests: true
+    secrets: inherit
+  build-rust-ubuntu:
```
prathikr: can remove ubuntu for now as it has known issues

.github/workflows/foundry-local-sdk-build.yml
```diff
@@ -56,4 +56,30 @@ jobs:
     with:
       version: '0.9.0.${{ github.run_number }}'
       platform: 'macos'
+    secrets: inherit
+
+  build-rust-windows:
```
prathikr: can you weave these into the platform specific blocks like it is for cs/js instead of having all the rust at the end? just so it is the same format as previous SDKs

sdk_v2/rust/src/detail/core_interop.rs
```diff
@@ -0,0 +1,470 @@
...
+type ExecuteCommandWithCallbackFn = unsafe extern "C" fn(
+    *const RequestBuffer,
+    *mut ResponseBuffer,
+    CallbackFn,
+    *mut std::ffi::c_void,
+);
+
+// ── Library name helpers ─────────────────────────────────────────────────────
+
+#[cfg(target_os = "windows")]
```
prathikr, line 67: can we just define the extension based on the platform instead of the entire filename and dynamically generate the full filename inline? that is how our other sdks do it

sdk_v2/rust/src/detail/core_interop.rs
```diff
@@ -0,0 +1,470 @@
...
+        }
+
+        Self::process_response(&response)
+    }
+
+    /// Async version of [`Self::execute_command`].
+    ///
+    /// Runs the blocking FFI call on a dedicated thread via
+    /// [`tokio::task::spawn_blocking`] so the async runtime is never blocked.
+    pub async fn execute_command_async(
```
prathikr, line 400: do we want async commands here? I believe executeCommand should be blocking

sdk_v2/rust/src/detail/core_interop.rs
```diff
@@ -0,0 +1,470 @@
...
+                return Ok(p);
+            }
+            // If the config value is the full path to the library itself.
+            let p = Path::new(dir);
+            if p.exists() && p.is_file() {
+                return Ok(p.to_path_buf());
+            }
+        }
+
+        // 2. Compile-time environment variable set by build.rs.
```
prathikr: only #1 and #4 are necessary, we don't provide env var override or runtime var support in the other sdks

sdk_v2/rust/src/openai/mod.rs
```diff
@@ -0,0 +1,7 @@
+mod audio_client;
```
prathikr, line 1: what are these mod files for? is it for exporting objects or more like a utils file? if not required, can they be removed?

sdk_v2/rust/README.md
```diff
@@ -0,0 +1,110 @@
+# Foundry Local Rust SDK
```
prathikr, line 1: should be expanded to call out features and provide examples on how to do each feature similar to js/c#. AI can help you with this :P

sdk_v2/rust/src/openai/mod.rs
nenad1002: This is required, this tells you where code lives (e.g. audio_client.rs)

sdk_v2/rust/src/detail/core_interop.rs
nenad1002: I've been looking at this as well, it's an async method, however tokio is blocking, look at the method spawn_blocking.

I'm not convinced Rust should be async accross the board, it would be good if someone else with Rust expertise weight in. We are making async methods which internally does sync thing, making our code more complicated accross the board cause async immediately means we have to use Arc and Mutex which are more complicated and heavy interfaces. Async vs sync code is very different in Rust. 

I would have to think about this, but I would probably try to make this method sync, and then update everything upstream where it can be sync, simplify things.

sdk_v2/rust/src/catalog.rs
samuel100: Fixed. Replaced all 8 `.lock().unwrap()` calls in `catalog.rs` with `.map_err(|_| FoundryLocalError::Internal { ... })?`. Added a new `Internal` variant to `FoundryLocalError`.

sdk_v2/rust/src/catalog.rs
samuel100: Fixed. Maps now use `Arc<Model>` and `Arc<ModelVariant>`. Public methods return Arc-wrapped types. Cloning is now cheap (reference count bump instead of full object copy).

sdk_v2/rust/src/detail/core_interop.rs
samuel100: Fixed. Changed `ResponseBuffer` length fields from signed to `u32`.

sdk_v2/rust/src/detail/core_interop.rs
samuel100: Fixed. `process_response` now takes `ResponseBuffer` by value (move semantics), enforcing single-use.

sdk_v2/rust/src/detail/core_interop.rs
samuel100: Fixed. Added `// SAFETY:` comments to every `unsafe` block throughout `core_interop.rs`. Extracted `read_native_buffer()` as a dedicated unsafe helper with safety documentation.

sdk_v2/rust/src/detail/model_load_manager.rs
samuel100: Fixed. `load()` now returns `Result<()>` instead of `Result<String>`. Callers updated.

sdk_v2/rust/GENERATE-DOCS.md
samuel100: Yes — we re-export `async-openai` types for user convenience so callers don't need a direct dependency. Keeping them in the generated docs makes sense.

sdk_v2/rust/src/model.rs
samuel100: Fixed. `path()` now returns `PathBuf` instead of `String`.

sdk_v2/rust/src/foundry_local_manager.rs
samuel100: Fixed. Removed unused `_config` and `_model_load_manager` fields entirely.

sdk_v2/rust/src/foundry_local_manager.rs
samuel100: Fixed. `start_web_service` now propagates errors via `Result` instead of silently falling back.

sdk_v2/rust/tests/integration.rs
samuel100: Fixed. Split into 6 separate test modules under `tests/integration/`: `manager_test.rs`, `catalog_test.rs`, `model_test.rs`, `chat_client_test.rs`, `audio_client_test.rs`, `web_service_test.rs`. Uses directory-based test pattern with shared `common/mod.rs`.

.github/workflows/build-rust-steps.yml
samuel100: Fixed. Changed `run-integration-tests` default to `true` so tests run on all platforms.

.github/workflows/build-rust-steps.yml
samuel100: Fixed. Changed `|| ''` to `|| false` for boolean readability.

.github/workflows/build-rust-steps.yml
samuel100: Fixed. Added comment: Clippy is Rust's official linter for catching common mistakes, enforcing idioms, and improving code quality.

.github/workflows/build-rust-steps.yml
samuel100: Fixed. Added comment explaining `--allow-dirty` flag purpose.

.github/workflows/foundry-local-sdk-build.yml
samuel100: Fixed. Removed ubuntu from the Rust CI matrix.

.github/workflows/foundry-local-sdk-build.yml
samuel100: Fixed. Rust jobs are now woven into the platform-specific blocks matching the cs/js format.

sdk_v2/rust/src/detail/core_interop.rs
samuel100: Fixed. Now defines only `LIB_EXTENSION` per platform and builds the full filename dynamically via `format!`.

sdk_v2/rust/src/detail/core_interop.rs
samuel100: Acknowledged — both reviewers lean toward sync. The native calls are blocking (`spawn_blocking` wraps them). Deferring this architectural change to a follow-up PR after explicit sign-off, as it would require changes across many files (removing Arc/Mutex, changing async method signatures).

sdk_v2/rust/src/detail/core_interop.rs
samuel100: Fixed. Removed the runtime env var override (option #3). Kept: (1) explicit config path, (2) compile-time `option_env!("FOUNDRY_NATIVE_DIR")` from build.rs (needed for dev builds), (3) exe-sibling directory.

sdk_v2/rust/src/openai/mod.rs
samuel100: Confirmed — `mod.rs` files are required in Rust to declare submodules. No change needed.

sdk_v2/rust/README.md
samuel100: Fixed. README expanded from ~110 lines to ~477 lines with features list, installation guide, quick start, per-feature usage examples, error handling, configuration reference, and sample links.

.github/workflows/build-rust-steps.yml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: Build
+        run: cargo build ${{ env.CARGO_FEATURES }}
+
+      - name: Run unit tests
+        run: cargo test --lib ${{ env.CARGO_FEATURES }}
+
+      - name: Run integration tests
+        if: ${{ inputs.run-integration-tests }}
+        run: cargo test --test '*' ${{ env.CARGO_FEATURES }} -- --test-threads=1 --nocapture
```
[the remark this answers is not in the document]
samuel100: Fixed. Changed `cargo test --test '*'` to `cargo test --tests`. Also added `--include-ignored` to run ignored integration tests in CI.

sdk_v2/rust/tests/integration.rs
```diff
@@ -0,0 +1,1081 @@
...
+    }
+
+    #[tokio::test]
+    async fn should_throw_when_completing_streaming_chat_with_invalid_callback() {
+        let (client, model) = setup_chat_client().await;
+        let messages: Vec<ChatCompletionRequestMessage> = vec![];
+
+        let result = client.complete_streaming_chat(&messages, None).await;
+        assert!(result.is_err(), "Expected error even with empty messages");
+
```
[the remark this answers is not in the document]
samuel100: Fixed. Removed the duplicate `should_throw_when_completing_streaming_chat_with_invalid_callback` test.

samples/rust/audio-transcription-example/src/main.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    // ── 2. Pick the whisper model and ensure it is downloaded ────────────
+    let model = manager.catalog().get_model(ALIAS).await?;
+    println!("Model: {} (id: {})", model.alias(), model.id());
+
+    if !model.is_cached().await? {
+        println!("Downloading model...");
+        model
+            .download(Some(|progress: &str| {
+                print!("\r  {progress:.1}%");
```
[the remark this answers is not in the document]
samuel100: Fixed. Changed `{progress:.1}%` to `{progress}%` in all 4 sample apps.

sdk_v2/rust/src/openai/audio_client.rs
```diff
@@ -0,0 +1,212 @@
...
+                .ok_or_else(|| FoundryLocalError::Validation {
+                    reason: "audio file path is not valid UTF-8".into(),
+                })?;
+        Self::validate_path(path_str)?;
+
+        let mut request = self.settings.serialize(&self.model_id, path_str);
+        if let Some(map) = request.as_object_mut() {
+            map.insert("stream".into(), json!(true));
+        }
+
```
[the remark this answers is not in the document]
samuel100: Fixed. Removed `stream: true` from the audio transcription streaming request. Streaming is now handled purely via the native callback, matching JS/C# SDKs.

sdk_v2/rust/src/detail/core_interop.rs
```diff
@@ -0,0 +1,470 @@
...
+) {
+    if data.is_null() || length <= 0 {
+        return;
+    }
+    let closure = &mut *(user_data as *mut Box<dyn FnMut(&str)>);
+    let slice = std::slice::from_raw_parts(data, length as usize);
+    if let Ok(chunk) = std::str::from_utf8(slice) {
+        closure(chunk);
+    }
+}
```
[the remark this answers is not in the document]
samuel100: Fixed. Wrapped `streaming_trampoline` body in `std::panic::catch_unwind(AssertUnwindSafe(...))` to prevent UB from panics crossing the FFI boundary.

sdk_v2/rust/src/detail/model_load_manager.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        let ids: Vec<String> = serde_json::from_str(&raw)?;
+        Ok(ids)
+    }
+
+    async fn http_get(url: &str) -> Result<String> {
+        let body = reqwest::get(url).await?.text().await?;
+        Ok(body)
+    }
```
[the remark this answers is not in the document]
samuel100: Fixed. Added `client: reqwest::Client` field to `ModelLoadManager`, initialized once in the constructor. All HTTP calls now reuse it for connection pooling.

sdk_v2/rust/build.rs
```diff
@@ -0,0 +1,304 @@
...
+    let ext = native_lib_extension();
+    if let Ok(entries) = fs::read_dir(out_dir) {
+        for entry in entries.flatten() {
+            if let Some(name) = entry.file_name().to_str() {
+                if name.ends_with(&format!(".{ext}")) {
+                    return true;
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
samuel100: Fixed. `libs_already_present` now checks specifically for the expected core library filename (`foundry_local_core.dll` / `libfoundry_local_core.so` / `libfoundry_local_core.dylib`) instead of any native lib.

.github/workflows/build-rust-steps.yml
```diff
@@ -0,0 +1,110 @@
...
+
```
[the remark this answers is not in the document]
samuel100: Fixed. Added `--include-ignored` to the CI test command to align with the `#[ignore]` test annotation strategy.

samples/rust/tool-calling-foundry-local/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    if !model.is_cached().await? {
+        println!("Downloading model...");
+        model
+            .download(Some(|progress: &str| {
+                print!("\r  {progress:.1}%");
+                io::stdout().flush().ok();
+            }))
+            .await?;
+        println!();
```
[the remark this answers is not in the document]
samuel100: Fixed. Changed `{progress:.1}%` to `{progress}%`.

sdk_v2/rust/src/detail/core_interop.rs
```diff
@@ -0,0 +1,470 @@
...
+        }
+        free(ptr as *mut std::ffi::c_void);
+    }
+    #[cfg(windows)]
+    {
+        extern "system" {
+            fn CoTaskMemFree(pv: *mut std::ffi::c_void);
+        }
+        CoTaskMemFree(ptr as *mut std::ffi::c_void);
+    }
```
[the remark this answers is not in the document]
samuel100: Fixed. Switched from `CoTaskMemFree` to `LocalFree` (kernel32) on Windows, matching `Marshal.FreeHGlobal` / `LocalAlloc` used by the native core.

sdk_v2/rust/tests/integration.rs
```diff
@@ -0,0 +1,1081 @@
...
+    #[tokio::test]
+    async fn should_transcribe_audio_without_streaming_with_temperature() {
+        let (mut client, model) = setup_audio_client().await;
+        client.language("en").temperature(0.0);
+
+        let response = client
+            .transcribe(&audio_file())
+            .await
+            .expect("transcribe with temperature failed");
```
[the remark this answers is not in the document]
samuel100: Fixed. Changed temperature from `0.0` to `0.5` in the `with_temperature` transcription test.

sdk_v2/rust/tests/integration.rs
```diff
@@ -0,0 +1,1081 @@
...
+        let (mut client, model) = setup_audio_client().await;
+        client.language("en").temperature(0.0);
+
+        let mut full_text = String::new();
+
+        let mut stream = client
+            .transcribe_streaming(&audio_file())
+            .await
+            .expect("transcribe_streaming with temperature setup failed");
+
```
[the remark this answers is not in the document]
samuel100: Fixed. Changed temperature from `0.0` to `0.5` in the streaming `with_temperature` transcription test.

sdk_v2/rust/src/foundry_local_manager.rs
```diff
@@ -0,0 +1,109 @@
...
+            _model_load_manager: model_load_manager,
+            catalog,
+            urls: std::sync::Mutex::new(Vec::new()),
+        };
+
+        // Attempt to store; if another thread raced us, return whichever won.
+        match INSTANCE.set(manager) {
+            Ok(()) => Ok(INSTANCE.get().unwrap()),
+            Err(_) => Ok(INSTANCE.get().unwrap()),
+        }
```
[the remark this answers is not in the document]
samuel100: Fixed. Replaced non-atomic `get()`+`set()` pattern with `Once` + `OnceLock` to ensure initialization runs exactly once (stable Rust compatible — `get_or_try_init` is nightly-only).

samples/rust/native-chat-completions/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+        println!("Downloading model...");
+        model
+            .download(Some(|progress: &str| {
+                print!("\r  {progress:.1}%");
+                io::stdout().flush().ok();
+            }))
+            .await?;
+        println!();
+    }
+
```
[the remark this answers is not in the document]
samuel100: Fixed. Changed `{progress:.1}` to `{progress}%`.

samples/rust/foundry-local-webserver/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    // ── 2. Download and load a model ─────────────────────────────────────
+    let model_alias = "qwen2.5-0.5b";
+    let model = manager.catalog().get_model(model_alias).await?;
+
+    if !model.is_cached().await? {
+        print!("Downloading model {model_alias}...");
+        model
+            .download(Some(move |progress: &str| {
+                print!("\rDownloading model... {progress:.1}%");
```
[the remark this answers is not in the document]
samuel100: Fixed. Changed `{progress:.1}%` to `{progress}%`.

samuel100:
## Additional fix: WinAppSDK Bootstrap auto-detection

During testing, discovered that the Rust SDK was failing to load WinML/OpenVINO execution providers because the native core's WinAppSDK bootstrapper was never being triggered.

**Root cause:** The Rust SDK pre-loaded `Microsoft.WindowsAppRuntime.Bootstrap.dll` but never set `Bootstrap = true` in the config params passed to the native `initialize` command. Without this param, the native core skips the WinAppSDK bootstrapper, which means plug-in EPs (like OpenVINO) are unavailable.

**Fix:** `CoreInterop::new` now auto-detects the Bootstrap DLL next to the resolved core library and sets `Bootstrap = true` in config params before `initialize` is called. This matches the JS SDK behavior (see `sdk_v2/js/src/detail/coreInterop.ts:56-63`).

sdk_v2/rust/src/detail/core_interop.rs
```diff
@@ -0,0 +1,470 @@
...
+        self: &Arc<Self>,
+        command: String,
+        params: Option<Value>,
+    ) -> Result<String> {
+        let this = Arc::clone(self);
+        tokio::task::spawn_blocking(move || this.execute_command(&command, params.as_ref()))
+            .await
+            .map_err(|e| FoundryLocalError::CommandExecution {
+                reason: format!("task join error: {e}"),
+            })?
```
nenad1002, line 410: Ah, this is swallowed by tokio, if you use Claude ask it to return an error out, then change the method execute_command_async to return Result<String, FoundryLocalError>.

This will force you to return Result<(), FoundryLocalError> or similar on above funs. 

load() should return Error if there is an error.

sdk_v2/rust/src/detail/model_load_manager.rs
nenad1002: Once execute_command_async is fixed, this will return Result<(), FoundryLocalError>

sdk_v2/rust/src/detail/core_interop.rs
samuel100: I've restructured this so the error can no longer be swallowed.

  Rather than just fixing the consumer side, I moved the error propagation into `execute_command_streaming_channel` itself. The channel now carries `Result<String>` items instead of plain `String`, and after the FFI call completes,  any error from the native core response buffer is sent as the final channel item:

```rust
   tokio::task::spawn_blocking(move || {
       let tx_chunk = tx.clone();
       let result = this.execute_command_streaming(
           &command,
           params.as_ref(),
           move |chunk: &str| {
               let _ = tx_chunk.send(Ok(chunk.to_owned()));
           },
       );

       // Surface any error through the channel so it
       // cannot be silently swallowed.
       if let Err(e) = result {
           let _ = tx.send(Err(e));
       }
   });
```

  This eliminates the `JoinHandle<Result<String>>` from the return type entirely — it's now impossible for consumers to accidentally drop the error. `ChatCompletionStream` and `AudioTranscriptionStream` are much simpler as a result:
  no handle field, no `close()` method, and `poll_next` just forwards Ok/Err items from the channel. Net ~100 lines deleted.

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+
+Primary entry point for interacting with Foundry Local. Singleton — created once via `create()`.
+
+```rust
+pub struct FoundryLocalManager { /* private fields */ }
+```
+
+| Method | Signature | Description |
+|--------|-----------|-------------|
+| `create` | `fn create(config: FoundryLocalConfig) -> Result<&'static Self>` | Initialise the SDK. First call creates the singleton; subsequent calls return the existing instance (config is ignored after first call). |
```
nenad1002: So as already suggested, let's go with builder pattern, it's idiomatic in Rust, and we already break the contract on this method in C++ anyway.

nenad1002: Also btw, where is logger? Did you port logger?

nenad1002: Cause in C++, we allow custom logger:
    _explicit FoundryLocalManager(Configuration configuration, ILogger* logger = nullptr);_

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+| `catalog` | `fn catalog(&self) -> &Catalog` | Access the model catalog. |
+| `urls` | `fn urls(&self) -> Result<Vec<String>>` | URLs the local web service is listening on. Empty until `start_web_service` is called. |
+| `start_web_service` | `async fn start_web_service(&self) -> Result<Vec<String>>` | Start the local web service and return listening URLs. |
```
nenad1002: Afaik, this shouldn't return anything, why do you return listening URLs?
GetUrls() method returns the list of urls unless if the API changed recently, and you pass urls through configuration.

Either way, returning urls can probably be [&str], cause you set them beforehand an FL Manager and is owning. You are here doing unefficient clone to return this as FL manager has to own the urls to move them in the first place.

Write to Claude:
urls can almost certainly be made owned by the manager, can you make GetUrls method that returns [&str] and make StartServer not to return anything except Result that might be needed?

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+
+| Method | Signature | Description |
+|--------|-----------|-------------|
+| `new` | `fn new(app_name: impl Into<String>) -> Self` | Create a new configuration. All fields default to `None`. |
+
+**Example:**
+```rust
+let config = FoundryLocalConfig {
+    log_level: Some(LogLevel::Debug),
+    ..FoundryLocalConfig::new("my_app")
```
nenad1002: This will use build pattern in future, so examples would slightly change to be idiomatic

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+
+```rust
+pub struct Catalog { /* private fields */ }
+```
+
+| Method | Signature | Description |
+|--------|-----------|-------------|
+| `name` | `fn name(&self) -> &str` | Catalog name as reported by the native core. |
+| `update_models` | `async fn update_models(&self) -> Result<()>` | Refresh catalog if cache expired or invalidated. |
+| `get_models` | `async fn get_models(&self) -> Result<Vec<Arc<Model>>>` | Return all known models. |
```
nenad1002: ListModels** 
This is the method we don't want to change name. Vec<Arc<Model>> is probably correct, I don't think we can do better.

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+| `get_model` | `async fn get_model(&self, alias: &str) -> Result<Arc<Model>>` | Look up a model by alias. |
+| `get_model_variant` | `async fn get_model_variant(&self, id: &str) -> Result<Arc<ModelVariant>>` | Look up a variant by unique id. |
+| `get_cached_models` | `async fn get_cached_models(&self) -> Result<Vec<Arc<ModelVariant>>>` | Return only variants cached on disk. |
+| `get_loaded_models` | `async fn get_loaded_models(&self) -> Result<Vec<String>>` | Return ids of models currently loaded in memory. |
```
nenad1002: Should return ModelVariant, not String, unless again, my API is out of date.

nenad1002: @prathikr , can you confirm this?

sdk_v2/rust/src/error.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// An I/O error occurred.
+    #[error("I/O error: {0}")]
+    Io(#[from] std::io::Error),
+    /// An internal SDK error (e.g. poisoned lock).
+    #[error("internal error: {reason}")]
+    Internal { reason: String },
+}
+
+/// Convenience alias used throughout the SDK.
+pub type Result<T> = std::result::Result<T, FoundryLocalError>;
```
nenad1002, line 36: The internal alias in error.rs is fine for our own code, but remove pub use self::error::Result from lib.rs. Users should use Result<T, FoundryLocalError> explicitly. Otherwise we shadow standard Result. 

Let's be very explicit in public API. Not Result<()>, but Result<(), FoundryLocalError>

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+| `id` | `fn id(&self) -> &str` | Unique identifier of the selected variant. |
+| `variants` | `fn variants(&self) -> &[ModelVariant]` | All variants in this model. |
+| `selected_variant` | `fn selected_variant(&self) -> &ModelVariant` | Currently selected variant. |
+| `select_variant` | `fn select_variant(&mut self, id: &str) -> Result<()>` | Select a variant by id. |
+| `is_cached` | `async fn is_cached(&self) -> Result<bool>` | Whether the selected variant is cached on disk. |
+| `is_loaded` | `async fn is_loaded(&self) -> Result<bool>` | Whether the selected variant is loaded in memory. |
+| `download` | `async fn download<F>(&self, progress: Option<F>) -> Result<()>` | Download the selected variant. `F: FnMut(&str) + Send + 'static` |
+| `path` | `async fn path(&self) -> Result<PathBuf>` | Local file-system path of the selected variant. |
+| `load` | `async fn load(&self) -> Result<()>` | Load the selected variant into memory. |
+| `unload` | `async fn unload(&self) -> Result<String>` | Unload the selected variant from memory. |
```
nenad1002: Result<(), FoundryLocalError>

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+| `remove_from_cache` | `async fn remove_from_cache(&self) -> Result<String>` | Remove the selected variant from the local cache. |
```
nenad1002: Result<(), FoundryLocalError>

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+pub struct Model { /* private fields */ }
+```
+
+| Method | Signature | Description |
+|--------|-----------|-------------|
+| `alias` | `fn alias(&self) -> &str` | Alias shared by all variants. |
+| `id` | `fn id(&self) -> &str` | Unique identifier of the selected variant. |
+| `variants` | `fn variants(&self) -> &[ModelVariant]` | All variants in this model. |
+| `selected_variant` | `fn selected_variant(&self) -> &ModelVariant` | Currently selected variant. |
+| `select_variant` | `fn select_variant(&mut self, id: &str) -> Result<()>` | Select a variant by id. |
```
nenad1002: Let's make this select_variant(&self, id: &str)

And write to AI to use interior mutability to achieve this. Selected variant is not really a state we need to expose mutability for. This goes for both Model & ModelVariant.

sdk_v2/rust/src/model_variant.rs
```diff
@@ -0,0 +1,143 @@
...
+use crate::detail::core_interop::CoreInterop;
+use crate::detail::ModelLoadManager;
+use crate::error::Result;
+use crate::openai::AudioClient;
+use crate::openai::ChatClient;
+use crate::types::ModelInfo;
+
+/// Represents one specific variant of a model (a particular id within an alias
+/// group).
+#[derive(Debug, Clone)]
```
nenad1002: You are leaking internal impl with Debug here. Debug gives ability to print contents of the object.

Consider implementing custom _impl fmt::Debug for ModelVariant_

And only returning info. Same for Model.

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+---
+
+### AudioTranscriptionResponse
+
+```rust
+pub struct AudioTranscriptionResponse {
+    pub text: String,                              // The transcribed text
+    pub language: Option<String>,                  // Language of input audio (if detected)
+    pub duration: Option<f64>,                     // Duration in seconds (if available)
+    pub segments: Option<Vec<serde_json::Value>>,  // Transcription segments (if available)
```
nenad1002: Dont return serde_json in public API,

do use string or something like this


    __pub segments: Option<Vec<TranscriptionSegment>>

    pub words: Option<Vec<TranscriptionWord>>__

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+    pub model_type: String,
+    pub prompt_template: Option<PromptTemplate>,
+    pub publisher: Option<String>,
+    pub model_settings: Option<ModelSettings>,
+    pub license: Option<String>,
+    pub license_description: Option<String>,
+    pub cached: bool,
+    pub task: Option<String>,
+    pub runtime: Option<Runtime>,
+    pub file_size_mb: Option<f64>,
```
nenad1002: should be u64 (measuring in bytes) or if you want to keep f64, then have check in place so it cannot go negative

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+
+### ModelInfo
+
+Full metadata for a model variant as returned by the catalog.
+
+```rust
+pub struct ModelInfo {
+    pub id: String,
+    pub name: String,
+    pub version: i64,
```
nenad1002: Same here, u64

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+    pub license: Option<String>,
+    pub license_description: Option<String>,
+    pub cached: bool,
+    pub task: Option<String>,
+    pub runtime: Option<Runtime>,
+    pub file_size_mb: Option<f64>,
+    pub supports_tool_calling: Option<bool>,
+    pub max_output_tokens: Option<i64>,
+    pub min_fl_version: Option<String>,
+    pub created_at_unix: i64,
```
nenad1002: u64

sdk_v2/rust/docs/api.md
```diff
@@ -0,0 +1,502 @@
...
+Discovers, caches, and looks up available models.
+
+```rust
+pub struct Catalog { /* private fields */ }
+```
+
+| Method | Signature | Description |
+|--------|-----------|-------------|
+| `name` | `fn name(&self) -> &str` | Catalog name as reported by the native core. |
+| `update_models` | `async fn update_models(&self) -> Result<()>` | Refresh catalog if cache expired or invalidated. |
```
nenad1002: Again, @prathikr can give feedback here, but I don't think this method is public

sdk_v2/rust/docs/api.md
samuel100: Fixed in b6381d2. `FoundryLocalConfig` now uses a builder pattern with chainable setters. Also added `Logger` trait stub.

sdk_v2/rust/docs/api.md
samuel100: Fixed in b6381d2. `start_web_service()` now returns `Result<(), FoundryLocalError>`. URLs are accessed separately via `manager.urls()?`. Updated the webserver sample accordingly.

sdk_v2/rust/docs/api.md
samuel100: Fixed in b6381d2. `FoundryLocalConfig` now uses builder pattern with chainable setters — fields are private.

sdk_v2/rust/docs/api.md
samuel100: Acknowledged — `get_models()` name and `Vec<Arc<Model>>` return type kept as-is.

sdk_v2/rust/docs/api.md
samuel100: Fixed in b6381d2. `get_loaded_models()` now returns `Vec<Arc<ModelVariant>>` — resolves loaded IDs against the catalog's known variants.

sdk_v2/rust/src/error.rs
samuel100: Fixed in b6381d2. Removed `pub use self::error::Result` from `lib.rs`. The alias remains in `error.rs` for internal use only. Public API signatures use `Result<T, FoundryLocalError>` explicitly.

sdk_v2/rust/docs/api.md
samuel100: Fixed in b6381d2 — all public signatures now use explicit `Result<T, FoundryLocalError>`.

sdk_v2/rust/docs/api.md
samuel100: Fixed in b6381d2.

sdk_v2/rust/docs/api.md
samuel100: Fixed in b6381d2. `select_variant` now takes `&self` using interior mutability — `selected_index` is an `AtomicUsize` with `Relaxed` ordering. Since `Model` is behind `Arc`, this is safe and avoids requiring `&mut self`.

sdk_v2/rust/src/model_variant.rs
samuel100: Fixed in b6381d2. Both `Model` and `ModelVariant` now have custom `fmt::Debug` impls that expose only public info (alias, id, variants count) — internal fields like `core_interop` and `model_load_manager` are no longer leaked.

sdk_v2/rust/docs/api.md
samuel100: Fixed in b6381d2. Replaced `serde_json::Value` with typed `TranscriptionSegment` and `TranscriptionWord` structs. Both are exported from the crate root.

sdk_v2/rust/docs/api.md
samuel100: Fixed in b6381d2. `file_size_mb` is now `Option<u64>`. Kept the field name as `file_size_mb` for consistency with the C# SDK (`FileSizeMb`) and native core JSON format.

sdk_v2/rust/docs/api.md
samuel100: Fixed in b6381d2. `max_output_tokens` changed to `Option<u64>`.

sdk_v2/rust/docs/api.md
samuel100: Fixed in b6381d2. `version` and `created_at_unix` changed to `u64`.

sdk_v2/rust/docs/api.md
samuel100: Fixed in b6381d2. Removed `pub use self::detail::ModelLoadManager` from `lib.rs` — it's no longer part of the public API.

sdk_v2/rust/src/detail/core_interop.rs
samuel100: The error propagation was restructured in a prior commit. This latest push (b6381d2) also updates all public APIs to return explicit `Result<T, FoundryLocalError>` as part of the Result re-export removal.

sdk_v2/rust/src/foundry_local_manager.rs
samuel100: Fixed in b6381d2. `start_web_service` now returns `Result<(), FoundryLocalError>`. All public APIs use explicit Result types — the `Result` alias is no longer re-exported.

sdk_v2/rust/src/detail/core_interop.rs
samuel100: Already addressed in prior commit (u32 for ResponseBuffer lengths). Confirmed still in place.

PR #525, opened 2026-03-16
streamline js examples

samuel100:
Simplify the samples so that users only need to:

```bash
npm run setup
node app.js
```

The setup script detects if the machine is Windows and installs FLC winml.


samples/js/copilot-sdk-foundry-local/setup.cjs
```diff
@@ -0,0 +1,11 @@
+// Cross-platform dependency installer
```
baijumeswani, line 1: this script seems repeated for each example. can we have a common file and parametrtize the example we want to build as an argument to setup?

PR #601, opened 2026-04-06
Add private model catalog SDK support (AddCatalog, SelectCatalog, GetCatalogNames)

kobby-kobbs:
## Summary
- Adds SDK support for registering and managing private model catalogs (MDS integration)
- Exposes three new async methods on `ICatalog`: `AddCatalogAsync`, `SelectCatalogAsync`, `GetCatalogNamesAsync`
- Removes logging of command input to prevent leaking credentials in debug logs

### `sdk/cs/src/ICatalog.cs`
Adds three new interface methods with full XML documentation:
- `AddCatalogAsync` — Register a private catalog with OAuth2/bearer auth support
- `SelectCatalogAsync` — Filter models to a specific catalog (or null to show all)
- `GetCatalogNamesAsync` — List all registered catalog names

### `sdk/cs/src/Catalog.cs`
Implements the three new methods with:
- Input validation (URI scheme check, token endpoint URL validation, null/whitespace guards)
- `forceRefresh` parameter on `UpdateModels` so Add/Select immediately refresh the model list
- Null-safe deserialization in `GetCatalogNamesAsync`

### `sdk/cs/src/Detail/CoreInterop.cs`
Removes logging of `commandInput` in error paths to prevent leaking credentials (client secrets, bearer tokens) in debug logs.

### `sdk/cs/src/Detail/JsonSerializationContext.cs`
Registers `List<string>` for AOT-compatible JSON serialization of catalog names.

### `sdk/cs/test/FoundryLocal.Tests/CatalogManagementTests.cs`
New test class with two tests:
- `Test_AddAndSelectCatalog` — Verifies add + select + reset flow
- `Test_GetCatalogNames` — Verifies catalog name retrieval and deserialization


sdk/cs/src/Catalog.cs
```diff
@@ -190,10 +190,10 @@ private async Task<IModel> GetLatestVersionImplAsync(IModel modelOrModelVariant,
         return latest.Id == modelOrModelVariant.Id ? modelOrModelVariant : latest;
     }
 
-    private async Task UpdateModels(CancellationToken? ct)
+    private async Task UpdateModels(CancellationToken? ct, bool forceRefresh = false)
```
baijumeswani: There is a function for invalidating the cache: `InvalidateCache`. Could you use that instead of adding forceRefresh?

sdk/cs/src/Catalog.cs
```diff
@@ -249,4 +249,104 @@ public void Dispose()
     {
         _lock.Dispose();
     }
+
+    public async Task AddCatalogAsync(string name, Uri uri, string? clientId = null,
+                                      string? clientSecret = null, string? bearerToken = null,
+                                      string? tokenEndpoint = null, string? audience = null,
```
baijumeswani: Can we keep only the essential arguments in the function and move all the optional arguments in a map argument.
Different catalogs may have different arguments that may be needed. So, the basic ones can be explicit arguments and the optional ones can be in a map?

sdk/cs/src/Catalog.cs
```diff
@@ -249,4 +249,104 @@ public void Dispose()
...
+                {
+                    ["Name"] = name,
+                    ["Uri"] = uri.ToString(),
+                    ["ClientId"] = clientId ?? "",
+                    ["ClientSecret"] = clientSecret ?? "",
+                    ["BearerToken"] = bearerToken ?? "",
+                    ["TokenEndpoint"] = tokenEndpoint ?? "",
+                    ["Audience"] = audience ?? ""
+                }
+            };
```
baijumeswani: Can all of this logic be moved to Core so each sdk need not have this logic?

sdk/cs/src/Catalog.cs
kobby-kobbs: I reverted to the original signature, it now calls `InvalidateCache` in `UpdateModels` in both `AddCatalogAsync` and `SelectCatalogAsync`

sdk/cs/src/Catalog.cs
kobby-kobbs: I see, I changed the signature to AddCatalogAsync(string name, Uri uri, Dictionary<string, string>? options = null, CancellationToken? ct = null)

sdk/cs/src/Catalog.cs
kobby-kobbs: I think the actual catalog registration logic already lives in Core, the SDK calls the native add_catalog interop command which handles it. What is left in the SDK is just input validation (C# type safety), parameter marshaling into the interop request, and refreshing the local model cache afterward. These are all SDK-specific concerns that each SDK needs to handle in its own language, so there isn't much to move.

PR #612, opened 2026-04-08
Add Nemotron-ASR streaming inference to Python SDK

rui-ren:
### Description

Adds real-time audio streaming support to the Foundry Local Python SDK, enabling live microphone-to-text transcription via ONNX Runtime GenAI's StreamingProcessor API (Nemotron ASR).

This is the Python port of C# PR #485 with full feature parity. The existing `AudioClient` only supports file-based transcription. This PR introduces `LiveAudioTranscriptionSession` that accepts continuous PCM audio chunks (e.g., from a microphone) and returns partial/final transcription results as a synchronous generator.

### What's included

**New files**
- `src/openai/live_audio_transcription_client.py` — Streaming session with `start()`, `append()`, `get_transcription_stream()`, `stop()`
- `src/openai/live_audio_transcription_types.py` — `LiveAudioTranscriptionResponse` (ConversationItem-shaped), `LiveAudioTranscriptionOptions`, `CoreErrorResponse`, `TranscriptionContentPart`
- `test/openai/test_live_audio_transcription.py` — 22 unit tests for deserialization, settings, state guards, streaming pipeline
- `test/openai/test_live_audio_transcription_e2e.py` — E2E test with real native DLLs and nemotron model
- `test/openai/conftest.py` — DLL preload for E2E tests
- `samples/python/live-audio-transcription/src/app.py` — Live microphone transcription demo

**Modified files**
- `src/openai/audio_client.py` — Added `create_live_transcription_session()` factory method
- `src/detail/core_interop.py` — Added `StreamingRequestBuffer` struct, `execute_command_with_binary()`, `start_audio_stream`, `push_audio_data`, `stop_audio_stream` methods, and `_load_dll_win()` for robust DLL loading on Windows
- `src/openai/__init__.py` — Exported new live transcription types
- `test/conftest.py` — Pre-load ORT/GenAI DLLs before brotli import to avoid Windows DLL search conflicts

### API surface

```python
audio_client = model.get_audio_client()
session = audio_client.create_live_transcription_session()

session.settings.sample_rate = 16000
session.settings.channels = 1
session.settings.language = "en"

session.start()

# Push audio from microphone callback (thread-safe)
session.append(pcm_bytes)

# Read results as synchronous generator
for result in session.get_transcription_stream():
    print(result.content[0].text)

session.stop()
```

### C# parity

| C# API | Python API | Notes |
|---|---|---|
| `CreateLiveTranscriptionSession()` | `create_live_transcription_session()` | ✅ |
| `StartAsync(ct)` | `start()` | Sync (matches Python SDK convention) |
| `AppendAsync(ReadOnlyMemory<byte>, ct)` | `append(bytes)` | Thread-safe, copies data |
| `GetTranscriptionStream()` | `get_transcription_stream()` | Generator (sync equivalent of IAsyncEnumerable) |
| `StopAsync(ct)` | `stop()` | Drains push queue, sends native stop, surfaces final result |
| `IAsyncDisposable` | Context manager (`with`) | Idiomatic Python equivalent |
| `LiveAudioTranscriptionOptions` | `LiveAudioTranscriptionOptions` | Same fields: sample_rate, channels, bits_per_sample, language, push_queue_capacity |
| `LiveAudioTranscriptionResponse` | `LiveAudioTranscriptionResponse` | ConversationItem-shaped: content[0].text/transcript, is_final, start_time, end_time |

### Design highlights
- **Output type alignment** — `LiveAudioTranscriptionResponse` uses the OpenAI Realtime `ConversationItem` shape (`content[0].text/transcript`) for forward compatibility
- **Internal push queue** — Bounded `queue.Queue` serializes audio pushes from any thread (safe for mic callbacks) with backpressure
- **Fail-fast on errors** — Push loop terminates immediately on any native error (no retry logic)
- **Settings freeze** — Audio format settings are snapshot-copied at `start()` and immutable during the session
- **Buffer copy** — `append()` copies input data to avoid issues with callers reusing buffers (e.g., PyAudio)
- **Routes through existing exports** — `start_audio_stream` and `stop_audio_stream` route through `execute_command`; `push_audio_data` routes through `execute_command_with_binary` — no new native entry points required
- **DLL loading fix** — Uses `LoadLibraryExW` with `LOAD_WITH_ALTERED_SEARCH_PATH` on Windows to prevent conflicts with stale system-level ORT DLLs

### Verified working
- ✅ 22 unit tests passing (deserialization, settings, state guards, streaming pipeline with mocked core)
- ✅ E2E test passing (SDK → Core.dll → onnxruntime-genai.dll → onnxruntime.dll with nemotron model)
- ✅ Full session lifecycle: start → push synthetic PCM → stop → verify results
- ✅ Existing tests unaffected


sdk/python/src/detail/core_interop.py
```diff
@@ -173,6 +221,17 @@ def _initialize_native_libraries() -> 'NativeBinaryPaths':
                                                       ctypes.c_void_p]  # user_data
         lib.execute_command_with_callback.restype = None
 
+        # execute_command_with_binary is required for audio streaming but may
+        # not be present in older Core builds.  Register it if available;
+        # the method will raise AttributeError at call-time if missing.
```
kunal-vaishnavi: The FL Core version is updated so this should always exist.

sdk/python/src/openai/live_audio_transcription_client.py
```diff
@@ -0,0 +1,310 @@
...
+            self._session_handle = response.data
+            if self._session_handle is None:
+                raise FoundryLocalException(
+                    "Native core did not return a session handle."
+                )
+
+            self._started = True
+            self._stopped = False
+
+            # Start the push loop in a daemon thread
```
kunal-vaishnavi: Do we need to use a daemon thread? Should we make this non-daemon and require `stop()` + `join()` instead? One of the features of FL Core that we have [advertised](https://devblogs.microsoft.com/foundry/foundry-local-ga/) is that it does not require a daemon.

sdk/python/test/openai/test_live_audio_transcription_e2e.py
```diff
@@ -0,0 +1,303 @@
...
+from foundry_local_sdk.openai.live_audio_transcription_types import (
+    LiveAudioTranscriptionResponse,
+)
+
+
+def _get_e2e_test_pkgs_path() -> Path:
+    """Locate the e2e-test-pkgs directory."""
+    current = Path(__file__).resolve().parent
+    while True:
+        candidate = current / "samples" / "python" / "e2e-test-pkgs"
```
kunal-vaishnavi: This looks like a similar function to what is inside `conftest.py`. Can either this or `conftest.py` be removed then?

sdk/python/test/openai/conftest.py
```diff
@@ -0,0 +1,71 @@
...
+        if candidate.exists():
+            return candidate
+        parent = current.parent
+        if parent == current:
+            break
+        current = parent
+    raise FileNotFoundError("e2e-test-pkgs not found")
+
+
+if sys.platform.startswith("win"):
```
kunal-vaishnavi: Given there already exists a `conftest.py` file, can we move any modifications to that file and remove this one?

sdk/python/test/openai/test_live_audio_transcription_e2e.py
```diff
@@ -0,0 +1,303 @@
...
+    ),
+    pytest.mark.skipif(
+        not sys.platform.startswith("win"),
+        reason="E2E test requires Windows (DLLs are .dll files)",
+    ),
+]
+
+
+def _preload_and_init():
+    """Pre-load DLLs from e2e-test-pkgs and initialize the SDK.
```
kunal-vaishnavi: Can we move this function into `conftest.py` instead so that the E2E unit tests are cleaner to look at?

sdk/python/src/detail/core_interop.py
rui-ren: Yes, thank you. get rid of the `try ... except ...`

sdk/python/src/openai/live_audio_transcription_client.py
rui-ren: Thanks.

sdk/python/test/openai/test_live_audio_transcription_e2e.py
rui-ren: 👍

sdk/python/src/detail/core_interop.py
rui-ren: @kunal-vaishnavi 
The CI got this error, seems like we need to update the 

the macOS CI runner still has an older Core.dylib that doesn't export `execute_command_with_binary`

```
/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/foundry_local_sdk/foundry_local_manager.py:53: in initialize
    FoundryLocalManager(config)
/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/foundry_local_sdk/foundry_local_manager.py:65: in __init__
    self._initialize()
/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/foundry_local_sdk/foundry_local_manager.py:75: in _initialize
    self._core_interop = CoreInterop(self.config)
                         ^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/foundry_local_sdk/detail/core_interop.py:242: in __init__
    paths = CoreInterop._initialize_native_libraries()
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/foundry_local_sdk/detail/core_interop.py:224: in _initialize_native_libraries
    lib.execute_command_with_binary.argtypes = [ctypes.POINTER(StreamingRequestBuffer),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/ctypes/__init__.py:392: in __getattr__
    func = self.__getitem__(name)
           ^^^^^^^^^^^^^^^^^^^^^^
```

sdk/python/src/detail/core_interop.py
```diff
@@ -108,6 +125,32 @@ class CoreInterop:
...
+        _LOAD_WITH_ALTERED_SEARCH_PATH = 0x00000008
+
+        handle = kernel32.LoadLibraryExW(dll_path, None, _LOAD_WITH_ALTERED_SEARCH_PATH)
+        if not handle:
+            logger.warning("LoadLibraryExW failed for %s (error %d), falling back to ctypes.CDLL",
+                           dll_path, kernel32.GetLastError())
+            return ctypes.CDLL(dll_path)
+
+        # DLL is now mapped; ctypes.CDLL will reuse the loaded module
+        return ctypes.CDLL(dll_path)
```
baijumeswani: Can you explain why this code is needed? The Python SDK 1.0.0 works well on windows. I am not sure what changed that would warrant this library loading code here.

sdk/python/src/detail/core_interop.py
```diff
@@ -142,15 +185,20 @@ def _initialize_native_libraries() -> 'NativeBinaryPaths':
             for native_dir in paths.all_dirs():
                 os.add_dll_directory(str(native_dir))
 
+            # Set the DLL search directory so that when ORT/GenAI load their
+            # own dependencies, they find sibling DLLs from the correct
+            # directory rather than stale copies in system directories.
+            ctypes.windll.kernel32.SetDllDirectoryW(str(paths.ort_dir))
```
baijumeswani: This should also not be needed. Ort and OrtGenAI should have their code in a way such that they load their dependencies correctly.

samples/python/live-audio-transcription/src/app.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+)
+print()
+print(f"Loading model {model.id}...", end="")
+model.load()
+print("done.")
+
+# Create a live transcription session
+audio_client = model.get_audio_client()
+session = audio_client.create_live_transcription_session()
```
baijumeswani: I am wondering why we have such overly explicit names like `create_livea_transcription_session()`? Would it be better to have a name like `audio_client.create_session()` and the client should have enough context on what kind of session to create give the model type?

samples/python/live-audio-transcription/src/app.py
```diff
@@ -0,0 +1,117 @@
...
+session.settings.sample_rate = 16000
```
baijumeswani: Where are these values coming from? Do application developers need to know these values ahead of time? Can we set defaults?

samples/python/live-audio-transcription/src/app.py
```diff
@@ -0,0 +1,117 @@
...
+session.settings.channels = 1
+session.settings.language = "en"
```
baijumeswani: Again, can we set defaults for these?

samples/python/live-audio-transcription/src/app.py
```diff
@@ -0,0 +1,117 @@
...
+
+session.start()
+print("       Session started")
+
+# Start reading transcription results in a background thread
+def read_results():
+    try:
+        for result in session.get_transcription_stream():
```
baijumeswani: Since this is already a transcription session, can the API be called session.stream()?

samples/python/live-audio-transcription/src/app.py
baijumeswani: Is the expectation that these settings may not be known at construction time? Can these settings be passed into the `create_live_transcription_session` function? That way, we can potentially also avoid needing a `session.start()`.

sdk/python/src/detail/core_interop.py
```diff
@@ -295,6 +347,66 @@ def execute_command_with_callback(self, command_name: str, command_input: Option
...
+        response_str = ctypes.string_at(resp.Data, resp.DataLength).decode("utf-8") if resp.Data else None
+        error_str = ctypes.string_at(resp.Error, resp.ErrorLength).decode("utf-8") if resp.Error else None
+
+        lib.free_response(resp)
+
+        return Response(data=response_str, error=error_str)
+
+    # --- Audio streaming session support ---
+
+    def start_audio_stream(self, command_input: InteropRequest) -> Response:
```
baijumeswani, line 373: Any reason these names are different from the names in FLC? Can we keep them consistent?

sdk/python/src/openai/live_audio_transcription_client.py
```diff
@@ -0,0 +1,311 @@
...
+                    "No active streaming session. Call start() first."
+                )
+
+            push_queue = self._push_queue
+            if push_queue is None:
+                raise FoundryLocalException(
+                    "No active streaming session. Call start() first."
+                )
+
+            push_queue.put(data_copy)
```
baijumeswani: What happens if we are at capacity (since this is a bounded queue)? Will this be blocking then?

sdk/python/src/openai/live_audio_transcription_client.py
```diff
@@ -0,0 +1,311 @@
...
+            if self._started:
+                raise FoundryLocalException(
+                    "Streaming session already started. Call stop() first."
+                )
+
+            # Freeze settings
+            self._active_settings = self.settings.snapshot()
+
+            self._output_queue = queue.Queue()
+            self._push_queue = queue.Queue(maxsize=self._active_settings.push_queue_capacity)
```
baijumeswani: Any reason to maintain all of this queueing logic here and not in the core?

sdk/python/src/openai/live_audio_transcription_client.py
```diff
@@ -0,0 +1,311 @@
...
+                final_result = LiveAudioTranscriptionResponse.from_json(response.data)
+                text = (final_result.content[0].text
+                        if final_result.content else "")
+                if text:
+                    self._output_queue.put(final_result)
+            except Exception as parse_ex:
+                logger.debug("Could not parse stop response as transcription result: %s", parse_ex)
+
+        # 4. Complete the output queue
+        self._output_queue.put(_SENTINEL)
```
baijumeswani, line 248: You're putting a _SENTINEL on the output queue and then you're also making it = None? Is this thread safe? Do you need to acquire a lock here?

sdk/python/src/openai/live_audio_transcription_client.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def stop(self) -> None:
+        """Signal end-of-audio and stop the streaming session.
+
+        Any remaining buffered audio in the push queue will be drained to
+        native core first.  Final results are delivered through
+        :meth:`get_transcription_stream` before it completes.
+        """
+        with self._lock:
+            if not self._started or self._stopped:
```
baijumeswani, line 218: Why do you need both started and stopped? Can you not track it using just one?

sdk/python/test/openai/test_live_audio_transcription.py
```diff
@@ -0,0 +1,380 @@
...
+        )
+
+        session = LiveAudioTranscriptionSession("test-model", mock_interop)
+        session.start()
+
+        session.append(b'\x00' * 3200)
+
+        # Give push loop time to process
+        import time
+        time.sleep(0.5)
```
baijumeswani: Why does push loop need time to process?

sdk/python/test/conftest.py
```diff
@@ -14,6 +14,54 @@
 
 import os
 import logging
+import sys
+
+# ---------------------------------------------------------------------------
+# Pre-load ORT/GenAI DLLs from e2e-test-pkgs BEFORE importing the SDK.
```
baijumeswani: What is the need for all of this code?

sdk/python/src/detail/core_interop.py
rui-ren: You are right, as for my local `dev`, I include `ORT/ORT-genai.dll` locally.

I will delete `_local_dll_win` and this `core_interops.py`. Thanks

sdk/python/src/detail/core_interop.py
rui-ren: This is for my local dev to resolve the conflict of local `DLLs` loading. 

I will remove and keep it as a robustness measure.

samples/python/live-audio-transcription/src/app.py
rui-ren: The name matches the C# SDK exactly (`CreateLiveTranscriptionSession()` in `AudioClient.cs`). The Python SDK was designed for strict feature parity with the C# PR #485. 

If we want to simplify the name, I can do it across both C# and Python SDKs together to keep them consistent

samples/python/live-audio-transcription/src/app.py
rui-ren: Defaults already exist in LiveAudioTranscriptionOptions:
```
  - sample_rate = 16000
  - channels = 1
  - bits_per_sample = 16
  - language = None
```
 The sample sets them explicitly for clarity (showing developers what values to use), but they're all optional — if you omit them, the defaults apply. I can remove the explicit settings from the sample to make this clearer.

samples/python/live-audio-transcription/src/app.py
rui-ren: Yes, we have the default value in  `LiveAudioTranscriptionOptions`
```
  - sample_rate = 16000
  - channels = 1
  - bits_per_sample = 16
  - language = None
 ```

File: samples/python/live-audio-transcription/src/app.py [hunk 19]
```diff
@@ -0,0 +1,117 @@
+# Live Audio Transcription — Foundry Local SDK Example (Python)
```
bmehta001: Updates samples/python/README to reflect this new sample

sdk/python/src/detail/core_interop.py
rui-ren: The Python method names (`start_audio_stream`, `push_audio_data`, `stop_audio_stream`) match the C# SDK's CoreInterop method names (`StartAudioStream`, `PushAudioData`, `StopAudioStream`) — just
 converted to Python snake_case convention. Internally, they route to the native commands `audio_stream_start`, `audio_stream_push`, `audio_stream_stop`. The naming follows the same pattern
 as the rest of the SDK where method names are SDK-facing and command strings are internal.

sdk/python/src/openai/live_audio_transcription_client.py
rui-ren: Yes, `queue.Queue.put()` blocks when the queue is full. This is intentional backpressure — it prevents unbounded memory growth if the native core processes audio slower than real-time.
 This matches the C# design which uses `BoundedChannelFullMode.Wait` (the `AppendAsync` call awaits until space is available).

 The default capacity is 100 chunks (~3 seconds of audio at typical mic callback sizes), which is configurable via `session.settings.push_queue_capacity`.

sdk/python/test/conftest.py
rui-ren: This is test infrastucture only, I will delete it.

samples/python/live-audio-transcription/src/app.py
rui-ren: Yes, I moved the sample code to another PR. 
https://github.com/microsoft/Foundry-Local/pull/672

sdk/python/src/openai/live_audio_transcription_client.py
```diff
@@ -0,0 +1,311 @@
...
+                logger.debug("Could not parse stop response as transcription result: %s", parse_ex)
+
+        # 4. Complete the output queue
+        self._output_queue.put(_SENTINEL)
+
+        # 5. Clean up — set _output_queue to None so subsequent calls to
+        # get_transcription_stream() fail fast instead of hanging.
+        self._session_handle = None
+        self._started = False
+        self._output_queue = None
```
bmehta001: The C# + JS SDK keep the completed stream readable instead of making it inaccessible, so maybe we should do the same here

samples/python/live-audio-transcription/src/app.py
```diff
@@ -0,0 +1,117 @@
...
+def read_results():
+    try:
+        for result in session.get_transcription_stream():
+            text = result.content[0].text if result.content else ""
+            if result.is_final:
+                print()
+                print(f"  [FINAL] {text}")
+            elif text:
+                print(f"\033[96m{text}\033[0m", end="", flush=True)
+    except Exception:
```
bmehta001: Do you think we should print exceptions if they are raised to make it easier for users to debug?

rui-ren: Yes, I will add the exception error messages. Thanks

sdk/python/src/openai/live_audio_transcription_client.py
rui-ren: removed.

sdk/python/test/conftest.py
```diff
@@ -24,6 +25,20 @@
 
 logger = logging.getLogger(__name__)
 
+
+def _get_e2e_test_pkgs_path():
```
kunal-vaishnavi: Why are any changes needed in this file? Can we just do a quick check for whether the path exists on the machine? These changes seem like they are being done for local testing purposes instead of SDK changes.

rui-ren: removed, thank you!

sdk/python/test/openai/test_live_audio_transcription.py
rui-ren: You are right, we do not need this `time.sleep(0.5)`, the queue will block until data arrives.

sdk/python/src/openai/live_audio_transcription_client.py
rui-ren: Both flags are needed to track distinct states: `_started=False` means the session was never started (reject all operations), while `_stopped=True` means it was running and has been cleaned up. Using just one flag,
 `stop()` couldn't distinguish "never started" from "already stopped" - which matters because cleanup (draining the push queue, calling native stop) should only run once. This matches the C# implementation which also tracks both `_started` and `_stopped`.

sdk/python/src/openai/live_audio_transcription_client.py
rui-ren: resolved, thanks!

sdk/python/src/openai/live_audio_transcription_client.py
rui-ren: The queues handle SDK-level threading concerns that don't belong in the Core:


  - Push queue: serializes audio chunks from any thread (e.g. mic callbacks) into a single thread that calls the synchronous native `audio_stream_push` FFI. The native call isn't thread-safe.
  - Output queue: bridges results from the background push loop thread to the user's generator on the main thread.

 This matches the C# SDK which uses Channel<T> for both (`_pushChannel` and `_outputChannel` in `LiveAudioTranscriptionClient.cs`). The Core's FFI layer is synchronous - it's the SDK's job to provide the async/threaded
 wrapper.

sdk/python/src/openai/live_audio_transcription_client.py
baijumeswani: Please document this in the code for all SDKs

sdk/python/src/openai/live_audio_transcription_client.py
baijumeswani: Sorry, I am not trying to suggest that it does not match C#. I am trying to see if the logic could be simplified across all SDKs.

sdk/python/requirements.txt
```diff
@@ -2,7 +2,7 @@ pydantic>=2.0.0
 requests>=2.32.4
 openai>=2.24.0
 # Standard native binary packages from the ORT-Nightly PyPI feed.
-foundry-local-core==1.0.0rc1
+foundry-local-core==1.0.0.dev202604082039
```
baijumeswani: Why is this change necessary? Please avoid changing to internal dev build since main should only refer to released packages.

rui-ren: ok, sure, thank you, that was testing code, update now.

kunal-vaishnavi (dismissed):

sdk/deps_versions.json
```diff
@@ -1,7 +1,7 @@
 {
   "foundry-local-core": {
     "nuget": "1.0.0",
-    "python": "1.0.0"
+    "python": "1.0.0rc1"
```
baijumeswani: Please keep this at 1.0.0

sdk/deps_versions_winml.json
```diff
@@ -1,7 +1,7 @@
 {
   "foundry-local-core": {
     "nuget": "1.0.0",
-    "python": "1.0.0"
+    "python": "1.0.0rc1"
```
baijumeswani: Same for this.

sdk/python/test/openai/test_live_audio_transcription_e2e.py
```diff
@@ -0,0 +1,299 @@
...
+        candidate = current / "samples" / "python" / "e2e-test-pkgs"
+        if candidate.exists():
+            return candidate
+        parent = current.parent
+        if parent == current:
+            return None
+        current = parent
+
+
+def _has_e2e_assets() -> bool:
```
baijumeswani: Is this debug code?

sdk/deps_versions.json
rui-ren: Yes, updating. Thanks!

sdk/deps_versions_winml.json
rui-ren: Updating, thanks!

sdk/python/test/openai/test_live_audio_transcription_e2e.py
```diff
@@ -0,0 +1,299 @@
...
+    ),
+    pytest.mark.skipif(
+        not sys.platform.startswith("win"),
+        reason="E2E test requires Windows (DLLs are .dll files)",
+    ),
+]
+
+
+def _preload_and_init_e2e():
+    """Load DLLs from e2e-test-pkgs and initialize the SDK for E2E tests."""
```
baijumeswani: What are e2e test packages?

sdk/python/test/openai/test_live_audio_transcription_e2e.py
rui-ren: Not debug - it's a skip guard. The whole E2E module is skipped via `pytestmark`when these DLLs/models aren't checked out (which is the default for non-E2E test runs)."

sdk/python/test/openai/test_live_audio_transcription_e2e.py
rui-ren: This is my local dev package, i will delete it.

baijumeswani (approved):

Issue #628, opened 2026-04-13
Microsoft.AI.Foundry.Local cannot be consumed from .NET 8.0 projects


scottaddie:
The 1.0.0 release of the `Microsoft.AI.Foundry.Local` NuGet package targets `net9.0` only:

https://github.com/microsoft/Foundry-Local/blob/d8cada4ed96203e0f91b6695643a5bcee29b382b/sdk/cs/src/Microsoft.AI.Foundry.Local.csproj#L16

This configuration prevents the package from being consumed in projects that target .NET 8.0&mdash;a supported LTS version of .NET until November 10th (per https://dotnet.microsoft.com/platform/support/policy#dotnet-core). Note that .NET 9.0 becomes unsupported at that time too. For a customer trying to use this package in a .NET 8.0 project, the acquisition experience in Visual Studio is this:

<img width="877" height="162" alt="Image" src="https://github.com/user-attachments/assets/db2405ad-c10c-4954-b33f-ba72fa3402aa" />

Consider revisiting the framework targeting strategy to accommodate more versions and flavors of .NET. For inspiration, here's how the Azure SDK team approaches this problem: https://azure.github.io/azure-sdk/dotnet_introduction.html#target-frameworks.

baijumeswani:
Thanks for bringing this to our attention and thanks for reviewing the PR as well. We will do a patch release with the fix in place (sometime next week).

scottaddie:
This limitation was addressed in https://github.com/microsoft/Foundry-Local/releases/tag/v1.1.0

PR #629, opened 2026-04-14
adds support for netstandard2.0

prathikr:
https://github.com/microsoft/Foundry-Local/issues/628


sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -88,7 +88,7 @@
       <PackageId>Microsoft.AI.Foundry.Local.WinML</PackageId>
       <AssemblyName>Microsoft.AI.Foundry.Local.WinML</AssemblyName>
       <DefineConstants>$(DefineConstants);IS_WINML</DefineConstants>
-      <TargetFramework>net9.0-windows10.0.26100.0</TargetFramework> <!-- override -->
+      <TargetFramework>net8.0-windows10.0.26100.0</TargetFramework> <!-- override -->
```
scottaddie: Note that this change still won't allow .NET Framework customers to use the NuGet package. Per the [support policy](https://learn.microsoft.com/lifecycle/products/microsoft-net-framework), .NET Framework 4.6.2 and later are still supported. If you don't want to multi-target, a single `netstandard2.0` target will allow you to support .NET Framework customers and more. See [this table](https://learn.microsoft.com/dotnet/standard/net-standard?tabs=net-standard-2-0).

scottaddie (changes requested):

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -13,7 +13,7 @@
       <RepositoryUrl>https://github.com/microsoft/Foundry-Local</RepositoryUrl>
       <RepositoryType>git</RepositoryType>
 
-      <TargetFramework>net9.0</TargetFramework>
+      <TargetFramework>net8.0</TargetFramework>
```
scottaddie: See my other comment about .NET Framework support

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
prathikr: thanks, added netstandard2.0 target

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
prathikr: thanks, added netstandard2.0 target

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
jsquire, line 49: If your intent is to use the constants that are being defined in the code to determine what platform you're running on, this approach isn't going to work.  This will tell you the build-time platform only.  You may want to consider removing these and using a runtime check.   

For example, this will cover all supported platforms:

```csharp
 using System.Runtime.InteropServices;
 
 public static class PlatformHelper
 {
     public static bool IsWindows =>
 #if NET5_0_OR_GREATER
         OperatingSystem.IsWindows();
 #else
         RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
 #endif
 
     public static bool IsLinux =>
 #if NET5_0_OR_GREATER
         OperatingSystem.IsLinux();
 #else
         RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
 #endif
 
     public static bool IsMacOS =>
 #if NET5_0_OR_GREATER
         OperatingSystem.IsMacOS();
 #else
         RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
 #endif
 
     public static string OSDescription =>
         RuntimeInformation.OSDescription;
 
     // Version-gated check (only precise on .NET 5+)
     public static bool IsWindowsVersionAtLeast(int major, int minor = 0, int build = 0) =>
 #if NET5_0_OR_GREATER
         OperatingSystem.IsWindowsVersionAtLeast(major, minor, build);
 #else
         RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
             && Environment.OSVersion.Version >= new Version(major, minor, build);
 #endif
 }
```

The .NET Framework targets would need the `System.Runtime.InteropServices.RuntimeInformation` polyfill package from NuGet.

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -13,15 +13,16 @@
...
+      <TargetFrameworks>netstandard2.0;net8.0</TargetFrameworks>
       <RuntimeIdentifiers>win-x64;win-arm64;linux-x64;linux-arm64;osx-arm64</RuntimeIdentifiers> 
 
       <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
       <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
       <ImplicitUsings>enable</ImplicitUsings>
-      <IsAotCompatible>True</IsAotCompatible>
-      <IsTrimmable>True</IsTrimmable>
+      <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">True</IsAotCompatible>
+      <IsTrimmable Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">True</IsTrimmable>
```
jsquire: The trimmable node is redundant in this context; setting `<IsAotCompatible>` implicitly sets trimmable as well.

sdk/cs/test/FoundryLocal.Tests/Microsoft.AI.Foundry.Local.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net9.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
jsquire: I'd assume you would want to run your tests on `net462` as well, since it's the lowest supported .NET Framework version.

sdk/cs/test/FoundryLocal.Tests/Microsoft.AI.Foundry.Local.Tests.csproj
jsquire, line 45: nit: There's no benefit to using 9.x here.   I'd suggest aligning them to your 8.x dependency.

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
jsquire, line 128: Foundry is aligned on the official OpenAI library for its developer experience.  Why is a random third-party package used here?

scottaddie: I created https://github.com/microsoft/Foundry-Local/issues/632 to track that problem.

prathikr: yes, we are in the process of moving off this dependency

sdk/cs/test/FoundryLocal.Tests/Microsoft.AI.Foundry.Local.Tests.csproj
prathikr: Is there a downside to using 9.x?

jsquire: For this specific set, not that I see.  Generally, yes.   Many of the BCL packages are polyfills on modern runtimes and those on your version line would be able to elide the package entirely and use the built-in version.

sdk/cs/test/FoundryLocal.Tests/Microsoft.AI.Foundry.Local.Tests.csproj
prathikr: We do not want to support lower than net8.0

scottaddie: >We do not want to support lower than net8.0

@prathikr Does that imply you don't want .NET Framework customers using the library? For clarity, TFM `net462` corresponds to .NET Framework 4.6.2.

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -88,7 +88,7 @@
       <PackageId>Microsoft.AI.Foundry.Local.WinML</PackageId>
       <AssemblyName>Microsoft.AI.Foundry.Local.WinML</AssemblyName>
       <DefineConstants>$(DefineConstants);IS_WINML</DefineConstants>
-      <TargetFramework>net9.0-windows10.0.26100.0</TargetFramework> <!-- override -->
+      <TargetFrameworks>netstandard2.0;net8.0-windows10.0.26100.0</TargetFrameworks>
```
scottaddie: I assume WinML is inherently Windows only. If so, the `netstandard2.0` TFM would resolve for Linux and macOS, but fail at runtime. Consider restricting to the following targets instead:

```suggestion
      <TargetFrameworks>net462;net8.0-windows10.0.26100.0</TargetFrameworks>
```

The result is a more honest configuration that supports .NET Framework 4.6.2+ and supported, modern versions of .NET on Windows.

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
scottaddie: Is there a target date for completion of that work? Let us know if you have questions. Jesse and I represent the team that ships the official OpenAI library.

sdk/cs/test/FoundryLocal.Tests/Microsoft.AI.Foundry.Local.Tests.csproj
jsquire: > We do not want to support lower than net8.0

It's a valid decision, but it runs counter to the common developer experience and guidelines for both Azure Foundry and Azure itself.   Was this limitation discussed with leadership or made in isolation?

jsquire: In this context, since you've added `netstandard2.0` to your project targets, not running a .NET Framework leg in your tests does not remove support for it from the product - it simply leaves all .NET Framework runtimes untested.   Since there is a divergence between runtimes, this exposes you to a nasty class of bugs that is hard to root cause in isolation.

I'd very, very strongly suggest that you align your test runs with the frameworks that your product supports.

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -13,15 +13,15 @@
       <RepositoryUrl>https://github.com/microsoft/Foundry-Local</RepositoryUrl>
       <RepositoryType>git</RepositoryType>
 
-      <TargetFramework>net9.0</TargetFramework>
+      <TargetFrameworks>netstandard2.0;net8.0</TargetFrameworks>
```
jsquire: Expanding on my comment in the tests project - there's a misalignment here between what runtimes your product supports (this list) and what you're actively testing.

Your statement about not wanting to support runtimes earlier than 8.0 is in opposition to the target list here, as `netstandard2.0` will allow running on the supported .NET Framework runtimes  (4.26 - 4.81). 

Note that doing so is compliant with the Azure developer experience guidelines and the Azure Foundry developer experience across products.

prathikr: ok, will add netstandard2.0 for cross-platform and net462 for winml. is that the correct setup to be compliant with Azure Foundry developer experience?

.pipelines/templates/build-cs-steps.yml
```diff
@@ -37,10 +37,10 @@ steps:
       Write-Host "##vso[task.setvariable variable=testDataDir]$testDataDir"
 
 - task: UseDotNet@2
-  displayName: 'Use .NET 9 SDK'
+  displayName: 'Use .NET 8 SDK'
```
jsquire: I'd very, very strongly recommend using the .NET 10 SDK.   That will ensure that you've got the latest tooling, fixes, and performance enhancements.

To be explicit:  using the latest SDK does not prevent you from targeting other platforms and will work just fine with your `netstandard2.0` approach.

.pipelines/templates/build-cs-steps.yml
```diff
@@ -127,7 +127,7 @@ steps:
     targetType: inline
     script: |
       $base = "$(repoRoot)/sdk/cs/src/bin/Release"
-      # The SDK targets net9.0 (standard) or net9.0-windows10.0.26100.0 (WinML).
+      # The SDK targets net8.0 (standard) or net8.0-windows10.0.26100.0 (WinML).
```
jsquire: Now it targets `netstandard2.0` and any runtime compatible with that.

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -13,15 +13,15 @@
       <RepositoryUrl>https://github.com/microsoft/Foundry-Local</RepositoryUrl>
       <RepositoryType>git</RepositoryType>
 
-      <TargetFramework>net9.0</TargetFramework>
+      <TargetFrameworks>netstandard2.0</TargetFrameworks>
       <RuntimeIdentifiers>win-x64;win-arm64;linux-x64;linux-arm64;osx-arm64</RuntimeIdentifiers>
```
jsquire, line 17: Is there a reason why you need to explicitly state the `RuntimeIdentifiers` in the general case?  Do you have a native dependency bound to these?

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -13,15 +13,15 @@
...
 
       <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
       <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
       <ImplicitUsings>enable</ImplicitUsings>
-      <IsAotCompatible>True</IsAotCompatible>
-      <IsTrimmable>True</IsTrimmable>
+      <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">True</IsAotCompatible>
       <Nullable>enable</Nullable>
+      <LangVersion>latest</LangVersion>
       <WarningsAsErrors />
```
jsquire:
```suggestion
```

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
prathikr: Yes, we have a Foundry Local Core dependency which is a native AOT c# application.

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
prathikr: It is still under discussion, will update when I have a timeline.

sdk/cs/src/Detail/ICoreInterop.cs
```diff
@@ -22,7 +22,7 @@ internal record Response
     public delegate void CallbackFn(string callbackData);
 
     [StructLayout(LayoutKind.Sequential)]
-    protected unsafe struct RequestBuffer
+    public unsafe struct RequestBuffer
```
skottmckay: Do these need to be `public` or could they be `internal`? SDK users don't need to know anything about these types.

sdk/cs/src/Detail/ModelLoadManager.cs
```diff
@@ -156,7 +156,7 @@ private async Task WebUnloadModelAsync(string modelId, CancellationToken? ct = n
                                                 $"{response.ReasonPhrase}");
         }
 
-        var content = await response.Content.ReadAsStringAsync(ct ?? CancellationToken.None).ConfigureAwait(false);
+        var content = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
```
skottmckay, line 159: Why not pass `ct` through to the ReadAsStringAsync calls if we have it?

sdk/cs/src/Configuration.cs
```diff
@@ -109,17 +109,17 @@ internal Dictionary<string, string> AsDictionary()
...
         if (!string.IsNullOrEmpty(ModelCacheDir))
         {
-            configValues.Add("ModelCacheDir", ModelCacheDir);
+            configValues.Add("ModelCacheDir", ModelCacheDir!);
         }
 
         if (!string.IsNullOrEmpty(LogsDir))
         {
-            configValues.Add("LogsDir", LogsDir);
+            configValues.Add("LogsDir", LogsDir!);
```
skottmckay, line 122: Given we just called `string.IsNullOrEmpty` on these 3 values why is the `!` needed?

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -1,134 +1,131 @@
...
+     -->
+    <UseWinML>false</UseWinML>
+  </PropertyGroup>
+
+  <!--
+    A WinML build places constraints on the allowable target frameworks and
+    runtime properties due to the dependency.  Redefine with the proper overrides.
+  -->
+  <PropertyGroup Condition="'$(UseWinML)' == 'true'">
+    <TargetFrameworks>net8.0-windows10.0.26100.0</TargetFrameworks>
```
skottmckay, line 46: Was this going to be changed to `<TargetFrameworks>net462;net8.0-windows10.0.26100.0</TargetFrameworks>`? 
https://github.com/microsoft/Foundry-Local/pull/629#discussion_r3087711326 mentioned doing that

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -1,134 +1,131 @@
...
-      <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
-      <PackageProjectUrl>https://github.com/microsoft/Foundry-Local</PackageProjectUrl>
-      <PackageDescription>Microsoft AI Foundry Local SDK for .NET</PackageDescription>
-      <PackageTags>Microsoft AI Foundry SDK</PackageTags>
-      <PackageReadmeFile>README.md</PackageReadmeFile>
-      <RepositoryUrl>https://github.com/microsoft/Foundry-Local</RepositoryUrl>
-      <RepositoryType>git</RepositoryType>
-
-      <TargetFramework>net9.0</TargetFramework>
-      <RuntimeIdentifiers>win-x64;win-arm64;linux-x64;linux-arm64;osx-arm64</RuntimeIdentifiers>
```
skottmckay, line 17: Was there a reason to delete the RuntimeIdentifiers list to align with builds in FL Core or does that happen implicitly and that's good enough? If implicit, why do we define them in the WinML section?

sdk/cs/test/FoundryLocal.Tests/Microsoft.AI.Foundry.Local.Tests.csproj
```diff
@@ -1,27 +1,33 @@
...
 
-  <PropertyGroup Condition="'$(RuntimeIdentifier)'==''">
-    <RuntimeIdentifier>$(NETCoreSdkRuntimeIdentifier)</RuntimeIdentifier>
+  <!--
+    .NET Framework test executables default to Prefer32Bit=true, causing the build to
+    look for a RID-specific target that restore doesn't generate. Set RuntimeIdentifier
+    explicitly so restore and build agree.
+  -->
+  <PropertyGroup Condition="$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)')) == '.NETFramework'">
+    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
```
skottmckay, line 23: Is this required because we removed `RuntimeIdentifiers` from sdk/cs/src/Microsoft.AI.Foundry.Local.csproj?

sdk/cs/test/FoundryLocal.Tests/Microsoft.AI.Foundry.Local.Tests.csproj
```diff
@@ -42,14 +48,28 @@
...
 
+  <!--
+    Warn during build when integration test infrastructure is missing.
+    The dotnet test terminal logger hides individual skip reasons from
+    the default output; MSBuild warnings are always visible.
+  -->
+  <Target Name="_WarnMissingTestInfrastructure" BeforeTargets="Build">
+    <PropertyGroup>
+      <_RepoRoot>$([MSBuild]::NormalizePath('$(MSBuildProjectDirectory)', '..', '..', '..', '..'))</_RepoRoot>
+      <_TestDataSharedPath>$([MSBuild]::NormalizePath('$(_RepoRoot)', '..', 'test-data-shared'))</_TestDataSharedPath>
```
skottmckay: This assumes the default location is being used, but that can be overridden with the `TestModelCacheDirName` config value.

prathikr: reverted

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
prathikr: @jsquire mentioned to me that win-x64, win-arm64, linux-x64, and osx-arm64 are the only platforms for modern .NET so defining these was unnecessary

sdk/cs/test/FoundryLocal.Tests/Microsoft.AI.Foundry.Local.Tests.csproj
prathikr: No, it was to get the build working locally. Will remove.

prathikr:
Needed to apply some fixes for netstandard2.0

1. the string.IsNullOrEmpty() guard ensures non-null, but the netstandard2.0 nullable annotations don't narrow the type.
2. the CancellationToken overload doesn't exist in netstandard2.0's HttpContent
3. WinML does not support netstandard2.0, they require minimum net8.0

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
skottmckay: Does it not support linux-arm64? The .NET SDK has linux arm64 packages.

linux-arm64 is a potential scenario coming up but we don't have a genai build for it (we do have an ORT one).

sdk/cs/src/Detail/ModelLoadManager.cs
skottmckay: What's the resolution here? PR still removes the pass through of `ct`

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
skottmckay: What's the resolution here? Is it that WinML doesn't support netstandard2.0?

sdk/cs/src/Detail/ModelLoadManager.cs
jsquire: The overload with a cancellation token doesn't exist in the `netstandard2.0` target.    It was removed based on analysis that the response payload does not appear to be streamed and cancellation isn't likely to be triggered/honored on this path.  To restore it, you'd have to conditionalize the flow based on targets and the value doesn't seem to warrant the complexity.

skottmckay: Thanks for the explanation.

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
jsquire: Modern .NET runtimes (.NET 8.0+) cover all the x64 targets, yes.   The runtime identifiers are emitted automatically.   The only time that you'd need/want to constrain them is when you've got native dependencies that don't support all targets.   In the case of the .NET Framework, 32-bit targets do come into play, but restore would cover the failure case should there be a dependency gap.

skottmckay: The native Core library doesn't have a linux-arm64 build due to its GenAI dependency not having a linux-arm64 build currently. As soon as we add the GenAI build we can linux-arm64 support above that.

What is the best practice in this case? Let the restore fail or explicitly declare the runtimes we support which are bound by the native libraries support?

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
jsquire: AOT != native.   Those are different considerations.   The .NET compiler will ensure that AOT scenarios are emitting the right runtime builds by default.   When I say `native` in this context, I mean that you're packaging and including a native Win32/64 DLL.

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
jsquire: Generally, as a library rather than an application, you'd want to allow restore to fail and let the application that consumes you assert control over the runtimes.

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
skottmckay: The Core nuget package is AOT compiled to native from C# code so that the C API it provides is available for use from other language bindings. So the code is C#, but what is used as a dependency in the C#/Python/JS/Rust SDK is a native binary.  

afaik that's the only way to get the C entry points exported from C# code. Has that changed?

skottmckay (dismissed):

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
jsquire: Libraries operate differently than applications.  As a library, this package is not AOT compiled for distribution - just published as a standard NuGet package.  By marking as AOT-friendly, it allows the application that consumes it to opt for AOT compilation, if desired.   Any consumer that references the package will have control.   So, it's safe to mark the Core package as AOT eligible and not impact your C interop scenarios.  

That said, I'm not sure if there are AOT compilation options that allow exports for C interop or not.   I can point you at the right folks to talk to if you're curious.

baijumeswani (approved):

scottaddie (approved):
As communicated via Teams:

- Mention in your release notes that you've significantly expanded your library's reach in the .NET ecosystem. It's no longer just .NET 9+ customers that can use it.
- Once the new package version ships, update the prereqs in this doc section: https://learn.microsoft.com/azure/foundry-local/get-started?tabs=windows&pivots=programming-language-csharp#prerequisites

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -1,134 +1,132 @@
...
+    A WinML build places constraints on the allowable target frameworks and
+    runtime properties due to the dependency.  Redefine with the proper overrides.
+  -->
+  <PropertyGroup Condition="'$(UseWinML)' == 'true'">
+    <!-- WinML does not support netstandard2.0, so we only target net8.0 for the WinML build. -->
+    <TargetFrameworks>net8.0-windows10.0.26100.0</TargetFrameworks>
+    <RuntimeIdentifiers>win-x64;win-arm64</RuntimeIdentifiers>
+    <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
+
+    <IsAotCompatible>true</IsAotCompatible>
```
jsquire, line 50: I'd restore the previous designation of setting this conditionally for modern .NET targets as AOT compatible.  If you already support it for WinML, there's no reason to not declare support fully.

As mentioned in a comment elsewhere, this does not package your library as AOT, it simply marks you as compliant so that an application consuming your library can choose to compile itself as an AOT executable.

jsquire (approved):

sdk/cs/src/Microsoft.AI.Foundry.Local.csproj
prathikr: will do in a separate PR, thanks!

PR #630, opened 2026-04-14
expands tests to run on linux-x64 machines

prathikr:


prathikr:
continued in https://github.com/microsoft/Foundry-Local/pull/638

PR #664, opened 2026-04-22
Add ORT-Nightly as fallback for nuget package source at installation time

baijumeswani:
The JS and Rust SDK installers download three native NuGet packages (Microsoft.AI.Foundry.Local.Core, 
Microsoft.ML.OnnxRuntime.Foundry, Microsoft.ML.OnnxRuntimeGenAI.Foundry) from a single hard-coded feed (
api.nuget.org). Dev / pre-release versions are published to the public ORT-Nightly Azure DevOps feed before they
reach nuget.org, so any build pinned to a dev version fails outright today.

Change

Both installers now try each feed in order and fall back to the next on any failure:

 1. https://api.nuget.org/v3/index.json (primary)
 2. https://pkgs.dev.azure.com/aiinfra/PublicPackages/_packaging/ORT-Nightly/nuget/v3/index.json (fallback)

### DIFF
--- a/sdk/js/script/install-utils.cjs
+++ b/sdk/js/script/install-utils.cjs
@@ -29,7 +29,15 @@ const REQUIRED_FILES = [
   `${os.platform() === 'win32' ? '' : 'lib'}onnxruntime-genai${EXT}`,
 ];
 
-const NUGET_FEED = 'https://api.nuget.org/v3/index.json';
+// Feeds tried in order. Primary: nuget.org (stable releases). Fallback:
+// the public ORT-Nightly Azure DevOps NuGet feed (where dev / pre-release
+// builds of Foundry Local Core, ONNX Runtime and ONNX Runtime GenAI live
+// before they reach nuget.org). If a download from a feed fails for any
+// reason, the next feed is tried.
+const FEEDS = [
+    'https://api.nuget.org/v3/index.json',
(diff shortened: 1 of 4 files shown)


sdk/js/script/install-utils.cjs
```diff
@@ -104,6 +109,48 @@ async function getBaseAddress(feedUrl) {
     return baseAddress.endsWith('/') ? baseAddress : baseAddress + '/';
 }
 
+// Cache per (feed, package) of the flat-container "index.json" listing all
+// available versions for a package id on that feed.
+const versionsIndexCache = new Map();
+
+async function getPackageVersions(feedUrl, pkgName) {
```
prathikr: do we need to list all versions if we are looking for a specific version? it can just look for that one version in nuget.org or ort-nightly and fail if not found without enumerating all versions

sdk/rust/build.rs
```diff
@@ -184,6 +190,66 @@ fn resolve_base_address(feed_url: &str) -> Result<String, String> {
     ))
 }
 
+/// Fetch the list of available versions for a given package id on a NuGet v3
+/// feed via the flat-container index.json. Returns lowercase version strings.
+/// Any failure (404, network, missing resource) is treated as "unavailable".
+fn fetch_package_versions(feed_url: &str, pkg_name: &str) -> Vec<String> {
```
prathikr: same comment as for js

prathikr (approved):

PR #668, opened 2026-04-23
Added embedding API to C++ SDK

phanindraraja:
## Summary

Adds embedding support to the C++ SDK and aligns the embedding samples across all SDK languages to a consistent initialization sequence. Includes optimizations to the existing embedding client implementations.

### C++ SDK — embedding client

Public `OpenAIEmbeddingClient` exposing `GenerateEmbedding` (single input) and `GenerateEmbeddings` (batch input). Empty and whitespace-only inputs are rejected client-side. Responses parse into strongly-typed `EmbeddingCreateResponse` / `EmbeddingObject` / `EmbeddingUsage` structs via nlohmann/json.

- **Tests:** 13 unit tests with a mocked core (no native DLL required) and one E2E test that exercises the full download → load → embed → unload flow against a real model.
- **In-tree sample:** `sdk/cpp/sample/main.cpp` gains an embeddings example. The example prefers a CPU variant, validates that the model is loaded, generates one single-input embedding and one batch embedding, and unloads the model.
- **Standalone sample:** `samples/cpp/embeddings/{main.cpp,README.md}` provides a top-level entry point for the embeddings demo. Discovers and registers execution providers, then runs the same single + batch flow.
- **Documentation:** `sdk/cpp/README.md` lists embeddings under Features, gains a usage section with a code snippet, and registers `OpenAIEmbeddingClient` in the API reference table and `embedding_client.h` in the project structure.

### Embedding sample alignment

Every embedding sample runs the same seven-step sequence: create the manager → discover execution providers (prints a table) → download and register execution providers → get the catalog → get the model → download the model → load the model. The model alias is `qwen3-embedding-0.6b` across all samples.

### Optimizations to existing embedding clients

- Materialize the batch input collection once before validation and serialization, avoiding multiple enumerations.
- Rename a setup-method local that shadowed a class-level static.
- Remove a test helper that mutated a shared model config file in place without restoring it.
- Replace a value-discarding map call with an idiomatic insert-if-absent pattern when injecting missing response fields.
- Drop redundant length checks on collections already covered by truthiness.
- Remove a stale TOC entry from the API docs.

## Testing

- C++ unit tests: 13/13 pass under `ctest --preset x64-debug`.
- Full C++ test suite: 168/168 ran tests pass.
- C++ E2E embedding test runs locally with the native Core DLL staged alongside the test binary and `--gtest_also_run_disabled_tests`.
- C# SDK builds clean for both `net8.0` and `netstandard2.0` targets.
- All embedding integration surfaces are unchanged from the user's perspective.


phanindraraja:
Closing in favor of #705, which open-sources FoundryLocalCore and will auto-enable embedding in the C++ SDK.

PR #669, opened 2026-04-23
Add MIT License for Nemotron Speech Streaming

samuel100:
Added an MIT license file into the licenses folder for the Nemotron speech model. As this is a derivative model of https://huggingface.co/nvidia/nemotron-speech-streaming-en-0.6b it has a different license applied.


samuel100:
Model licenses do not go in GH repo

PR #671, opened 2026-04-23
test(samples/js): add Responses web service sample and integration coverage

MaanavD:
## Summary

Reworks this PR to avoid changing the JavaScript SDK API surface. The PR now focuses on using an existing Foundry Local OpenAI-compatible web service for Responses API scenarios, similar to the existing chat-completions web-server sample.

## Changes

- Restored the JS SDK Responses client/types/factory surface to the main-branch baseline.
- Added `samples/js/web-server-responses`, a JavaScript sample that uses the OpenAI JS SDK against a running Foundry Local `/v1` web service and demonstrates:
  - `openai.responses.create()` non-streaming calls,
  - streaming Responses events,
  - Responses tool calling with `function_call_output` follow-up.
- Added `sdk/js/test/openai/responsesWebService.test.ts` integration coverage that calls `/v1/responses` directly with `fetch` for:
  - non-streaming response creation,
  - SSE streaming events,
  - function calling over the web service.
- The web-service integration tests do not instantiate `FoundryLocalManager` or the JS native addon. They require an already-running service and are enabled with:
  - `FOUNDRY_LOCAL_RESPONSES_ENDPOINT` or `FOUNDRY_LOCAL_ENDPOINT`
  - `FOUNDRY_LOCAL_RESPONSES_MODEL` or `FOUNDRY_LOCAL_MODEL`
- Updated sample READMEs to list the new JavaScript sample.

## Local validation

```text
npm run build: passed
npx mocha --import=tsx test/openai/responsesClient.test.ts test/openai/responsesWebService.test.ts --grep "Integration" --invert: 29 passing
npx mocha --import=tsx test/openai/responsesWebService.test.ts --grep "Integration": 3 pending (skipped because no external web-service env vars were set)
```


sdk/js/src/openai/vision.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @throws If the file extension is not a supported image format.
+ */
+export function createImageContentFromFile(filePath: string, detail?: 'low' | 'high' | 'auto'): InputImageContent {
+    const ext = path.extname(filePath).toLowerCase();
+    const mediaType = MEDIA_TYPE_MAP[ext];
+    if (!mediaType) {
+        throw new Error(`Unsupported image format: ${ext}. Supported formats: ${Object.keys(MEDIA_TYPE_MAP).join(', ')}`);
+    }
+
+    const data = fs.readFileSync(filePath);
```
apsonawane: Can we check if the file exists or not

sdk/js/src/openai/vision.ts
```diff
@@ -0,0 +1,63 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+// -------------------------------------------------------------------------
+
+import * as fs from 'fs';
+import * as path from 'path';
+import type { InputImageContent } from '../types.js';
+
+const MEDIA_TYPE_MAP: Record<string, string> = {
+    '.png': 'image/png',
```
apsonawane: We should support bmp as well, python supports it

sdk/js/src/openai/vision.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Creates an `InputImageContent` part by reading an image file from disk.
+ * The file is base64-encoded and embedded directly in the content part.
+ *
+ * @param filePath - Absolute or relative path to the image file.
+ * @param detail - Optional detail level hint for the model ('low' | 'high' | 'auto').
+ * @returns An `InputImageContent` object with base64-encoded image data.
+ * @throws If the file extension is not a supported image format.
+ */
+export function createImageContentFromFile(filePath: string, detail?: 'low' | 'high' | 'auto'): InputImageContent {
```
apsonawane: Add image resize here as well

sdk/js/src/types.ts
```diff
@@ -127,6 +127,20 @@ export interface InputTextContent {
     text: string;
 }
 
+export interface InputImageContent {
+    type: 'input_image';
+    image_url?: string;
+    image_data?: string;       // base64-encoded
+    media_type: string;        // e.g. "image/png"
```
apsonawane: Make media type optional, so that the server will infer the type

sdk/js/src/openai/vision.ts
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param url - Public URL pointing to the image.
+ * @param detail - Optional detail level hint for the model ('low' | 'high' | 'auto').
+ * @returns An `InputImageContent` object with the image URL.
+ */
+export function createImageContentFromUrl(url: string, detail?: 'low' | 'high' | 'auto'): InputImageContent {
+    const content: InputImageContent = {
+        type: 'input_image',
+        image_url: url,
+        media_type: 'image/unknown', // server will detect from URL
```
apsonawane: Instead of setting it unknown, just omit it here

sdk/js/src/openai/vision.ts
```diff
@@ -54,10 +91,44 @@ export function createImageContentFromUrl(url: string, detail?: 'low' | 'high' |
...
+
+    if (Math.max(width, height) <= maxDimension) {
+        return data; // already within bounds
+    }
+
+    const resized: Buffer = await sharp(data)
+        .resize({ width: maxDimension, height: maxDimension, fit: 'inside', withoutEnlargement: true })
+        .toBuffer();
+
+    return resized;
```
apsonawane: We need to return media_type after resize, similar to python sdk

sdk/js/src/openai/responsesClient.ts
```diff
@@ -76,7 +77,8 @@ export class ResponsesClientSettings {
             tool_choice: this.toolChoice,
             truncation: this.truncation,
             parallel_tool_calls: this.parallelToolCalls,
-            store: this.store,
+            // Default store to true when not explicitly set
+            store: this.store !== undefined ? this.store : true,
```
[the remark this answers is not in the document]
baijumeswani: @MaanavD is this a valid concern?

sdk/js/src/types.ts
```diff
@@ -127,6 +127,20 @@ export interface InputTextContent {
     text: string;
 }
 
+export interface InputImageContent {
+    type: 'input_image';
+    image_url?: string;
+    image_data?: string;       // base64-encoded
+    media_type?: string;       // e.g. "image/png"; omit to let the server infer
+    detail?: 'low' | 'high' | 'auto';
+}
```
baijumeswani: Feedback from AI:

```
 - The OpenAI Responses API spec uses image_url carrying either an https URL or a data:<mime>;base64,... URI. There is no separate image_data/media_type pair in the 
public OpenAI schema.
 - Grep across the repo finds zero references to input_image, image_data, or image_url in any other SDK or server code, so this contract isn't validated by anything 
else in-tree.
 - Recommend: either point at the Foundry Local server schema that defines these fields, or fall back to OpenAI's data-URI convention (and have 
createImageContentFromFile produce image_url: 'data:${mime};base64,${b64}').
```

sdk/js/examples/responses.ts
```diff
@@ -121,6 +122,35 @@ async function main() {
...
+        // =================================================================
+        console.log('\n--- Example 6: List stored responses ---');
+        const allResponses = await client.list();
+        console.log(`Listed ${allResponses.data.length} stored responses`);
+
+        // =================================================================
+        // Example 7: Vision — describe an image
+        // =================================================================
+        console.log('\n--- Example 7: Vision ---');
+        const testImagePath = 'path/to/test-image.png'; // Replace with a real image path
```
baijumeswani: The example is broken if it requires user to replace it with a real image (and no warning/error message from the example). Could you either check in an image or download one from a trusted source at example run time?

sdk/js/src/openai/vision.ts
baijumeswani: @MaanavD could you list bmp as supported media type in other parts of the PR (documentation and vision endpoints).

sdk/js/src/openai/responsesClient.ts
```diff
@@ -275,6 +282,14 @@ export class ResponsesClient {
         );
     }
 
+    /**
+     * Lists all stored responses.
+     * @returns The list of Response objects.
+     */
+    public async list(): Promise<ListResponsesResult> {
+        return this.fetchJson<ListResponsesResult>('/v1/responses', { method: 'GET' });
+    }
```
baijumeswani: From AI:

```
- OpenAI's list endpoints accept limit, after, before, order and the response carries first_id, last_id, has_more. The PR's ListResponsesResult only models object and
 data. If the server actually returns those pagination fields, callers can't use them without a cast.
 - Either add the optional pagination params + response fields, or note explicitly that the Foundry Local server doesn't support them.
```

sdk/js/src/types.ts
```diff
@@ -419,6 +433,55 @@ export interface FunctionCallArgsDoneEvent {
     sequence_number: number;
 }
 
+export interface ReasoningSummaryPartAddedEvent {
+    type: 'response.reasoning_summary_part.added';
+    item_id: string;
+    part: ContentPart;
+    sequence_number: number;
+}
```
baijumeswani: Could you suggest what these events do? Are they correctly emitted from the code?

sdk/js/test/testUtils.ts
```diff
@@ -46,6 +46,25 @@ export const TEST_CONFIG: FoundryLocalConfig = {
...
+    const candidatePaths = [
+        path.join(sdkRoot, 'prebuilds', platformKey, 'foundry_local_napi.node'),
+        path.join(sdkRoot, 'native', 'build', 'Release', 'foundry_local_napi.node'),
+        path.join(sdkRoot, 'dist', 'prebuilds', platformKey, 'foundry_local_napi.node'),
+        path.join(sdkRoot, 'dist', 'native', 'build', 'Release', 'foundry_local_napi.node'),
+    ];
+    return candidatePaths.some(p => fs.existsSync(p));
+}
+
+export const IS_NATIVE_ADDON_AVAILABLE = checkNativeAddonAvailable();
```
baijumeswani: Why is this check necessary?

sdk/js/test/openai/responsesClient.test.ts
```diff
@@ -377,7 +581,7 @@ describe('ResponsesClient Tests', () => {
 
         before(async function() {
             this.timeout(30000);
-            if (IS_RUNNING_IN_CI) {
+            if (IS_RUNNING_IN_CI || !IS_NATIVE_ADDON_AVAILABLE) {
```
baijumeswani: This would accidentally skip tests if dev forgets to build the addon? What is the need for this?

sdk/js/src/openai/responsesClient.ts
MaanavD: Done. `ListResponsesResult` now includes `first_id`, `last_id`, and `has_more`, and `ResponsesClient.list()` now accepts `{ limit, order, after }`. I did not add `before` because the current Foundry Local server endpoint only binds `limit`, `order`, and `after`.

sdk/js/src/openai/responsesClient.ts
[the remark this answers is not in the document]
MaanavD: Yes, the concern is valid in that this is a behavior change, but it is intentional for Foundry Local's Responses SDK. The implementation spec explicitly calls out `store` defaulting to `true` to match OpenAI convention even though the server default is currently `false`. I added README documentation showing that responses are stored by default and that callers can opt out via `client.settings.store = false` or per-request `store: false`.

sdk/js/test/openai/responsesClient.test.ts
MaanavD: Agreed, removed. This no longer auto-skips when the native addon is missing, so a local developer who forgets to build the addon will see the integration test setup fail instead of silently skipping.

sdk/js/src/openai/vision.ts
MaanavD: Done. BMP is now listed in the `createImageContentFromFile()` JSDoc and in the JS README vision-helper docs alongside png/jpg/jpeg/gif/webp.

sdk/js/src/types.ts
MaanavD: This is Foundry Local's server contract rather than the public OpenAI wire shape. The server DTO defines `input_image` with `image_url`, `image_data`, and `media_type`, and `ResponseInputConverter` turns `image_data + media_type` into the `data:<mime>;base64,...` URL form before calling the underlying chat completion path. I also documented this in the JS README so consumers know `createImageContentFromFile()` emits the Foundry Local contract.

sdk/js/test/testUtils.ts
MaanavD: Removed. The file-existence/native-addon availability check is no longer needed; integration tests now only keep the existing CI skip and otherwise fail normally if a local dev tries to run them without building the addon.

sdk/js/examples/responses.ts
MaanavD: Done. The example now creates a temporary 1x1 PNG at runtime and passes that to `createImageContentFromFile()`, so it no longer requires users to replace a placeholder path or check in an image asset.

sdk/js/src/types.ts
MaanavD: Updated. These interfaces model the server DTOs in `neutron-server/src/FoundryLocalCore/Core/Responses/Contracts/StreamingEvents.cs`. They are contract types for reasoning/annotation stream events; current streaming production primarily emits lifecycle, output, text delta, and function-call events, but the server already declares these reasoning event shapes for compatibility/future emission. The TS interfaces now include the server fields (`output_index`, `summary_index`, `content_index`, `annotation_index`, and optional `obfuscation`) so they match the DTOs.

baijumeswani (dismissed):

PR #674, opened 2026-04-24
Propagate cancellation to model and EP downloads

### LINKED ISSUE #365: CancellationToken Not Honored in IModel.DownloadAsync Method
The `IModel.DownloadAsync(Action<float>, CancellationToken)` method accepts a `CancellationToken` parameter but does not properly check or honor cancellation requests during the download operation. This results in downloads continuing to completion even after cancellation has been explicitly requested, making it impossible to reliably cancel long-running model downloads.

## Environment
- **SDK**: Microsoft.AI.Foundry.Local.WinML v0.8.2.1
- **Platform**: Windows
- **Application**: WinUI 3 Desktop App
- **.NET Version**: .NET 8.0

## Problem
When calling `model.DownloadAsync(progressCallback, cancellationToken)` and subsequently canceling the operation via `CancellationTokenSource.Cancel()`, the download operation continues to execute until completion (100%) despite the cancellation request. The SDK only appears to check the cancellation token during the subsequent `LoadAsync()` operation, not during the actual download phase.

This behavior violates the standard .NET async cancellation pattern where operations accepting a `CancellationToken` should periodically check `token.IsCancellationRequested` and throw `OperationCanceledException` when cancellation is requested.

## Repro Step

1. Initialize a FoundryLocal catalog and obtain a model reference:
```csharp
var catalog = await foundryManager.GetCatalogAsync();
var model = await catalog.GetModelAsync(alias);
```

2. Start a model download with cancellation support:
```csharp
var cts = new CancellationTokenSource();
var downloadTask = model.DownloadAsync(
    progressPercent => {
        Debug.WriteLine($"Download progress: {progressPercent}%");
    },
    cts.Token
);
```

3. During download (e.g., at 50-60% progress), request cancellation:
```csharp
cts.Cancel();
```

4. Observe that the download continues to 100% completion despite cancellation request.

## Actual Behavior

**Debug Output Evidence:**

```
[FoundryClient] Download progress: 42.44%
[FoundryClient] Download progress: 49.79%
[App] User clicked cancel button
[App] Calling CancellationTokenSource.Cancel()
[App] CancellationToken AFTER Cancel: IsCancellationRequested: True

[FoundryClient] Progress callback - CancellationToken is REQUESTED!
[FoundryClient] Download progress: 57.19%
[FoundryClient] Download progress: 65.31%
[FoundryClient] Download progress: 73.44%
[FoundryClient] Download progress: 81.56%
[FoundryClient] Download progress: 89.69%
[FoundryClient] Download progress: 97.81%
[FoundryClient] Download progress: 100.00%
[FoundryClient] ========== SDK DownloadAsync COMPLETED ==========

[FoundryClient] Starting model preparation...
[FoundryClient] OperationCanceledException in PrepareModelAsync: The operation was canceled.
```

**Key Observations:**
1. `CancellationToken.IsCancellationRequested` returns `True` immediately after `Cancel()` is called
2. The progress callback continues to be invoked with increasing percentages (57% → 100%)
3. Within the progress callback, we can verify `token.IsCancellationRequested == True`, but the SDK ignores it
4. `DownloadAsync` completes successfully without throwing `OperationCanceledException`
5. Only the subsequent `LoadAsync()` operation respects the cancellation token

## Expected Behavior

When `CancellationToken.Cancel()` is called:
1. The SDK should detect `cancellationToken.IsCancellationRequested == True` during the download loop
2. `DownloadAsync` should stop the download operation promptly
3. `DownloadAsync` should throw `OperationCanceledException` to signal cancellation
4. Partial downloads should be cleaned up appropriately
5. The operation should not continue to 100% completion

This aligns with the standard .NET async cancellation pattern documented in Microsoft's async programming guidelines.

## Impact

- Users cannot cancel large model downloads, leading to wasted bandwidth and storage
- UI "Cancel" buttons appear non-functional, creating confusion
- No way to interrupt mistaken downloads of large models (multi-GB files)
- Applications cannot implement reliable cancellation for download operations
- Poor user experience in applications using Foundry Local

## Related Code References

- `IModel.DownloadAsync(Action<float>, CancellationToken)` - Primary method affected
- `IModel.LoadAsync(CancellationToken)` - Correctly honors cancellation

## Additional Context

This issue was discovered during user testing when users attempted to cancel large model downloads. The debug logging was added specifically to trace the cancellation flow, confirming that the application code correctly propagates the cancellation token through all layers, but the SDK's `DownloadAsync` method does not check it.

The fact that `LoadAsync()` properly throws `OperationCanceledException` when the token is canceled demonstrates that the SDK infrastructure supports cancellation - it just needs to be implemented consistently across all async operations, particularly `DownloadAsync`.

---

**Thank you for considering this issue. Proper cancellation support is critical for production applications managing large model downloads. Please let me know if you need any additional information or testing.**

bmehta001:
Adds cancellation support for SDK model and execution provider downloads across C#, Python, JavaScript/TypeScript, Rust, and C++.
 
- Cancellation is wired through the existing native streaming callback path so downloads can be interrupted at the next native progress/callback boundary while preserving the existing non-streaming fast path when no progress callback or cancellation token/signal is supplied.
 
 ## Changes
 
 - Added cancellable model download APIs across SDKs:
   - C#: `CancellationToken`
   - Python: cancellation event
   - JS/TS: `AbortSignal`, including `model.download(signal)`
   - Rust: `Arc<AtomicBool>`
   - C++: `CancellationCallback`
 - Preserved existing progress callback behavior and added cancellation support alongside progress callbacks.
 - Kept fast paths for downloads that do not request progress or cancellation.
 - Hardened callback handling:
   - Avoids converting successful operations into cancellation when cancellation is requested after the final observed callback.
   - Ensures native response buffers are cleaned up before surfacing cancellation/errors.
   - Avoids flushing buffered partial streaming data after cancellation.
 - Improved mixed progress parsing so status text and numeric progress chunks are handled consistently.
 - Updated SDK docs and samples
 
 ## Validation
 
 - Added cancellation/progress parsing tests for C#, Python, JS, Rust


sdk/python/src/detail/core_interop.py
```diff
@@ -264,14 +279,16 @@ def _execute_command(self, command: str, interop_request: InteropRequest = None,
...
+            callback_helper = CallbackHelper(callback, cancel_event)
             callback_py_obj = ctypes.py_object(callback_helper)
             callback_helper_ptr = ctypes.cast(ctypes.pointer(callback_py_obj), ctypes.c_void_p)
             callback_fn = CoreInterop.CALLBACK_TYPE(CallbackHelper.callback)
 
             lib.execute_command_with_callback(ctypes.byref(req), ctypes.byref(resp), callback_fn, callback_helper_ptr)
 
             if callback_helper.exception is not None:
+                if isinstance(callback_helper.exception, CancelledException):
+                    raise FoundryLocalException("Operation cancelled")
```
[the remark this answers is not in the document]
bmehta001: Fixed

sdk/cs/src/Detail/CoreInterop.cs
```diff
@@ -199,6 +199,11 @@ public Response ExecuteCommandImpl(string commandName, string? commandInput,
 
                 if (helper.Exception != null)
                 {
+                    if (helper.Exception is OperationCanceledException canceledException)
+                    {
+                        throw canceledException;
```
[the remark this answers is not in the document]
bmehta001: Fixed

sdk/cpp/src/core_helpers.h
```diff
@@ -47,37 +48,91 @@ namespace foundry_local::detail {
...
+        const std::string text(token);
+        size_t processed = 0;
+        try {
+            value = std::stod(text, &processed);
+        }
+        catch (...) {
+            return false;
+        }
+
+        return processed == text.size();
```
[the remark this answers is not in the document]
bmehta001: Fixed

sdk/cs/src/Detail/ModelVariant.cs
```diff
@@ -144,16 +146,22 @@ private async Task DownloadImplAsync(Action<float>? downloadProgress = null,
...
+                if (ct is CancellationToken cancellationToken)
+                {
+                    cancellationToken.ThrowIfCancellationRequested();
+                }
+
+                if (downloadProgress != null &&
+                    float.TryParse(progressString,
+                                   NumberStyles.Float,
+                                   CultureInfo.InvariantCulture,
+                                   out var progress))
```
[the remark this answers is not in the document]
bmehta001: Fixed

sdk/python/requirements.txt
```diff
@@ -1,9 +1,8 @@
 pydantic>=2.0.0
 requests>=2.32.4
 openai>=2.24.0
-# Standard native binary packages from the ORT-Nightly PyPI feed.
 foundry-local-core==1.1.0
-onnxruntime-core==1.25.1; sys_platform != "linux"
-onnxruntime-gpu==1.25.1; sys_platform == "linux"
-onnxruntime-genai-core==0.13.2; sys_platform != "linux"
-onnxruntime-genai-cuda==0.13.2; sys_platform == "linux"
+onnxruntime-gpu==1.25.1; platform_system == "Linux"
```
bmehta001: Automatically regenerated from `sdk/python/build_backend.py` - not sure why these changes are here, but I can revert

sdk/js/src/foundryLocalManager.ts
```diff
@@ -178,35 +185,84 @@ export class FoundryLocalManager {
...
+     */
+    public downloadAndRegisterEps(names: undefined, signal: AbortSignal): Promise<EpDownloadResult>;
+    /**
+     * Downloads and registers execution providers, preserving compatibility with callers that pass undefined for names.
+     * @param names - Undefined to download all EPs.
+     * @param progressCallback - Callback invoked with (epName, percent) as each EP downloads. Percent is 0-100.
+     * @param signal - Optional AbortSignal used to cancel an in-progress download.
+     * @returns A promise that resolves with an EpDownloadResult describing the outcome.
+     */
+    public downloadAndRegisterEps(names: undefined, progressCallback: (epName: string, percent: number) => void, signal?: AbortSignal): Promise<EpDownloadResult>;
```
baijumeswani, line 249: Functionally the overload looks good. But from a maintenance standpoint, I fear this will become hard to maintain if number of arguments keeps growing. Maybe we should add downloadAndRegisterEpsWithOptions? Something we can consider doing later.

sdk/rust/src/detail/model.rs
```diff
@@ -213,6 +213,23 @@ impl Model {
         self.selected_variant().download(progress).await
     }
 
+    /// Like [`Self::download`], but accepts a shared cancellation flag
+    /// (`Arc<AtomicBool>`). When the flag is set to `true`, the download
+    /// will be cancelled at the next progress callback and an error is
+    /// returned.
+    pub async fn download_cancellable<F>(
```
baijumeswani: AI suggested that we use a Builder pattern here:

```rust
pub fn download(&self) -> DownloadBuilder<'_> { DownloadBuilder::new(self) }
 
 pub struct DownloadBuilder<'a> { /* ... */ }
 impl<'a> DownloadBuilder<'a> {
     pub fn progress(mut self, cb: impl FnMut(f64)) -> Self { ... }
     pub fn cancel(mut self, flag: Arc<AtomicBool>) -> Self { ... }
     pub async fn run(self) -> Result<()> { ... }
 }
```

Call sites:

```rust
model.download().run().await?;
model.download().progress(|p| ...).cancel(flag).run().await?;
```

Not sure how feasible this is. Can you evaluate its feasibility?

bmehta001:
Addressed the latest API feedback in \1d32d091\: added \downloadAndRegisterEpsWithOptions\/\DownloadAndRegisterEpsOptions\ for JS so new call sites can avoid the overload growth, and added an additive Rust \DownloadBuilder\ via \model.download_builder().progress(...).cancel(...).run()\. I kept the existing Rust \download(...)\ and \download_cancellable(...)\ methods for compatibility, so the exact \model.download().run()\ shape would need a future breaking change.\n\nValidated with JS build + focused \downloadAndRegisterEps\ tests and Rust \cargo fmt --check\, \cargo check --workspace\, and \cargo clippy --workspace -- -D warnings\.

sdk/js/src/foundryLocalManager.ts
bmehta001: Will add downloadAndRegisterEpsWithOptions later, if we need it

sdk/rust/src/detail/model.rs
bmehta001: Yes, this works well

baijumeswani (approved):

PR #685, opened 2026-04-30
Add a non-blocking initialization path to the JS SDK

baijumeswani:
The JS SDK's FoundryLocalManager.create() and ModelVariant.load() perform synchronous native calls that block the Node.js event loop.
This PR adds a non-blocking path without breaking the existing GA APIs.

Changes

Native addon (sdk/js/native/foundry_local_napi.c)

Added a new executeCommandAsync entry point that runs execute_command on a libuv worker thread and returns a Promise. This is distinct from the existing executeCommandStreaming, which uses execute_command_with_callback and isn't compatible with all commands.

CoreInterop

Exposed executeCommandAsync() that delegates to the new native entry point.

FoundryLocalManager

 - Kept create() as-is (blocking) for backward compatibility.
 - Added createAsync() that initializes via async native calls.
 - Added a pendingCreate guard so concurrent callers share the same in-flight initialization, matching the locking strategy in the C#, Rust, and Python SDKs.
 - create() throws a clear error if called while createAsync() is in flight to prevent double native initialization.

ModelLoadManager

load(), unload(), and listLoaded() now use executeCommandAsync for the core interop path. These methods already returned Promises, so no API change.

ModelVariant.download()

The no-progress path now uses executeCommandAsync. Already async, no API change.

Catalog

 - getModels() and getCachedModels() now use executeCommandAsync for the native fetches. Already async, no API change.
 - updateModels() shares an in-flight refresh promise so concurrent callers don't redundantly refetch and rebuild the model maps.
 - The constructor accepts an optional catalogName so the manager can pass it in (avoiding a sync native call during async initialization), with a fallback to the existing sync lookup. Backward compatible.

Backward compatibility

No API breakages. All existing callers of create() and load() continue to work unchanged.

### DIFF
--- a/sdk/js/native/foundry_local_napi.c
+++ b/sdk/js/native/foundry_local_napi.c
@@ -6,10 +6,11 @@
  *
  * Replaces the koffi FFI bridge with a lightweight native addon that
  * dynamically loads the FoundryLocalCore shared library at runtime and
- * exposes three JavaScript-callable functions:
+ * exposes the following JavaScript-callable functions:
  *
  *   loadLibrary(corePath, depPaths?)  – load native libs, resolve symbols
  *   executeCommand(cmd, dataJson)     – synchronous command execution
+ *   executeCommandAsync(cmd, dataJson) – async command execution (Promise)
  *   executeCommandWithBinary(cmd, dataJson, binaryBuf) – with binary payload
  *   executeCommandStreaming(cmd, dataJson, callback) – async + streaming
  */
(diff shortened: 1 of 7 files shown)


sdk/js/src/foundryLocalManager.ts
```diff
@@ -40,11 +53,45 @@ export class FoundryLocalManager {
...
+                .then(() => {
+                    FoundryLocalManager.instance = manager;
+                    return manager;
+                })
+                .finally(() => {
+                    FoundryLocalManager.pendingCreate = undefined;
+                });
+        }
+        return FoundryLocalManager.pendingCreate;
+    }
```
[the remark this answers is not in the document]
baijumeswani: Covering this in tests would be difficult since it would require significant testing infrastructure to add these kinds of tests. It is out of scope for this PR.

sdk/js/src/detail/coreInterop.ts
```diff
@@ -115,6 +116,15 @@ export class CoreInterop {
...
     }
 
+    /**
+     * Asynchronously execute a native command without blocking the event loop.
+     * Runs the native call on a libuv worker thread.
+     */
+    public executeCommandAsync(command: string, params?: any): Promise<string> {
+        const dataStr = params ? JSON.stringify(params) : '';
+        return this.addon.executeCommandAsync(command, dataStr);
+    }
```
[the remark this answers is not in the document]
baijumeswani: the native addon ships with the sdk. This should not happen.

sdk/js/src/catalog.ts
```diff
@@ -17,11 +17,12 @@ export class Catalog {
     private modelAliasToModel: Map<string, Model> = new Map();
     private modelIdToModelVariant: Map<string, ModelVariant> = new Map();
     private lastFetch: number = 0;
+    private updatePromise?: Promise<void>;
 
-    constructor(coreInterop: CoreInterop, modelLoadManager: ModelLoadManager) {
+    constructor(coreInterop: CoreInterop, modelLoadManager: ModelLoadManager, catalogName?: string) {
```
prathikr, line 22: why do we need this new input param?

sdk/js/native/foundry_local_napi.c
```diff
@@ -474,6 +475,176 @@ static napi_value napi_execute_command_with_binary(napi_env env,
     return result;
 }
 
+/* ── Async (non-streaming) work data ──────────────────────────────────── */
+
+typedef struct {
```
prathikr, line 480: can you add some links to docs on napi? seems we are getting deeply dependent on it so would be good to have references for future devs. can be done separately.

prathikr (approved):

sdk/js/src/catalog.ts
baijumeswani: The catalog is constructed when the FoundryLocalManager is created. The pull-request introduces createAsync on FoundryLocalManager. The new createAsync needs to call `get_catalog_name` asynchronously. To do that, we call executeCommandAsync("get_catalog_name") and pass it into the catalog constructor so the catalog needn't call executeCommand("get_catalog_name").

sdk/js/native/foundry_local_napi.c
baijumeswani: I'll do this separately by creating an md file.

PR #690, opened 2026-05-01
[JS & Python] Add cancellation and structured error handling to live audio streaming

rui-ren:
## Summary

Adds cancellation and structured error handling to the **JavaScript** and **Python** SDKs' live audio transcription clients, closing two cross-language API parity gaps from the realtime-audio analysis.

### Structured errors

| Language | Surface |
|---|---|
| JS | New `CoreError` class  `code` + `isTransient` from `CoreErrorResponse` exposed on every error thrown by `start` / `pushLoop` / `stop`. Re-exported from `foundry-local-sdk`. |
| Python | `FoundryLocalException` messages still parse via existing `CoreErrorResponse.try_parse`; sample now demonstrates how to use it for transient-error recovery. |

### Cancellation

| Language | Surface |
|---|---|
| JS | `audioClient.createLiveTranscriptionSession(signal?: AbortSignal)`  set once on the session, applies to every subsequent `start` / `append` / `stop` / `getTranscriptionStream` call. Mirrors C# `CancellationToken` shape. |
| Python | `audio_client.create_live_transcription_session(cancel_event: Optional[threading.Event] = None)`  same pattern. Backpressured `append()` and idle `get_transcription_stream()` poll a documented `_CANCEL_POLL_INTERVAL` (100 ms) when cancellation is configured. **Zero overhead** on the fast path. |

Both APIs are intentionally symmetric: cancellation is configured **once** on the constructor, not threaded through every method. Aborted `append()` does NOT enqueue the chunk (no risk of late delivery to native core). Aborted session also best-effort releases the native handle so it doesn't leak.

## Files changed

| File | Change |
|---|---|
| `sdk/js/src/openai/liveAudioTranscriptionTypes.ts` | New `CoreError` class + `wrapCoreError()` helper |
| `sdk/js/src/openai/liveAudioTranscriptionClient.ts` | `AbortSignal` plumbing, abort-aware `AsyncQueue.write()`, listener-leak fixes, native-handle release on abort |
| `sdk/js/src/openai/audioClient.ts` | `createLiveTranscriptionSession(signal?)` |
| `sdk/js/src/index.ts` | Re-exports `CoreError` |
| `sdk/js/test/openai/liveAudioTranscription.test.ts` | +`CoreError` tests, +AbortSignal listener-cleanup test (Proxy-wrapped to actually count listeners), +non-Error `signal.reason` test |
| `samples/js/live-audio-transcription/app.js` | Demonstrates `CoreError` + `AbortSignal` graceful shutdown |
| `sdk/python/src/openai/audio_client.py` | `create_live_transcription_session(cancel_event=...)` |
| `sdk/python/src/openai/live_audio_transcription_client.py` | `cancel_event` plumbing, `_CANCEL_POLL_INTERVAL` constant |
| `sdk/python/test/openai/test_live_audio_transcription.py` | +cancellation tests covering pre-set cancel, backpressure unblocking, generator clean exit |
| `samples/python/live-audio-transcription/src/app.py` | Demonstrates `cancel_event` + `CoreErrorResponse.try_parse` |

## Tests

- **JS:** 19/19 live-audio tests pass
- **Python:** 25/25 live-audio tests pass
- E2E synthetic-audio test still skips gracefully when the Nemotron model is not cached
- Both samples syntax-verified

## Backwards compatibility

100%  every new parameter is optional. Existing callers see no behavior change.

## Out of scope (parity report items not in this PR)

- Missing `id` field in C++/Rust `LiveAudioTranscriptionResponse`  addressed in #687 / #689.
- C++ polling vs. async streams  different SDK; separate refactor.


sdk/js/test/openai/liveAudioTranscription.test.ts
```diff
@@ -116,6 +116,123 @@ describe('Live Audio Transcription Types', () => {
...
+                } finally {
+                    if (onAbort) signal.removeEventListener('abort', onAbort);
+                }
+            }
+
+            const finalCount = (signal as any).listenerCount?.('abort') ?? 0;
+            expect(finalCount).to.equal(initialCount);
+        });
+
+        it('should propagate AbortError when signal is fired during race', async () => {
```
kunal-vaishnavi, line 210: Can you show how these changes would work in the JS example and how they are different than the below error handling?

https://github.com/microsoft/Foundry-Local/blob/58bba9383c80a8ec0fc9416db106ec736da56589/samples/js/live-audio-transcription/app.js#L69-L72

rui-ren: Great question  the existing catch block is fine for the happy path but throws away every signal we now expose. Pushed 83e8dc4 to the sample (samples/js/live-audio-transcription/app.js) so the diff is concrete:

**Before** (the lines you linked):
```js
} catch (err) {
    if (err.name !== 'AbortError') {
        console.error('Stream error:', err.message);
    }
}
```
This treats every non-abort error the same and loses `code` + `isTransient`.

**After**:
```js
} catch (err) {
    if (err.name === 'AbortError') return;
    if (err instanceof CoreError) {
        if (err.isTransient) {
            console.warn(` Transient ASR error (\): \. Continuing...`);
            return;  // recoverable  don't kill the app on a momentary native stall
        }
        console.error(` Stream error [\]: \`);
        return;
    }
    console.error(' Stream error:', err.message);
}
```

So the user-visible difference is:
- **Transient blips** (e.g., native-side recoverable hiccup) no longer silently log-and-die  the loop continues, which is what a customer building a long-running transcription UI actually wants.
- **Hard errors** include the machine-readable `code`, so customers can route on it (telemetry, retry policy, UI message) instead of regex-matching `err.message`.

For `AbortSignal`, I also wired it through to show what it buys:
```js
const shutdown = new AbortController();
await session.start(shutdown.signal);
for await (const r of session.getTranscriptionStream(shutdown.signal)) { ... }
await session.append(pcm, shutdown.signal);

process.on('SIGINT', async () => {
    shutdown.abort();              // unblocks any in-flight append/iterator
    if (audioInput) audioInput.quit();
    await session.stop();           // now drain-and-stop without racing
    ...
});
```
Previously, if `session.append()` was backpressured (waiting for native-core to drain), Ctrl+C had to wait for `stop()` to finish draining  which could take seconds. With `shutdown.abort()`, the in-flight append resolves immediately with AbortError, the pump exits, and we shut down cleanly.

Let me know if you'd like the sample changes split out into a separate PR.

rui-ren:
Both points addressed in df4260c.

**JS  session-level signal (no more passing it everywhere)**
``````ts
const shutdown = new AbortController();
const session = audioClient.createLiveTranscriptionSession({ signal: shutdown.signal });

await session.start();           // signal applies automatically
await session.append(pcm);       // signal applies automatically
for await (const r of session.getTranscriptionStream()) { ... }
await session.stop();

process.on('SIGINT', () => shutdown.abort());
``````

Per-call signals still work as overrides  if both session-level and per-call are set, they compose via `AbortSignal.any` so EITHER aborting cancels. Sample (`samples/js/live-audio-transcription/app.js`) now uses this one-line pattern.

**Python  cancellation parity (closes the gap from the analysis doc)**
``````python
cancel = threading.Event()
signal.signal(signal.SIGINT, lambda *_: cancel.set())

session = audio_client.create_live_transcription_session(cancel_event=cancel)
session.start()
session.append(pcm)
for result in session.get_transcription_stream():
    ...
session.stop()
``````

Same composition rule: each method also accepts an optional per-call `cancel_event`; either being set cancels. Backpressured `append()` and idle `get_transcription_stream()` poll on a 100 ms timeout when cancellation is configured; **zero overhead on the fast path with no cancel sources**. An aborted `append()` does NOT enqueue the chunk (no late delivery to native core).

**Tests:** JS 20/20, Python 26/26 (+4 new Python cancellation tests covering pre-set cancel, backpressure unblocking, and clean generator exit).

File: sdk/python/src/openai/live_audio_transcription_client.py [hunk 2]
```diff
@@ -141,7 +168,11 @@ def start(self) -> None:
             self._push_thread = threading.Thread(target=self._push_loop, daemon=False)
             self._push_thread.start()
 
-    def append(self, pcm_data: bytes) -> None:
+    def append(
+        self,
+        pcm_data: bytes,
+        cancel_event: Optional[threading.Event] = None,
```
kunal-vaishnavi: Can we revert the session API changes to pass in `cancel_event` since we already have a `cancel_event` in the session constructor? We want JS and Python to have similar APIs for start, stop, append, etc.

rui-ren: Done in eee2cbe. Reverted the per-call ``signal`` / ``cancel_event`` parameters on both JS and Python so cancellation is configured **once** on the constructor and the method signatures stay symmetric.

**JS**  ``start()``, ``append()``, ``stop()``, ``getTranscriptionStream()`` now take no abort parameter; they read ``this.sessionSignal`` directly. Removed the ``resolveSignal`` / ``AbortSignal.any`` composition machinery.

**Python**  same: ``start()``, ``append()``, ``stop()``, ``get_transcription_stream()`` take no ``cancel_event``. ``_is_cancelled()`` simplified to check only the session-level event.

Final API now matches across both languages:

``````ts
const session = audioClient.createLiveTranscriptionSession({ signal: shutdown.signal });
await session.start();
await session.append(pcm);
for await (const r of session.getTranscriptionStream()) { ... }
await session.stop();
``````

``````python
session = audio_client.create_live_transcription_session(cancel_event=cancel)
session.start()
session.append(pcm)
for r in session.get_transcription_stream(): ...
session.stop()
``````

Tests: JS 19/19, Python 25/25 (-2 tests for the dropped per-call paths; backpressure-unblock test still validates the session-level cancel works).

File: samples/js/live-audio-transcription/app.js [hunk 3]
```diff
@@ -39,9 +39,16 @@ console.log('Loading model...');
...
+// createLiveTranscriptionSession({ signal }) — every subsequent
+// start() / append() / getTranscriptionStream() / stop() call picks it
+// up automatically, so we don't have to thread the signal through every
+// callsite.
+const shutdown = new AbortController();
+
 // Create live transcription session (same pattern as C# sample).
 const audioClient = model.createAudioClient();
-const session = audioClient.createLiveTranscriptionSession();
+const session = audioClient.createLiveTranscriptionSession({ signal: shutdown.signal });
```
kunal-vaishnavi, line 52: Instead of creating a new class to represent the parameters passed into the session, can we instead make an optional parameter for `signal` in `createLiveTranscriptionSession`? Then, we can remove the need for the `LiveAudioTranscriptionSessionOptions` class. The new optional parameter for `signal` can behave similar to a `CancellationToken` parameter.

rui-ren: sure, I will update the `cancellation` pattern. Thanks

rui-ren: Done in 6ab38c4. Dropped ``LiveAudioTranscriptionSessionOptions`` and made ``signal`` a plain optional parameter on ``createLiveTranscriptionSession``, matching C#'s ``CancellationToken`` shape:

``````ts
const shutdown = new AbortController();
const session = audioClient.createLiveTranscriptionSession(shutdown.signal);
``````

(was ``createLiveTranscriptionSession({ signal: shutdown.signal })``)

JS 19/19 tests pass; sample updated.

File: sdk/python/src/openai/live_audio_transcription_client.py [hunk 4]
```diff
@@ -240,8 +287,26 @@ def get_transcription_stream(
...
+            return
+
+        # Cancellation-aware path: poll periodically so we can return cleanly
+        # when the cancel event fires.
         while True:
-            item = q.get()
+            if self._is_cancelled():
+                return
+            try:
+                item = q.get(timeout=0.1)
```
kunal-vaishnavi: What is the timeout for?

sdk/python/src/openai/live_audio_transcription_client.py
```diff
@@ -201,12 +231,24 @@ def append(self, pcm_data: bytes) -> None:
...
+        # Cancellation-aware path: poll with a small timeout so we can
+        # surface a cancel set from another thread without enqueuing the chunk.
         # Performed outside the lock to avoid blocking stop() and other
         # state transitions while waiting for queue space.
-        push_queue.put(data_copy)
+        while True:
+            if self._is_cancelled():
+                raise FoundryLocalException("append() cancelled before the chunk was enqueued.")
+            try:
+                push_queue.put(data_copy, timeout=0.1)
```
kunal-vaishnavi: Same question as [here](https://github.com/microsoft/Foundry-Local/pull/690/changes#r3177152536)

sdk/python/src/openai/audio_client.py
```diff
@@ -62,24 +62,38 @@ def __init__(self, model_id: str, core_interop: CoreInterop):
         self.settings = AudioSettings()
         self._core_interop = core_interop
 
-    def create_live_transcription_session(self) -> LiveAudioTranscriptionSession:
+    def create_live_transcription_session(
+        self,
+        cancel_event: Optional[threading.Event] = None,
```
kunal-vaishnavi, line 67: Same question as [here](https://github.com/microsoft/Foundry-Local/pull/690/changes#r3175732972) but for the Python example

rui-ren: Done in 143a8b6. Updated `samples/python/live-audio-transcription/src/app.py` with the same pattern as the JS sample  concrete diff so the API value is visible.

**Before:** the sample created the session with no cancellation token and used a parallel local `stop_event` for the mic loop, which couldn't reach inside the SDK:
``````python
session = audio_client.create_live_transcription_session()
...
stop_event = threading.Event()
def shutdown(*_):
    stop_event.set()  # only the mic loop sees this
    session.stop()    # has to drain everything
``````

**After:** one event drives the whole shutdown path  SIGINT just sets it:
``````python
shutdown_event = threading.Event()
session = audio_client.create_live_transcription_session(cancel_event=shutdown_event)
...
def shutdown(*_):
    shutdown_event.set()
    #  this single call:
    #   - aborts any in-flight session.append() blocked on backpressure
    #     (FoundryLocalException, no late delivery to native core)
    #   - ends session.get_transcription_stream() iteration cleanly
    #   - short-circuits session.stop()'s drain wait
    #   - exits the mic capture loop
    session.stop()
``````

I also added `CoreErrorResponse.try_parse` in the read loop to inspect structured error metadata (`code` + `is_transient`)  transient blips no longer kill long-running sessions, mirroring the JS `CoreError` demo.

Tests: 25/25 pass; sample syntax-verified.

File: sdk/python/src/openai/live_audio_transcription_client.py (hunk 4 above)
Comment (rui-ren) [reply sdk/python/src/openai/live_audio_transcription_client.py]: The 100 ms is a **cancellation polling interval**. Pushed d5683c3 to extract it as `_CANCEL_POLL_INTERVAL` with a docstring at the top of the file so this isn't a magic number anymore.

**Why we poll at all:** `queue.Queue.get` / `put` in standard Python can't be interrupted by a `threading.Event`  there's no equivalent of `select()` over them. So when the user has configured a `cancel_event`, we fall back to a poll-with-timeout pattern: wait up to 100 ms for queue I/O, then check `_is_cancelled()` and either return / raise or retry.

**Why 100 ms specifically:** balances responsiveness vs. overhead.
- Cancellation latency: a SIGINT takes effect within ~100 ms (one poll period).
- Idle CPU overhead: ~10 wakeups/sec per blocked call  negligible.

**Important:** this is a **no-op on the fast path**. If the user never passes a `cancel_event`, the original blocking `put()` / `get()` is used unchanged with no polling overhead at all (see the `if self._cancel_event is None` branch in each method).

File: sdk/python/src/openai/live_audio_transcription_client.py (hunk 5 above)
Comment (rui-ren) [reply sdk/python/src/openai/live_audio_transcription_client.py]: The 100 ms is a **cancellation polling interval**. Pushed d5683c3 to extract it as `_CANCEL_POLL_INTERVAL` with a docstring at the top of the file so this isn't a magic number anymore.

**Why we poll at all:** `queue.Queue.get` / `put` in standard Python can't be interrupted by a `threading.Event`  there's no equivalent of `select()` over them. So when the user has configured a `cancel_event`, we fall back to a poll-with-timeout pattern: wait up to 100 ms for queue I/O, then check `_is_cancelled()` and either return / raise or retry.

**Why 100 ms specifically:** balances responsiveness vs. overhead.
- Cancellation latency: a SIGINT takes effect within ~100 ms (one poll period).
- Idle CPU overhead: ~10 wakeups/sec per blocked call  negligible.

**Important:** this is a **no-op on the fast path**. If the user never passes a `cancel_event`, the original blocking `put()` / `get()` is used unchanged with no polling overhead at all (see the `if self._cancel_event is None` branch in each method).

**Review by kunal-vaishnavi — DISMISSED** [review]

File: samples/js/live-audio-transcription/app.js [hunk 7]
```diff
@@ -182,9 +206,14 @@ try {
...
 
-// Handle graceful shutdown
+// Handle graceful shutdown.
+//
+// The AbortController fires the shared `shutdown` signal so any in-flight
+// session.append() / getTranscriptionStream() resolves promptly with an
+// AbortError instead of waiting for stop() to finish draining the queue.
 process.on('SIGINT', async () => {
     console.log('\n\nStopping...');
+    shutdown.abort();
```
baijumeswani, line 217: Is my understanding correct that shutdown and session.stop do the same thing but shutdown is similar to session.stop(hard=true) so we avoid draining the queue? Should we use the same API to represent that, or do we need separate handlers for shutdown vs session.stop()?

samples/js/live-audio-transcription/app.js
```diff
@@ -5,7 +5,7 @@
 //
 // Usage: node app.js
 
-import { FoundryLocalManager } from 'foundry-local-sdk';
+import { FoundryLocalManager, CoreError } from 'foundry-local-sdk';
```
baijumeswani: CoreError seems to imply that the error is a generic error from the Core? Is that the case, or is the error specific to real time audio transcription?
If the latter, then I suggest we rename the error type? If the former, then we should address this in a bigger PR that throws a common error across all APIs.

samples/js/live-audio-transcription/app.js
```diff
@@ -67,9 +74,22 @@ const readPromise = (async () => {
...
-        if (err.name !== 'AbortError') {
-            console.error('Stream error:', err.message);
+        // AbortError is expected on Ctrl+C; ignore quietly.
+        if (err.name === 'AbortError') return;
+
+        // CoreError surfaces native-core failure metadata (code + isTransient).
+        // Use it to retry quietly on transient blips instead of dying on the
+        // first hiccup. Without CoreError the only signal would be err.message.
+        if (err instanceof CoreError) {
+            if (err.isTransient) {
```
baijumeswani: Is it expected for FoundryLocalCore to throw transient errors? What do transient errors mean? If we detect transient errors, and there is no action that the application/user should take during such errors, can we catch them before they reach the application layer and continue as necessary?

samples/python/live-audio-transcription/src/app.py
```diff
@@ -15,6 +15,8 @@
 import time
 
 from foundry_local_sdk import Configuration, FoundryLocalManager
+from foundry_local_sdk.exception import FoundryLocalException
+from foundry_local_sdk.openai.live_audio_transcription_types import CoreErrorResponse
```
baijumeswani: Same here. Why not encapsulate the error response in FoundryLocalException? Do we need a CoreErrorResponse?

sdk/js/src/openai/liveAudioTranscriptionTypes.ts
```diff
@@ -93,3 +93,38 @@ export function tryParseCoreError(errorString: string): CoreErrorResponse | null
...
 }
+
+/**
+ * Error thrown by live audio streaming operations when the native core reports a failure.
+ * Surfaces structured fields (code, isTransient) so callers can implement targeted retry
+ * or telemetry logic instead of string-matching on `message`.
+ *
+ * `code` is `'UNKNOWN'` when the underlying error is not a structured CoreErrorResponse.
+ */
+export class CoreError extends Error {
```
baijumeswani: This error name is generic. If this is truly a core error, we should move this to another file and make its contents more generic.

sdk/python/src/openai/live_audio_transcription_client.py
```diff
@@ -48,6 +48,19 @@
...
+# ``queue.Queue.get`` / ``put`` cannot be interrupted by a ``threading.Event``
+# in standard Python, so when a ``cancel_event`` is configured we fall back
+# to a poll-with-timeout pattern: wait up to this interval for queue I/O,
+# then check the cancel flag and either return / raise or retry.
+#
+# 100 ms balances cancellation latency (a SIGINT takes effect within ~100 ms)
+# against idle CPU overhead (~10 wakeups/sec per blocked call, negligible).
+# This is a no-op on the fast path where no cancel_event is configured —
+# the original blocking ``put()`` / ``get()`` is used unchanged.
+_CANCEL_POLL_INTERVAL = 0.1
```
baijumeswani: I think we should avoid any kind of polling in our SDK since it may add unnecessary complexity to our SDK. We are trying to re-create cancellation and are building the simplified infrastructure ourselves.

samples/js/live-audio-transcription/app.js
rui-ren: You're right — the name overpromises. Currently `CoreError` is only thrown by LiveAudioTranscriptionSession; chat, audio-file
 transcription, and the Responses API all still throw plain Error. We can do this ?
 
Rename to scope-honest: `LiveAudioStreamError` — keeps it where it lives, no false promises.

samples/js/live-audio-transcription/app.js
rui-ren: They're not the same thing — they're orthogonal:
 
 
  - `shutdown.abort()` cancels in-flight calls (e.g., `append()` blocked on backpressure, `getTranscriptionStream()` waiting for 
results).
  - `session.stop()` signals end-of-audio to native core, drains the queue, and releases the native handle.

samples/python/live-audio-transcription/src/app.py
rui-ren: Yes — this breaks symmetry with the JS CoreError class. Right now Python users have to do:
 ```
  except FoundryLocalException as ex:
      info = CoreErrorResponse.try_parse(str(ex))
      if info and info.is_transient: ...
 ```
 vs. JS:
 ```
  } catch (err) {
      if (err instanceof CoreError && err.isTransient) { ... }
  }
 ```
 
I will update it once we decide the name for `JS`

sdk/js/src/openai/liveAudioTranscriptionTypes.ts
rui-ren: You're right  pushed 187810b that renames `CoreError`  `LiveAudioStreamError` (and `wrapCoreError`  `wrapAsLiveAudioStreamError`). The class stays in `liveAudioTranscriptionTypes.ts` since it's only thrown by this API.

Why option (a)  rename  instead of (b)  move to a generic `errors.ts` and apply across all SDK APIs:
- (b) means touching chat, audio-file transcription, and the Responses API to also throw structured errors. That's a bigger refactor that should stand on its own PR.
- (a) is honest about today's reality: this error type is genuinely scoped to live audio streaming.

The structured `code` / `isTransient` fields are unchanged  this is a pure rename. Happy to file a follow-up issue for the generic-errors PR. WDYT?

sdk/python/src/openai/live_audio_transcription_client.py
rui-ren: Yes, we need to remove all the polling and keep the API clean.  Thanks.

rui-ren: You're right  pushed 26e2079 that replaces the polling-based cancellation with an event-driven `_CancellableQueue` backed by `threading.Condition`. No more polling anywhere.

**How it works:**
- `_CancellableQueue.put` / `get` block on a single `Condition`.
- One watcher thread per session waits on the user's `cancel_event` and calls `cancel()` on both queues when it fires; `cancel()` does `notify_all()` and every blocked thread wakes instantly.
- Removed `_CANCEL_POLL_INTERVAL`, all `timeout=...` polling loops, and the per-call cancellation-aware fast-path branches.

**Trade-off I want to flag:**
The watcher is one daemon thread per session. If `cancel_event` is never set (normal `stop()` path), it stays blocked on `Event.wait()` for the lifetime of the user's event object  daemon, so it doesn't block process exit, but it does retain a reference to `self`. Bounded leak: 1 thread per session, ~8 KB stack.

Python's stdlib doesn't have a clean primitive for "wait on either of two events" without polling or spawning extra helpers (I tried  every alternative carries equal or worse complexity). Documented this in the watcher's docstring.

**Verification:** added `test_cancellation_unblocks_instantly_no_polling` that asserts cancel latency < 50 ms (the old polling impl could take up to 100 ms), so we have a regression test if anyone re-introduces polling later. **26/26 tests pass.**

Let me know if the daemon-thread trade-off is OK or if you'd rather I push for one of the messier alternatives.

samples/js/live-audio-transcription/app.js
rui-ren: You're right  the transient branch was speculative (no documented list of transient codes from FLC core today, no SDK-internal retry policy yet). Pushed dfc9dee that drops the transient handling from both the JS and Python samples.

What changed:
- Removed the `if (err.isTransient) ...` branch from both samples
- Added a NOTE comment in each sample saying SDK-internal retry on transient errors is future work
- Samples still demonstrate the value of the structured error type  customers can log/route on `LiveAudioStreamError.code` (JS) or `CoreErrorResponse.try_parse(...).code` (Python) instead of regex-matching `err.message`

What stayed:
- `LiveAudioStreamError.isTransient` and `CoreErrorResponse.is_transient` are still on the SDK API surface (no breaking change). Advanced users can still inspect them. We just don't lead the typical sample reader to expect recovery behavior the SDK doesn't deliver.

When the SDK gets a real retry policy + a documented set of transient codes, we can re-introduce internal handling and the field becomes meaningful end-to-end. Happy to file a follow-up issue tracking that work  let me know.

baijumeswani:
If this is still relevant, could we please port this over to sdk v2?

PR #699, opened 2026-05-07
bump dev version from 1.1.0 --> 1.2.0

prathikr:


baijumeswani:
Shall we also update https://github.com/microsoft/Foundry-Local/blob/main/sdk/deps_versions.json?

sdk/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "foundry-local-sdk",
-  "version": "1.0.0",
+  "version": "1.2.0",
```
baijumeswani, line 3: Should this be 1.1.0?

baijumeswani: Nvm. I think 1.2.0 is correct

baijumeswani (approved):

baijumeswani (approved):

PR #705, opened 2026-05-11
C++ implementation of FL Core + SDK.

skottmckay:
C++ implementation of FL Core + SDK.
C# and python SDKs to use new setup.
Currently in sdk_v2 for ease of comparison. Will replace sdk implementation pre-checkin.

To test locally build the c++ sdk first (sdk_v2/cpp/build.{sh|bat}. 
C# and python bindings look for the default RelWithDebInfo build from that.

CI produces nuget package that can also be used for testing. Not published to a feed yet though.

Various design decisions are captured in sdk_v2/cpp/docs for reference.

The current implementation uses a static catalog listing. Will change to live catalog when possible. Easily updated if we have additional models to add.

Set the TEST_MODEL_CACHE_DIR environment variable to point to test-data-shared if you have that locally and want to avoid some model downloads.


sdk_v2/cpp/examples/basic_chat/main.cc
```diff
@@ -0,0 +1,207 @@
...
+#include <iostream>
+#include <string>
+
+using namespace foundry_local;
+
+/// Non-streaming chat example.
+/// Creates a session, sends a request, reads the full response, and does a follow-up turn.
+void BasicChat(IModel& model) {
+  // 1. Create a session for multi-turn chat.
+  ChatSession session(model);
```
prathikr, line 20: To keep this more consistent with v1 should it be ChatClient instead of ChatSession?

sdk_v2/cpp/examples/basic_chat/main.cc
```diff
@@ -0,0 +1,207 @@
...
+/// Sets a streaming callback on the session so incremental tokens are printed as they arrive.
+void StreamingChat(IModel& model) {
+  // 1. Create a session for multi-turn chat.
+  ChatSession session(model);
+
+  // 2. Set a streaming callback. Each invocation receives one item in the queue.
+  //    We pop the item from the queue and print any message content incrementally.
+
+  // lambda can capture user data if needed (equivalent to the additional `void* user_data` parameter in the C API).
+  session.SetStreamingCallback([/*user_data*/](flStreamingCallbackData event) -> int {
```
prathikr, line 90: we decided to go with generator format for streaming instead of lambda/callbacks. does cpp support this?

sdk_v2/cpp/examples/embeddings/main.cc
```diff
@@ -0,0 +1,132 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+//
+// Example: Text embeddings with the Foundry Local C++ SDK.
+// Demonstrates: create manager, find an embeddings model, download/load it,
+// generate embeddings for single and batch inputs, and compute cosine similarity.
+
+#include <foundry_local/foundry_local_cpp.h>
```
prathikr, line 8: Should we reach out to Raja to review embeddings?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -0,0 +1,885 @@
...
+#include <specstrings.h>
+#define ORT_ALL_ARGS_NONNULL
+#endif
+
+/* -----------------------------------------------------------------------
+ * Foundry Local SDK API Version
+ * Incremented with each release.
+ * Used to request the API function table via FoundryLocalGetApi.
+ * ----------------------------------------------------------------------- */
+#define FOUNDRY_LOCAL_API_VERSION 1
```
prathikr, line 62: 2?

sdk_v2/python/_build_backend/__init__.py
```diff
@@ -0,0 +1,171 @@
+# PEP 517 backend shim for foundry-local-sdk.
```
prathikr, line 1: can we add or port over a python example?

sdk_v2/python/src/foundry_local_sdk/_native/api.py
```diff
@@ -0,0 +1,145 @@
...
+from __future__ import annotations
+
+import importlib
+import os
+
+from foundry_local_sdk._native.lib_loader import find_library, prepare_native_dependencies
+from foundry_local_sdk.exception import FoundryLocalException
+
+# FOUNDRY_LOCAL_API_VERSION = 1 (from foundry_local_c.h)
+_FOUNDRY_LOCAL_API_VERSION: int = 1
```
prathikr, line 20: 2?

sdk_v2/python/_build_backend/__init__.py
prathikr: added to onenote

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
prathikr: added to onenote

sdk_v2/python/src/foundry_local_sdk/_native/api.py
prathikr: added to onenote

sdk_v2/cpp/examples/embeddings/main.cc
prathikr: added to onenote

sdk_v2/cpp/examples/basic_chat/main.cc
prathikr: added to onenote

sdk_v2/cpp/examples/basic_chat/main.cc
prathikr: added to onenote

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -0,0 +1,885 @@
...
+/* --- Root API ---------------------------------------------------------- */
+typedef struct flApi {
+  /* Status */
+  FL_API_STATUS(Status_Create, flErrorCode error_code, _In_ const char* error_msg);
+  FL_TYPE_RELEASE(Status);
+  flErrorCode FL_API_T(Status_GetErrorCode, _In_ const flStatus* status);
+  const char* FL_API_T(Status_GetErrorMessage, _In_ const flStatus* status);
+
+  /* Manager lifecycle */
+  FL_API_STATUS(Manager_Create, _In_ const flConfiguration* config, _Outptr_ flManager** out_manager);
```
baijumeswani, line 544: Does Manager really need to be a Singleton? Could we decouple real process globals from the manager?

If we really feel like the singleton pattern is a better fit (so it owns the ortenv and other process globals), could we document that in the c api?

Another point to note is that Manager_Create called a second time will return an error. So, it becomes difficult for a single process that has two different usages of Foundry Local (maybe as two different plug-ins to an application that share the process) to both acquire the manager. It also becomes difficult to know which instance should release the manager if somehow the manager instance can be shared. Should we consider always returning the manager and internally refcount the number of manager creations and releases?

prathikr:
Can we align the naming of the artifacts to be the same as the existing pipeline so that the publishing pipeline can be re-used to publish to ORT-Nightly or Public Feeds? 

https://github.com/ai-platform-microsoft/onnxruntime-release-pipelines/blob/95661fb56d66a0bdc874c87e9a70b7d4897c3c51/foundry-local/foundry-local-publishing.yml#L6

prathikr:
@baijumeswani noticed the Core package was renamed to Microsoft.AI.Foundry.Local.Runtime. Any particular reason for that? We think that using the existing Microsoft.AI.Foundry.Local.Core naming would make it easier for customers to transition to the new core without them needing to change their package dependencies.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -0,0 +1,885 @@
...
+/* -----------------------------------------------------------------------
+ * Callback types
+ * ----------------------------------------------------------------------- */
+
+/** Progress callback for model downloads. value is 0.0 to 100.0.
+ *  Return 0 to continue, non-zero to cancel the operation reporting progress. */
+typedef int (*flProgressCallback)(float value, void* user_data);
+
+/** Streaming response event callback. Return 0 to continue, non-zero to cancel. */
+typedef int (*flStreamingCallback)(flStreamingCallbackData event, void* user_data);
```
baijumeswani, line 510: Should flStreamingCallbackData be passed in as a const pointer?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -0,0 +1,885 @@
...
+
+  /* EP detection */
+
+  /// Get discoverable execution providers and their registration status.
+  /// Returns parallel arrays of EP names and registration flags, plus count.
+  /// Returned data is owned by the Manager — valid until the next call to
+  /// Manager_GetDiscoverableEps or Manager destruction.
+  FL_API_STATUS(Manager_GetDiscoverableEps, _In_ const flManager* manager,
+                _Out_ const char* const** out_names,
+                _Out_ const int** out_is_registered,
```
baijumeswani: Some bool types are represented as int* input arguments. While some other methods seem to use bool in their return types? Shall we keep bool where it makes sense to instead of using ints?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -0,0 +1,885 @@
...
+                _In_opt_ const char* const* ep_names,
+                size_t num_ep_names,
+                _In_opt_ flEpProgressCallback callback,
+                _In_opt_ void* user_data);
+
+  /// Whether an EP download/registration operation is currently in progress.
+  bool FL_API_T(Manager_IsEpDownloadInProgress, _In_ const flManager* manager);
+
+  /// Begin graceful shutdown. Safe to call from any thread. Idempotent.
+  FL_API_STATUS(Manager_Shutdown, _In_ flManager* manager);
```
baijumeswani, line 611: Could we add a note about the differences between shutdown and manager_release?
I think having both makes sense. But to a consumer, it might be meaningful to understand when to call shutdown.

baijumeswani, line 611: On another note, does Manager_Release also release process globals? Can one re-create a manager after releasing it?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -0,0 +1,885 @@
...
+typedef struct flInferenceApi flInferenceApi;
+typedef struct flModelApi flModelApi;
+
+/* --- Root API ---------------------------------------------------------- */
+typedef struct flApi {
+  /* Status */
+  FL_API_STATUS(Status_Create, flErrorCode error_code, _In_ const char* error_msg);
+  FL_TYPE_RELEASE(Status);
+  flErrorCode FL_API_T(Status_GetErrorCode, _In_ const flStatus* status);
+  const char* FL_API_T(Status_GetErrorMessage, _In_ const flStatus* status);
```
baijumeswani, line 541: Could we add a note on the lifetime of the const char* (for all such functions that return a const char*)?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
baijumeswani: Instead of returning names and is_registered separately, could we return a type here (a list of epinfo). Then expose functions on each epinfo for name, is registered (other fields in the future)?

The list of epinfo can be owned by the manager and can also be made threadsafe with this change?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: We only allow one instance at a time to keep things simple. What if someone creates two manager instances with overlapping config values (e.g. using same cache directory)? What behavior do you expect if you create 2 and they both load the same model? If they're sharing the model, now you also need to track how many managers are using it to know when you can really unload it.  

Supporting multiple calls to create gets more complicated as you'd need logic to compare configurations to decide if it's the same as the current manager, or (assuming you allowed multiple manager instances) a valid configuration to create a new manager with.

These additions aren't necessarily hard to do, but they're adding more complexity and potential places for issues.
Is there a specific user scenario that requires multi-managers? If not ID prefer to minimize complexity in the first release and we can expand what's allowed when needed, and this is equivalent to the current setup in Core. 

We can definitely add more documentation though.

skottmckay:
> Can we align the naming of the artifacts to be the same as the existing pipeline so that the publishing pipeline can be re-used to publish to ORT-Nightly or Public Feeds?
> 
> https://github.com/ai-platform-microsoft/onnxruntime-release-pipelines/blob/95661fb56d66a0bdc874c87e9a70b7d4897c3c51/foundry-local/foundry-local-publishing.yml#L6

I think we can now. It was building both when first added and I wanted a clear way to know if something came from the new version or not. Consider all the pipeline stuff to just be "make it work for initial testing" not "this is the proposed final setup" so we can change anything there.

sdk_v2/cpp/examples/basic_chat/main.cc
skottmckay: We could potentially add something like that to the C++ wrapper. right now that's C++17 for maximum compatibility so would be a more complicated to do so without things like co_yield/co_return/co_await from C++20 or std::generator from C++23. 

but that's all stuff that is built above the C API which afaik would always be limited to returning streamed output via a callback.

sdk_v2/cpp/examples/basic_chat/main.cc
skottmckay: Deliberately different. The 'client' based types in V1 are OAI type based. This is a completely different world (Item/Session/Request/Response) that will co-exist in parallel.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: item_queue that it provides needs to be mutable so receiver can pop Item/s from it and check if it's marked as finished.

we _could_ hide the item queue by providing 2 functors (pop/is_finished), but that feels like it might be unnecessarily complicated.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: problematic for this scenario as the underlying storage is a vector and vector<bool> doesn't convert to bool* (it does not have a `data()` member as the implementation packs bits). could use `uint8_t` for slightly more clarity though.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: Should be able to re-create a manager after the previous one is released.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: Anything returning a const char* owns it (user doesn't free as it's const), but that also means if you call some other function that changes the underlying value you potentially invalidate the returned pointer. e.g. you restart the web service, it's using an emphemeral port, so values previously returned by Manager_WebServiceUrls become invalid, or you remove a model from the cache so a previous Model GetPath return value becomes invalid. 

Do you mean an overall note to that effect? or something more function specific?

Do we expect anyone to directly use the C API (more documentation potentially required) or that the C++/C#/Python/JS/Rust SDKs will be the user facing surface?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: sure. can this be a simple versioned data struct instead of fully opaque type?

skottmckay:
> @baijumeswani noticed the Core package was renamed to Microsoft.AI.Foundry.Local.Runtime. Any particular reason for that? We think that using the existing Microsoft.AI.Foundry.Local.Core naming would make it easier for customers to transition to the new core without them needing to change their package dependencies.

I slightly prefer Runtime as it's created from a completely different codebase, but not overly fussed. 

I would have expected Core to be an indirect dependency that the user doesn't need to manually add. What's the scenario where referencing the FL SDK package requires manually adding Core?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: actually the implementation could be pushed down (same with Manager_DownloadAndRegisterEpsImpl - too much implementation in c_api.cc which should be as thin as possible) and we could be smarter about clearing existing values. 

assuming the list of EPs never shrink (AFAIK it never would) we could `reserve` a vector for the info in the implementation (maybe 8 slots as that should never be exhausted) and only append new values to it. that would get rid of the unnecessary 'invalidated on next call' behavior it currently has.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: Design choice to be made. Does the API version need to align with the major version or not? 

I would lean towards 'not' so we only rev the version if we make a change, and if we really really needed to make a change in a minor version we could.

sdk_v2/python/_build_backend/__init__.py
skottmckay: Added

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -0,0 +1,1017 @@
...
+ private:
+  std::unique_ptr<detail::StreamingCallbackHelper> streaming_callback_helper_;
+};
+
+class ChatSession : public Session {
+ public:
+  explicit ChatSession(IModel& model);
+
+  /// Add a tool definition that is available for the entire session.
+  ChatSession& AddToolDefinition(const ToolDefinition& tool_definition);
```
sayanshaw24, line 942: `AddToolDefinition` is append-only - for multi-phase conversations, would callers must create a new session to narrow the tool set, losing context/history? should we add session `ClearToolDefinitions` or `RemoveToolDefinition` at the C layer?

sdk_v2/cpp/src/inferencing/generative/chat/search_options.cc
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  // top_p
+  if (options.top_p.has_value()) {
+    gen_params.SetSearchOption("top_p", static_cast<double>(*options.top_p));
+  }
+
+  // top_k
+  if (options.top_k.has_value()) {
+    gen_params.SetSearchOption("top_k", static_cast<double>(*options.top_k));
```
sayanshaw24, line 65: nit: `SearchOptions` wraps temperature/top_p/etc. but omits tool_choice. Consider adding:

  `std::optional<ToolChoice> tool_choice;  // enum { Auto, None, Required }`

so typos like `"requird"` are caught at compile time. The raw string path via
KeyValuePairs can remain for advanced use.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -0,0 +1,1017 @@
...
+
+  /// True when the message has exactly one part and it is a TEXT item.
+  bool IsSimpleText() const;
+
+  /// Text of the single TEXT part. Throws if !IsSimpleText().
+  std::string GetSimpleText() const;
+};
+
+/// Content returned from a TOOL_CALL item.
+struct ToolCallContent {
```
sayanshaw24, line 443: nit: `ToolCallContent` and `ToolResultContent` store `std::string` while `TextContent` uses
`std::string` (with move) and `ImageContent`/`AudioContent` use `string_view`. If the
item owns the backing buffer, `string_view` would be consistent and avoid copies.
If ownership transfer is intentional here, should we add a comment explaining why?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -0,0 +1,903 @@
...
+  flItemQueue* item_queue;  // One item is added to the queue for each callback.
+  // TBD if we need output_index of content_index values at the Session.Run callback level or whether that is
+  // handled above when converting to chat completions or responses api events.
+} flStreamingCallbackData;
+
+typedef struct flToolDefinition {
+  uint32_t version;         ///< Set to FOUNDRY_LOCAL_API_VERSION;
+  const char* name;         ///< Tool name.
+  const char* description;  ///< Tool description for model context.
+  const char* json_schema;  ///< JSON schema defining the tool's arguments.
```
sayanshaw24, line 497: `json_schema` is a raw `const char*` with no documented behavior for malformed input.
can we either validate JSON syntax in `Session_AddToolDefinition` (return error status) or
document that invalid schema is passed through and may cause model-level errors?

sdk_v2/cpp/test/sdk_api/chat_session_test.cc
```diff
@@ -0,0 +1,603 @@
...
+      std::cout << "Tool call: " << tc.name << "(" << tc.arguments << ")\n";
+      found_tool_call = true;
+      break;
+    }
+  }
+
+  EXPECT_TRUE(found_tool_call) << "No TOOL_CALL item in response";
+}
+
+TEST_F(ToolCallFixture, ToolCallWithResult) {
```
sayanshaw24, line 268: streaming + tool-call assembly is untested - there is no test exercising streaming callbacks with tool calls; ccallers need to know how partial tool-call JSON arrives (single complete item? multiple chunks?). perhaps we can add a test that sets a streaming callback, forces a tool call, and asserts on the items delivered via the `ItemQueue`?

sdk_v2/cpp/test/sdk_api/chat_session_test.cc
```diff
@@ -0,0 +1,603 @@
...
+          "second": { "type": "integer", "description": "The second number in the operation" }
+        },
+        "required": ["first", "second"]
+      })"});
+
+  Request request1{
+      SystemMessage("You are a helpful AI assistant. If necessary, you can use any provided tools to answer the question."),
+      UserMessage("What is the answer to 7 multiplied by 6?"),
+  };
+  request1.SetOptions({{FOUNDRY_LOCAL_PARAM_TOOL_CHOICE, "required"},
```
sayanshaw24, line 290: Turn 1 sets `tool_choice="required"`. Turn 2 omits it. The test doesn't assert whether "required" persists or resets to "auto". should we add a brief comment or explicit assertion documenting the expected scoping behavior?

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
skottmckay: Yes. What's going to be more useful? Clear + re-add or remove by name?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: Should validation be done at the lowest level (GenAI) to prevent model-level errors? Otherwise we're duplicating logic in higher layers if we validate there.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
skottmckay: Great catch. Underlying flItem owns the buffer so all the ...Content structs should be able to use string_view.

baijumeswani (approved):
We can address all open concerns/issues in a follow up PR.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
skottmckay: adding to onenote to address in follow up PR

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
skottmckay: updated. will be in follow up PR

sdk_v2/cpp/test/sdk_api/chat_session_test.cc
skottmckay: session options persist. per-request options are overrides and aren't sticky. 

I can make the this more explicit in the comments of the public APIs. that's probably more useful

sdk_v2/cpp/test/sdk_api/chat_session_test.cc
skottmckay: will add. do we not have that in any other language bindings either?

sdk_v2/cpp/test/sdk_api/chat_session_test.cc
skottmckay: Reworked the public API a little to have a dedicated RequestOptions type that now contains SearchOptions, tool calling mode, and arbitrary options so the SetOptions for Session and Request take a more concrete type. It was a little backwards with a helper type generating the property bag to call the SetOptions with.

Added comments to all SDKs calling out using in a Request is per-request only.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
skottmckay: added RemoveToolDefinition. if needed that can support taking a wildcard for the tool name to implement Clear so we don't need a separate function for that.

sdk_v2/cpp/test/sdk_api/chat_session_test.cc
skottmckay: good catch. definite gap. have addressed. will now accumulate and send ToolCallItem to streaming callback when complete. they are also returned in the overall response.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: added comments to a bunch of C API items about ownership

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: simplified implementation significantly when pushing down. added EpInfo type.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
sayanshaw24: Agreed - Extensions also validates other downstream json for chat templates and tool calling so that should actually already be handled there, we can validate any other peripherals in GenAI.

PR #723, opened 2026-05-22
support multi-lingual ASR models

rui-ren:
## Summary

Update live audio transcription samples to use the correct Foundry Local catalog alias for the multilingual Nemotron ASR model, and strip model-emitted language tags in the SDK so all consumers receive clean transcription text.

### Samples (all 5 languages)
- Updated the commented multilingual model alias from `Nemotron-3.5-ASR-Streaming-Multilingual-0.6b-onnx-int4` to `nvidia-nemotron-3.5-asr-streaming-multilingual-0.6b` to match the FL catalog naming convention (mirrors the HuggingFace repo name)
- Affected files:
  - `samples/cs/live-audio-transcription/Program.cs`
  - `samples/js/live-audio-transcription/app.js`
  - `samples/python/live-audio-transcription/src/app.py`
  - `samples/cpp/live-audio-transcription/main.cpp`
  - `samples/rust/live-audio-transcription/src/main.rs`

### SDK (`sdk/cs/src/OpenAI/LiveAudioTypes.cs`)
- Strip language identification tags (e.g. `<zh-CN>`, `<en-US>`) from transcription text in `LiveAudioTranscriptionResponse.FromJson()`
- Multilingual Nemotron models emit these tags at segment boundaries as language identifiers — following industry convention (Whisper, Azure Speech, Google STT), these should be returned as metadata rather than inline text
- All SDK consumers now receive clean text without needing client-side post-processing

## Testing

Verified end-to-end with `nvidia-nemotron-3.5-asr-streaming-multilingual-0.6b` (INT4) model:
- English speech → clean English transcription
- Chinese speech (`zh-CN`) → clean Chinese transcription
- Mixed language switching works correctly
- Language tags stripped at the SDK layer before reaching consumers


sdk/cs/src/OpenAI/LiveAudioTypes.cs
```diff
@@ -38,6 +42,9 @@ internal static LiveAudioTranscriptionResponse FromJson(string json)
             JsonSerializationContext.Default.LiveAudioTranscriptionRaw)
             ?? throw new FoundryLocalException("Failed to deserialize live audio transcription result");
 
+        // Strip language tags emitted by multilingual models (e.g. <zh-CN>, <en-US>)
+        var text = LangTagRegex.Replace(raw.Text, " ").Trim();
```
baijumeswani: What do these tags represent in the output text (is it the detected language)? Does that need to be somehow communicated to the application?

baijumeswani: Does doing regex matching on every transcription cause any performance degradation?

nenad1002: I think we need to skip special tokens, we can build a list, and if there is a special token coming, just skip it. Regex is probably not that slow compared to actual inference, but its probably not the most optimal solution.

rui-ren: > I think we need to skip special tokens, we can build a list, and if there is a special token coming, just skip it. Regex is probably not that slow compared to actual inference, but its probably not the most optimal solution.

Yes, that is optimal. 

Do you agree on this? @kunal-vaishnavi , I can implement it in Core/GenAI after the release.

rui-ren: in core:

```
// In AudioStreamingSession.DecodeTokens(StringBuilder sb)
private void DecodeTokens(StringBuilder sb)
{
    var tokens = _generator.GetStepTokens();
    foreach (var tokenId in tokens)
    {
        var decoded = _tokenizerStream.Decode(tokenId);
        
        // Skip language identification tokens emitted by multilingual models
        // (e.g. "<zh-CN>", "<en-US>"). These are single tokens that decode
        // to the full tag string.
        if (decoded.Length <= 8 && decoded.StartsWith('<') && decoded.EndsWith('>'))
            continue;
        
        sb.Append(decoded);
    }
}
```

The will solve the <zh-CN> language tag issue, and more elegant than fixing in GenAI C++ code. @nenad1002

rui-ren: @baijumeswani @nenad1002 @kunal-vaishnavi @hanbitmyths @skottmckay 

I recommend to keep the regex in the SDK code, as the latency is negligible. We can refactor in Core after the release ? 

### Performance Benchmark: Language Tag Stripping

Measured over 1M iterations (Python `timeit`), comparing three approaches for stripping language tags (e.g. `<zh-CN>`, `<en-US>`) emitted by the multilingual Nemotron model:

| Method | Per-call | At 100 tokens/sec | vs Inference (10–50 ms/token) |
|---|---|---|---|
| No processing | 0.35 µs | baseline | — |
| String check (`StartsWith`/`EndsWith`) | 0.95 µs | 0.095 ms/sec | 0.001% |
| Regex (`\s*<[a-z]{2}(-[A-Z]{2})?>\s*`) | 6.09 µs | 0.609 ms/sec | 0.006% |

**Conclusion:** Regex adds **6 microseconds per token**. Model inference takes **10,000–50,000 microseconds per token**. The regex overhead is **0.006%** of inference time — completely negligible. Not worth the complexity of moving to Core/GenAI token-level filtering for a 6 µs saving on a 10 ms+ operation.

samples/cs/live-audio-transcription/Program.cs
```diff
@@ -24,7 +24,10 @@
 
 var catalog = await mgr.GetCatalogAsync();
 
+// English-only:
 var model = await catalog.GetModelAsync("nemotron-speech-streaming-en-0.6b") ?? throw new Exception("Model \"nemotron-speech-streaming-en-0.6b\" not found in catalog");
+// Multi-lingual (supports 30+ languages including auto-detect):
+// var model = await catalog.GetModelAsync("nvidia-nemotron-3.5-asr-streaming-multilingual-0.6b") ?? throw new Exception("Model \"nvidia-nemotron-3.5-asr-streaming-multilingual-0.6b\" not found in catalog");
```
baijumeswani: Why is this commented out?

baijumeswani: ok, seems like it is a way to inform users of the model they should use.
Could you maybe extract the model alias in a variable? And then add a comment showing a different alias. You can update this separately.

samples/cs/live-audio-transcription/Program.cs
```diff
@@ -44,6 +47,10 @@ await model.DownloadAsync(progress =>
 session.Settings.SampleRate = 16000;  // Default is 16000; shown here to match the NAudio WaveFormat below
 session.Settings.Channels = 1;
 session.Settings.Language = "en";
+// Multi-lingual examples:
+// session.Settings.Language = "de";     // German
```
baijumeswani, line 53: Can you add a note in the SDK readme about the allowed language strings? It is not obvious what the strings should be. It can be done separately if you like.

baijumeswani (dismissed):

sdk/cs/src/OpenAI/LiveAudioTypes.cs
nenad1002: To answer another question, there is really no reason to output tags, EXCEPT in one scenario. 

If a user selects autodetect (I think thats lang ID 101), then this tag would help you tell what does the model think is the spoken language. Useful for debugging, otherwise not really useful.

samples/cs/live-audio-transcription/Program.cs
rui-ren: The sample defaults to the `English-only` model (nemotron-speech-streaming-en-0.6b) 

since it's smaller and most users will start with English. The multilingual line is commented as a `ready-to-use` alternative — users just uncomment it and comment the English line to switch.

samples/cs/live-audio-transcription/Program.cs
rui-ren: Ok, I will make a separate PR. Thanks!

sdk/cs/src/OpenAI/LiveAudioTypes.cs
hanbitmyths: We can go with this solution for now, but performance is not only concern. The language code is a standard, but regex itself would not be a standard. Let's design properly as a follow-up.

rui-ren: > We can go with this solution for now, but performance is not only concern. The language code is a standard, but regex itself would not be a standard. Let's design properly as a follow-up.

Sure, I will make follow-up PRs. Thanks

rui-ren: > To answer another question, there is really no reason to output tags, EXCEPT in one scenario.
> 
> If a user selects autodetect (I think thats lang ID 101), then this tag would help you tell what does the model think is the spoken language. Useful for debugging, otherwise not really useful.

Ok, I think we just strip the tag so far? 

later expose `DetectedLanguage` for reference to users?

samples/cs/live-audio-transcription/Program.cs
```diff
@@ -24,7 +24,10 @@
 
 var catalog = await mgr.GetCatalogAsync();
 
+// English-only:
 var model = await catalog.GetModelAsync("nemotron-speech-streaming-en-0.6b") ?? throw new Exception("Model \"nemotron-speech-streaming-en-0.6b\" not found in catalog");
```
bmehta001: Nit: Can you use something like what you used for the C++ sample here for consistency?

```
// English-only:
var modelAlias = "nemotron-speech-streaming-en-0.6b";
// Multi-lingual (supports 30+ languages including auto-detect):
// var modelAlias = "nvidia-nemotron-3.5-asr-streaming-multilingual-0.6b";
````

File: samples/rust/live-audio-transcription/src/main.rs [hunk 5]
```diff
@@ -57,7 +60,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("✓ Model loaded\n");
 
     let audio_client = model.create_audio_client();
-    let session = Arc::new(audio_client.create_live_transcription_session());
+    let session = audio_client.create_live_transcription_session();
+    // session.settings.language = Some("en".into());    // English (default)
```
bmehta001: Nit: Should the `English` line be uncommented for consistency with other SDKs?

bmehta001: Also, if it English the `default`, maybe you can add that as part of the default for the comment you have for the other SDKs

samples/rust/live-audio-transcription/src/main.rs
```diff
@@ -57,7 +60,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("✓ Model loaded\n");
 
     let audio_client = model.create_audio_client();
-    let session = Arc::new(audio_client.create_live_transcription_session());
+    let session = audio_client.create_live_transcription_session();
```
bmehta001: This may need to be `let mut session` in case users comment out one of the languages below

sdk/cs/src/OpenAI/LiveAudioTypes.cs
```diff
@@ -15,6 +16,9 @@ namespace Microsoft.AI.Foundry.Local.OpenAI;
 /// </summary>
 public class LiveAudioTranscriptionResponse : ConversationItem
 {
+    // Multilingual Nemotron models emit language tags like <zh-CN>, <en-US> at segment boundaries.
+    // Strip them so consumers get clean text.
+    private static readonly Regex LangTagRegex = new(@"\s*<[a-z]{2}(-[A-Z]{2})?>\s*", RegexOptions.Compiled);
```
bmehta001: Should something like this be in the other SDKs as well?

bmehta001: Should the regex support matching `<auto>`? Or is that not an actual tag that gets outputted

rui-ren: `<auto>` is never emitted by the model — it's only used as an input setting (language = "auto") to enable auto-detection. When auto-detecting, the model outputs the actual detected language tag (e.g. `<zh-CN>, <en-US>`), which is already matched by the current regex. No change needed.

rui-ren: For JS, Python, Rust, and C++ — the live audio streaming responses are parsed by Core (native AOT DLL) and returned as JSON strings. The C# SDK is the first consumer to parse these into typed objects, so this is where the tag stripping lives today. The other SDK wrappers would need equivalent handling when they add their own [LiveAudioTranscriptionResponse] parsing. I'll track that as a follow-up.

baijumeswani: I am not sure I understand this comment. Do other SDK languages not require doing this stripping of detected language? Or is it done in core for these SDK languages?

baijumeswani: It seems like all language bindings have this type LiveAudioTranscriptionResponse. And I guess all of them should address this?

kunal-vaishnavi (approved):

PR #725, opened 2026-05-26
Enable linux-arm64 support

baijumeswani:


prathikr (dismissed):
will need to re-run once ort-core and ort-genai-core publish, will likely fail right now...

prathikr (approved):

PR #736, opened 2026-05-28
[SDKv2] C++ SDK PR follow up

skottmckay:
### Refactor usage of options for session and requests
- create RequestOption type
  - add SearchOptions, tool choice and generic catch all
- use RequestOptions in Session.SetOptions and Request.SetOptions in public API
  - much better typed signal to users as to what is available
  - unfortunately touches a lot of test code but change is trivial and hopefully shows that the new usage is more user friendly

### Fix tool call when streaming
- direct ChatSession usage wasn't setup to detect tool call output from model
  - add infrastructure to accumulate the tool call and return to streaming callback
  - add tests and update tool calling example to include streaming usage

### Add ability to remove tool definition from session level
- remove by name
- can add wildcarded remove if 'clear' operation is needed without a change to the API surface

### Make ToolChoice and enum
- make options explicit and typed to prevent accidental misuse

### Use std::string_view consistently for the Content types in C++ wrapper
- owning Item that the content comes from owns the string in all cases

### Add more comments around output pointer lifetime and manager usage.


sayanshaw24 (dismissed):
looks good, thanks for addressing all the comments Scott! wanted to confirm we ran all the relevant tests locally to ensure everything is still functioning as expected following these changes?

sayanshaw24 (approved):

PR #746, opened 2026-05-30
[SDKv2] Add speech result types.

skottmckay:


sdk_v2/cpp/src/inferencing/generative/audio/SPEECH_TYPES.md
```diff
@@ -0,0 +1,129 @@
...
+  // so we can't remove anything added.
+  std::optional<float> confidence;        // 0..1 aggregate
+  std::optional<std::string> language;    // per-segment, for code-switching
+  std::optional<std::string> speaker_id;
+  std::optional<std::int32_t> channel;
+  // we could maybe use something more generic if we want to report these things instead of having per-value fields
+  // e.g. shared float[] of fixed size and an enum saying which value is in which slot.
+  std::optional<float> avg_logprob;       // Whisper-family diagnostic
+  std::optional<float> no_speech_prob;    // Whisper-family diagnostic
+  std::optional<float> compression_ratio; // Whisper-family diagnostic
```
skottmckay: Remove these. 
- speaker_id at the word level not segment
- unknown currently how/if we'd use `channel`
- We'll use a separate item type for diagnostics if needed

sdk_v2/cpp/src/inferencing/generative/audio/SPEECH_TYPES.md
```diff
@@ -0,0 +1,129 @@
...
+  // Utterance start signal — tagged on the first kPartial of a new utterance.
+  // Knowable at emission time. End-of-utterance is implicit (see design rules).
+  bool utterance_start = false;
+
+  std::vector<SpeechWord> words;          // word-timestamp opt-in
+
+  // Future / opt-in. Included here for visibility in review. 
+  // We should only add fields that we expect to use as the C API types need to be ABI stable,
+  // so we can't remove anything added.
+  std::optional<float> confidence;        // 0..1 aggregate
```
skottmckay: Also going to remove `confidence` given that's available at the word level.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -492,6 +495,70 @@ typedef struct flToolResultData {
...
+
+/// Versioned struct for a single word within a speech segment.
+/// All optional fields use sentinels (FOUNDRY_LOCAL_DURATION_UNSET / NULL) when absent.
+typedef struct flSpeechWord {
+  uint32_t version;        ///< Set to FOUNDRY_LOCAL_API_VERSION.
+  const char* text;        ///< UTF-8 word text. Always populated.
+  int64_t start_time_ms;   ///< Milliseconds from audio start. FOUNDRY_LOCAL_DURATION_UNSET if absent.
+  int64_t end_time_ms;     ///< Milliseconds from audio start. FOUNDRY_LOCAL_DURATION_UNSET if absent.
+  bool has_confidence;     ///< True iff `confidence` is populated.
+  float confidence;        ///< 0..1 model posterior. Valid iff has_confidence is true.
```
skottmckay: Should we have `has_confidence` or use FLOAT_MIN for 'not set'?

AI generated the '0..1' range. is that correct or could it differ and we shouldn't document it as exactly?

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -458,6 +458,40 @@ struct ToolResultContent {
...
+};
+
+/// Content returned from a SPEECH_RESULT item (output-only).
+/// `segments` exposes non-owning views over segment items owned by the result.
+struct SpeechResultContent {
+  std::string_view text;
+  std::optional<std::string_view> language;
+  std::optional<int64_t> duration_ms;
+  std::vector<Item> segments;
+};
```
skottmckay, line 493: Should these be simple structs (which forces population of all fields when creating) or classes with methods to access data so we only pay a cost on access?

sdk_v2/js/test/audio-session-streaming.test.ts
skottmckay, line 1: Moved tests into audio-session.test.ts

baijumeswani (dismissed):

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
kunal-vaishnavi: Confidence is a percentage so the 0 to 1 range makes sense when representing as a decimal. I see above that when there are no start and end times, the value set is `FOUNDRY_LOCAL_DURATION_UNSET`. Perhaps we can do the same with `confidence` and remove `has_confidence` to match all three values.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
kunal-vaishnavi: I think the existing structs approach is fine since it mirrors the other structs/class patterns in this file.

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -23,6 +25,54 @@
...
+    std::string text, std::vector<std::unique_ptr<SpeechSegmentItem>> segments) {
+  auto result = std::make_unique<SpeechResultItem>(std::move(text));
+  result->segments = std::move(segments);
+  result->Finalize();
+  return result;
+}
+
+// Initial capacity for the per-token accumulators. Picked empirically: a few seconds of speech
+// (~10s on Whisper, ~5s on Nemotron streaming) produces under 256 tokens, so most short-form
+// transcriptions avoid any reallocation. Longer transcriptions still grow geometrically.
```
kunal-vaishnavi, line 57: What is the reallocation policy?

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -231,14 +287,20 @@ void AudioSession::ProcessStreamingAudio(const AudioItem& format_item, ItemQueue
   auto tokenizer_stream = OgaTokenizerStream::Create(Model().GetOgaTokenizer());
 
   auto streaming_callback = CreateCallbackHandler(request);
-  std::string full_text;
+  std::vector<std::string> token_texts;
+  token_texts.reserve(kInitialTokenCapacity);
+  std::vector<std::unique_ptr<SpeechSegmentItem>> segments;
+  segments.reserve(kInitialTokenCapacity);
+  // Streaming ASR has no text prompt (input is audio), so prompt_tokens stays 0.
```
kunal-vaishnavi, line 294: If we were to use Whisper for streaming ASR in the future (e.g. with a VAD), we could potentially have a text prompt as well (see the `previous text tokens` part [here](https://github.com/openai/whisper/blob/main/approach.png) for reference).

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
skottmckay: Depends on the std::vector implementation. Some do 2x. Some use the golden ratio. Some use 1.5x.

sdk_v2/cpp/src/items/speech_segment_item.h
```diff
@@ -0,0 +1,74 @@
...
+/// the current segment, then exactly one FINAL closes it. Segment identity
+/// is implicit in stream order; there is no segment id. `utterance_start`
+/// tags the first segment of a new utterance.
+///
+/// PARTIAL `text` is the cumulative current hypothesis for the segment, not
+/// a delta.
+///
+/// As an entry of a SpeechResultItem, `kind` is FINAL (or NONE for a single
+/// non-segmented transcript).
+struct SpeechSegmentItem : Item {
```
kunal-vaishnavi, line 37: Every decoded token creates a heap-allocated `SpeechSegmentItem` with its own `Finalize()` + cached C ABI struct. For long transcriptions (thousands of tokens), this could be significant. A pool allocator or a flat struct array could reduce the per-segment allocation overhead, though this is an optimization concern rather than a correctness issue.

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
skottmckay: Is there anything we should add now to support this? It's internal code so we can add later if that's easier.

sdk_v2/cpp/src/items/speech_segment_item.h
skottmckay: Would prefer to see perf data before adding complexity here.

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
nenad1002: ASRs in general can have conditioning prompt (Whipser and Cohere have it), but most of that should probably be abstracted from the user IMO. I cannot think of a scenario where would a user need to care about it, so its internal.

skottmckay: this is only used for returning usage counts, which may not be super meaningful in local inferencing.

kunal-vaishnavi (approved):

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
skottmckay: can add in the future when meaningful

PR #758, opened 2026-06-02
updates webgpu ep version to be decoupled from foundry local version in cpp core

prathikr:


sdk_v2/cpp/src/ep_detection/webgpu_ep_bootstrapper.cc
```diff
@@ -25,82 +28,82 @@ namespace {
...
 constexpr const char* kPackageFileName = "webgpu-ep.zip";
 constexpr const char* kLockFileName = "webgpu-ep.lock";
+constexpr const char* kStagingDirName = "webgpu-ep-staging";
 constexpr const char* kUserAgent = "FoundryLocal";
 constexpr int kMaxInstallAttempts = 5;
 
-// Platform-specific download URL suffix.
+// Manifest URL — always uses prod.
+constexpr const char* kManifestUrl =
+    "https://foundrypackages-ffhrdhbxb7gpdreh.b02.azurefd.net/webgpu_ep_prod.json";
```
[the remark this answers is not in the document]
prathikr: this is the same behavior as previous c# core and was discussed to be safe

sdk_v2/cpp/src/ep_detection/webgpu_ep_bootstrapper.cc
```diff
@@ -118,6 +121,32 @@ bool WebGpuEpBootstrapper::IsRegistered() const {
...
+    if (!std::equal(hash.begin(), hash.end(), expected_hash.begin(), expected_hash.end(),
+                    [](char a, char b) { return std::toupper(a) == std::toupper(b); })) {
+      logger.Log(LogLevel::Warning,
+                 fmt::format("WebGPU EP: hash mismatch for {}: got {}, expected {}",
+                             filename, hash, expected_hash));
+      return false;
+    }
+  }
+
+  return true;
```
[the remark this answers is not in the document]
prathikr: this is ok, we expect it to always be populated and it should correctly fail if the sha is missing

sdk_v2/cpp/src/ep_detection/webgpu_ep_bootstrapper.cc
```diff
@@ -25,82 +28,82 @@ namespace {
...
-  const char* sha256;
-};
+constexpr const char* kRegistrationName = "Foundry.WebGPU";
 
-constexpr ExpectedBinary kExpectedBinaries[] = {
-    {kWebGpuProviderLib, kWebGpuProviderHash},
+/// Parsed manifest entry for a single platform.
+struct ManifestPackageInfo {
+  std::string url;
+  std::unordered_map<std::string, std::string> sha256;  // filename -> hash
```
baijumeswani: for webgpu, we just have a single file `onnxruntime_providers_webgpu.dll` / `libonnxruntime_providers_webgpu.dylib`. We can avoid using an undordered_map here since the mapping only contains a single entry.

sdk_v2/cpp/src/manager.cc
```diff
@@ -247,17 +247,19 @@ Manager::Manager(const Configuration& config)
...
+      auto cuda_ep_dir = cache_dir + "/cuda-ep";
       bootstrappers.push_back(std::make_unique<CudaEpBootstrapper>(std::move(cuda_ep_dir), register_ep));
     }
 
     // WebGPU EP — always available (no hardware detection needed).
     // Skipped in WinML builds because the WinML-aligned ORT (1.23.2) is older
     // than the ORT API version required by the WebGPU EP plugin (>= 24).
 #if !(defined(FOUNDRY_LOCAL_USE_WINML) && FOUNDRY_LOCAL_USE_WINML)
-    auto webgpu_ep_dir = *config_.model_cache_dir + "/webgpu-ep";
+    auto webgpu_ep_dir = cache_dir + "/webgpu-ep";
```
baijumeswani: Let's add a TODO here so that we remember to enable winml 2.0 + webgpu path when @bmehta001 adds the support.

sdk_v2/cpp/src/ep_detection/webgpu_ep_bootstrapper.h
```diff
@@ -37,6 +39,11 @@ class WebGpuEpBootstrapper : public IEpBootstrapper {
                            ILogger& logger) override;
 
  private:
+  /// Verify all expected binaries exist in @p dir with correct SHA-256 hashes.
+  static bool VerifyPackage(const std::filesystem::path& dir,
+                            const std::unordered_map<std::string, std::string>& expected_hashes,
+                            ILogger& logger);
```
baijumeswani: Why static? Can we move this to a common file if the implementation can be shared across eps? Also, why static?

sdk_v2/cpp/src/ep_detection/webgpu_ep_bootstrapper.cc
```diff
@@ -136,61 +165,91 @@ bool WebGpuEpBootstrapper::DownloadAndRegister(bool force,
   attempts_++;
 
   auto ep_dir = std::filesystem::path(ep_dir_);
-  auto lock_path = ep_dir.parent_path() / kLockFileName;
-  auto zip_path = ep_dir.parent_path() / kPackageFileName;
+  auto parent_dir = ep_dir.parent_path();
 
   try {
```
baijumeswani, line 139: In case of errors with download+register, will the code gracefully skip over webgpu and continue with registration of other eps. We should log errors, and report it to users through api without failing mid registration. I think this is already wired up, but we should check if it is indeed working that way.

sdk_v2/cpp/src/ep_detection/ep_utils.h
```diff
@@ -0,0 +1,54 @@
...
+namespace fl {
+
+/// Verify a set of binaries in @p dir all exist and match their expected SHA-256 hashes.
+///
+/// @param dir            Directory containing the extracted EP binaries.
+/// @param expected       List of (filename, expected_sha256_hex) pairs.
+/// @param ep_name        EP name used in warning log messages (e.g. "CUDA EP").
+/// @param logger         Logger for diagnostic output.
+/// @return true if every file exists and its hash matches; false otherwise.
+inline bool VerifyEpPackage(
```
baijumeswani: Does this need to be inlined?

sdk_v2/cpp/src/manager.cc
```diff
@@ -247,19 +247,19 @@ Manager::Manager(const Configuration& config)
 #endif
 
   if (config_.model_cache_dir.has_value()) {
+    auto cache_dir = std::filesystem::path(*config_.model_cache_dir).parent_path().string();
+
     // CUDA EP — only if an NVIDIA GPU is detected
     if (CudaEpBootstrapper::HasNvidiaGpu()) {
-      auto cuda_ep_dir = *config_.model_cache_dir + "/cuda-ep";
+      auto cuda_ep_dir = cache_dir + "/cuda-ep";
```
baijumeswani: Since `cache_dir` is a filesystem path, can we do

```cpp
const auto cache_dir = std::filesystem::path(*config_.model_cache_dir).parent_path();
const auto cuda_ep_dir = cache_dir / "cuda-ep";
```

sdk_v2/cpp/src/manager.cc
baijumeswani: Looks like the check was removed? We can remove the todo then.

baijumeswani (approved):

PR #782, opened 2026-06-05
migrates SDK from Microsoft.ML.OnnxRuntime.Foundry to Microsoft.ML.OnnxRuntime

prathikr:
Eliminates the need for a foundry-specific ORT package and fixes pipeline to read versions from `deps_versions.json`

NOTE: blocked on WebGPU/CUDA being fully plug-in


baijumeswani:
Thinking more about this. We probably need to continue using the .Foundry package at least for now. Let's re-visit this again once we have a clearer plan.

PR #788, opened 2026-06-09
[SDKv2] Migrate WinML EP downloads to WinML 2.x (reg-free runtime)

bmehta001:
## Summary

Migrates WinML EP-download path from `Microsoft.WindowsAppSDK.ML` 1.8.2192 (WinML 1.x, requires the Windows App SDK bootstrap and Win11 24H2+) to `Microsoft.Windows.AI.MachineLearning` 2.1.70 (WinML 2.x, registration-free, Win10 19H1+)

## Highlights

- **Remove WinAppSDK Bootstrap plumbing across all SDKs, as is any code that existed only to support Bootstrapping.** 
- **Unify ORT** to 1.25.1 and OnnxRuntimeGenAI to 0.13.2 for both flavors
- Drop the **WebGPU-EP skip** on WinML builds (both flavors now satisfy ORT API ≥ 24).
- Drop the `IsWindows11_24H2OrLater()` runtime guard in `winml_ep_bootstrapper.cc`, since `LoadLibraryW` is a sufficient probe.
- Gate `find_package(WinMLEpCatalog)` behind `FOUNDRY_LOCAL_USE_WINML` so non-WinML builds don't silently link the catalog DLL.
- CMake: update DLL path `runtimes-framework/<rid>/native/` → `runtimes/<rid>/native/` and bump default `WINML_EP_CATALOG_VERSION` to 2.1.70.
- Pipelines: bump `cppWinmlVersion` to 2.1.70, drop `cppOrtVersionWinml` and the entire `Microsoft.WindowsAppSDK.Foundation` resolution + `Bootstrap.dll` staging in `steps-prefetch-nuget.yml` / `steps-build-windows.yml`.
- NuGet pack: replace `Microsoft.WindowsAppRuntime.Bootstrap.dll` with `Microsoft.Windows.AI.MachineLearning.dll` in `OPTIONAL_SIBLINGS`.
- Bump gtest `DISCOVERY_TIMEOUT` to 60 s for `foundry_local_tests` and `cache_only_tests` — the WinML build's delay-loaded DLL resolution + static initializers can exceed the 5 s default.
- **C# WinML SKU TFM relaxation**: drops the `net9.0-windows10.0.26100.0` lock and targets `net8.0;net9.0` instead. The WinML 2.x OS floor (Win10 19H1+) is enforced at runtime by `winml_ep_bootstrapper.cc` and by the EP catalog NuGet, not by the C# csproj, so consumers no longer need the Win11 24H2 SDK to build against `Microsoft.AI.Foundry.Local.WinML`. `netstandard2.0` is intentionally *not* added to the WinML SKU — the standard SKU keeps it (and the `net462` test lane) for .NET Framework consumers; the WinML SKU's audience is Win11-era + modern .NET, so the `net462` test lane is gated off for the WinML test matrix.

## WinML 2.x EP semantics (informational)

- **Per-EP downloads happen one at a time today.** 2.x exposes `WinMLEpEnsureReadyAsync(WinMLAsyncBlock*)`, but the OS-side MSIX/BITS work is largely serialized anyway. Left as a future optimization.
- **Catalog (`WinMLEpCatalogHandle`) is only created during EnsureReady at the beginning.** The 2.x C API has no Refresh/Invalidate, so if new EPs are released mid-process, they will only be visible after process restart.
- **Foundry's `AzureModelCatalog` cache** is correctly invalidated in `Manager::DownloadAndRegisterEps` (manager.cc:547) after a successful register, so model-filter results reflect the new EP set immediately.

Maybe the extra error-checking in @sdk_v2/cpp/src/ep_detection/winml_ep_bootstrapper.cc is unnecessary, but I have added it for diagnostics.


sdk_v2/cpp/src/ep_detection/winml_ep_bootstrapper.cc
```diff
@@ -19,50 +19,37 @@
...
-  if (osvi.dwMajorVersion == 10 && osvi.dwMinorVersion == 0) {
-    return osvi.dwBuildNumber >= 26100;
+  RTL_OSVERSIONINFOW info{};
+  info.dwOSVersionInfoSize = sizeof(info);
+  if (rtl_get_version(&info) != 0) {
+    return 0;
   }
-
-  return false;
+  return info.dwBuildNumber;
```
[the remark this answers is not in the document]
bmehta001: The original code did compile cleanly because `RTL_OSVERSIONINFOW` is a typedef alias of `OSVERSIONINFOW` in `<winnt.h>` (which `<windows.h>` already includes), but switching to the canonical public name `OSVERSIONINFOW` removes the ambiguity for readers and static analysers.

sdk_v2/cpp/cmake/FindWinMLEpCatalog.cmake
```diff
@@ -22,29 +27,40 @@ if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
...
+set(_WINML_EP_CATALOG_MIN_VERSION "2.1.70")
 
-# WINML_EP_CATALOG_VERSION may be set explicitly; otherwise pick the minimum
-# known-good version. We deliberately do NOT inherit WINML_SDK_VERSION here:
-# the WinML SDK and the EP catalog package have independent compatibility
-# requirements (the EP catalog ships only in newer WindowsAppSDK.ML packages,
-# and our build no longer uses the WinML-bundled ORT regardless).
 if(NOT WINML_EP_CATALOG_VERSION)
     set(WINML_EP_CATALOG_VERSION "${_WINML_EP_CATALOG_MIN_VERSION}")
 endif()
```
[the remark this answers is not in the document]
bmehta001: Updated the PR description to `2.1.70`

.pipelines/v2/templates/steps-prefetch-nuget.yml
```diff
@@ -74,34 +74,15 @@ steps:
...
-          Write-Host "  -> $($foundationNupkg.FullName) ($($foundationNupkg.Length) bytes)"
-          $defines += "WINML_EP_CATALOG_FETCH_URL=$($mlNupkg.FullName)"
-          $defines += "WINAPPSDK_FOUNDATION_FETCH_URL=$($foundationNupkg.FullName)"
+          # WinML 2.x (Microsoft.Windows.AI.MachineLearning) is reg-free — a single self-contained
+          # native package with no transitive Windows App SDK Foundation dependency to resolve.
+          $winmlUrl = "$feed/Microsoft.Windows.AI.MachineLearning/${{ parameters.winmlVersion }}"
+          $winmlOut = "$cacheDir/winml.nupkg"
+          Write-Host "Downloading Microsoft.Windows.AI.MachineLearning ${{ parameters.winmlVersion }}"
+          Invoke-WebRequest -Uri $winmlUrl -OutFile $winmlOut -UseBasicParsing
+          if (-not (Test-Path $winmlOut)) { throw "WinML download failed" }
```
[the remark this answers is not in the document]
bmehta001: This claim isn't accurate — `-UseBasicParsing` was *not* removed in PowerShell 6+. Per the [official `Invoke-WebRequest` docs](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest), it's retained for backward compatibility and is silently accepted as a no-op in PS 6+. The script will execute fine. Leaving the flag as-is.

sdk_v2/cpp/src/ep_detection/winml_ep_bootstrapper.cc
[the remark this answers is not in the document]
bmehta001: Already addressed in a prior commit on this PR — QueryWindowsBuild() now uses OSVERSIONINFOW from <windows.h> (the union/payload-compatible alias of RTL_OSVERSIONINFOW), so no <winternl.h> is needed. Calling RtlGetVersion(OSVERSIONINFOW*) is safe because the two structs are layout-compatible; this is the same idiom Chromium and Wine use to avoid pulling in <winternl.h>'s macro pollution.

sdk_v2/cpp/cmake/FindWinMLEpCatalog.cmake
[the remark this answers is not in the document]
bmehta001: No discrepancy — the PR description consistently says 2.1.70 (no 2.1.6 reference anywhere). Both FindWinMLEpCatalog.cmake and deps_versions.json agree on 2.1.70, and the CMake module now reads the version from deps_versions.json as the single source of truth (see commit 5f107d1a).

sdk_v2/cpp/src/ep_detection/winml_ep_bootstrapper.cc
```diff
@@ -1,75 +1,132 @@
...
+          reinterpret_cast<LPCWSTR>(&LoadWinMLDllFromOwnDirectory),
+          &own_module) ||
+      own_module == nullptr) {
+    return nullptr;
   }
 
-  // Windows 11 24H2 = build 26100+
-  if (osvi.dwMajorVersion > 10) {
-    return true;
+  wchar_t module_path[MAX_PATH];
```
skottmckay: Does this length support potential values when the user has enabled long paths?

sdk_v2/cs/test/FoundryLocal.Tests/Microsoft.AI.Foundry.Local.Tests.csproj
```diff
@@ -1,19 +1,23 @@
...
+
+    <!-- Always test on the .NET (Core) LTS pair. The Windows default SKU also targets
+         net462 to exercise the netstandard2.0 surface of the SDK end-to-end: the
+         netstandard-only DllLoader path, the Utf8 unsafe-pointer fallback, the
+         System.Memory / Bcl.AsyncInterfaces / System.Threading.Channels polyfills,
+         and the vtable async/await pipeline all run on .NET Framework. CI and local
+         scripts pass an explicit framework filter so the .NET (Core) test suite
+         only runs once. -->
+    <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(IsWindows)' == 'true' And '$(UseWinML)' != 'true'">net462;net8.0;net9.0</TargetFrameworks>
```
skottmckay: Is there benefit to testing .net8 and .net9 or would picking just one be sufficient?

skottmckay (dismissed):

.pipelines/foundry-local-packaging.yml
```diff
@@ -54,17 +54,13 @@ parameters:
 variables:
 - group: FoundryLocal-ESRP-Signing
 # C++ SDK (sdk_v2/cpp) native dependency versions. Must match cmake defaults
-# in sdk_v2/deps_versions.json and sdk_v2/deps_versions_winml.json.
+# in sdk_v2/deps_versions.json.
 - name: cppOrtVersion
   value: '1.25.1'
```
baijumeswani: Does this need to be 1.26.0?

bmehta001: Fixed

sdk_v2/cs/test/FoundryLocal.Tests/Microsoft.AI.Foundry.Local.Tests.csproj
bmehta001: Just .NET8.0 should probably be fine, since .NET9.0 should be backwards-compatible. I will simplify

sdk_v2/cpp/src/ep_detection/winml_ep_bootstrapper.cc
bmehta001: You're right - if the path is too long and the LoadLibraryW fallback fails (if the DLL will be on the system PATH or in the app directory, which I think should typically be the case but may not be), then this would fail. I will double the buffer size until the path fits.

skottmckay (dismissed):

baijumeswani (approved):

PR #789, opened 2026-06-09
Enable live catalog with region-aware fallback engine

baijumeswani:
Replaces the static, file-based model catalog with a live Azure Foundry catalog client, and makes both catalog queries and model downloads region-aware. The client now detects the caller's Azure region from cluster headers, queries that region's catalog endpoint, and transparently retries against nearby regions when an endpoint is unhealthy. Downloads resolve their SAS URI from the same region that served the model's catalog entry.

This ports three behaviors from the C# core into the C++ SDK:
1. Fetch the live catalog instead of reading a baked-in static snapshot.
2. Detect the region from `azureml-served-by-cluster` headers (instead of always using `eastus`).
3. Region-aware catalog query and model download.

## How it works

```
                        ┌──────────────────────────────────────────┐
                        │  AzureModelCatalog.FetchModels()         │
                        └───────────────────┬──────────────────────┘
                                            │
                                            ▼
                        ┌──────────────────────────────────────────┐
                        │  AzureCatalogClient                      │
                        │                                          │
                        │  1. Detect region                        │
                        │     POST probe ─► read                   │
                        │     "azureml-served-by-cluster" header   │
                        │     vienna-<region>-NN ─► region         │
                        │     (explicit config wins; else eastus)  │
                        └───────────────────┬──────────────────────┘
                                            │ for each device filter set
                                            ▼
                        ┌──────────────────────────────────────────┐
                        │  RegionFallback.Execute(startRegion)     │
                        │                                          │
                        │   chain = [start]                        │
                        │         + proximal[start]                │
                        │         + one random public region       │
                        └───────────────────┬──────────────────────┘
                                            │
                       ┌────────────────────┼─────────────────────┐
                       ▼                     ▼                     ▼
                  try region A ─►       try region B ─►       try region C
                   2xx? ✓ done          retryable(0/408/        ...exhausted
                   permanent(404)? ✓     429/5xx)? ─► next       ─► fail set
                   stop                                          (others continue)
                       │
                       │ page 1 pins the serving region;
                       │ later pages stay on it (tokens are region-specific)
                       ▼
            ┌──────────────────────────────────────────┐
            │  Models stamped with detected_region,    │
            │  committed atomically per filter set     │
            └───────────────────┬──────────────────────┘
                                │  (download)
                                ▼
            ┌──────────────────────────────────────────┐
            │  DownloadManager.DownloadModel(info)     │
            │                                          │
            │  region = explicit config                │
            │         → info.detected_region           │
            │         → eastus                         │
            └───────────────────┬──────────────────────┘
                                ▼
            ┌───────────────────────────────────────────┐
            │  ModelRegistryClient.ResolveModelContainer│
            │  (same RegionFallback chain) ─► SAS URI   │
            └───────────────────┬───────────────────────┘
                                ▼
                          Blob download
```

- `catalog_region`: `""`/`"auto"` → detect from headers; any other value → explicit override.
- `DisableRegionFallback`: disables cross-region retries (single-region behavior).


sdk_v2/cpp/src/manager.cc
```diff
@@ -278,11 +278,21 @@ Manager::Manager(const Configuration& config)
     }
   }
 
+  // Read whether cross-region fallback should be disabled (default: enabled).
+  // Accepts case-insensitive true/1/yes.
+  bool disable_region_fallback = false;
+  const auto fallback_it = config_.additional_options.find("DisableRegionFallback");
+  if (fallback_it != config_.additional_options.cend()) {
+    const auto value = to_lower(fallback_it->second);
+    disable_region_fallback = (value == "true" || value == "1" || value == "yes");
```
bmehta001: Not necessary to change in this PR, but no other parameter that gets parsed from the config accepts "yes". I think it's a good idea to keep here, but maybe we should standardize what truthy-values get accepted with a helper function somewhere

sdk_v2/cpp/src/http/http_client.cc
```diff
@@ -116,6 +123,39 @@ std::string HttpGet(const std::string& url, const std::string& user_agent, bool
   return HttpRequest(Azure::Core::Http::HttpMethod::Get, url, "", user_agent, close_connection);
 }
 
+HttpResponse HttpPostWithResponse(const std::string& url,
+                                  const std::string& json_body,
+                                  const std::string& user_agent,
+                                  std::chrono::milliseconds timeout,
+                                  bool close_connection) {
```
bmehta001: Nit: Maybe consider changing the order of `close_connection` and `timeout` here and in `HttpGetWithResponse` to align more with HttpRequestRaw

Or, maybe there should be an HttpRequestOptions struct that gets passed in to store user_agent, timeout, and close_connection? Maybe it's fine as is if we don't anticipate any more arguments being added in the future.

sdk_v2/cpp/src/http/http_client.h
```diff
@@ -6,11 +6,42 @@
...
+HttpResponse HttpGetWithResponse(const std::string& url,
+                                 const std::string& user_agent = "",
+                                 std::chrono::milliseconds timeout = std::chrono::seconds(30),
+                                 bool close_connection = false);
+
+/// Human-readable, length-bounded description of a failed response for error messages.
+/// `status == 0` becomes "transport failure"; otherwise "HTTP <status>". When the body is
+/// non-empty it is appended (truncated to `max_body_chars`) so server-side or transport
+/// diagnostics are preserved without bloating logs.
+std::string DescribeFailure(const HttpResponse& response, std::size_t max_body_chars = 512);
```
bmehta001, line 44: Makes sense for consumers; I will need to log ErrorEvent in here when telemetry is added instead of logging later and missing out on some information

sdk_v2/cpp/src/download/download_manager.cc
```diff
@@ -20,6 +22,7 @@ namespace {
 const char* kDownloadSignalFileName = "download.tmp";
 const char* kGenAIConfigFileName = "genai_config.json";
 const char* kInferenceModelFileName = "inference_model.json";
+const char* kDefaultRegistryRegion = "eastus";
```
bmehta001: Should we potentially use another region as the default?

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -0,0 +1,411 @@
...
+constexpr int kPageSize = 50;
+
+// Region detection probe.
+constexpr const char* kRegionProbeUrl = "https://api.catalog.azureml.ms/asset-gallery/v1.0/models";
+constexpr const char* kRegionProbeBody = R"({"filters":[],"pageSize":1})";
+constexpr const char* kServedByClusterHeader = "azureml-served-by-cluster";
+constexpr const char* kDefaultRegion = "eastus";
+
+// The catalog and registry gateways reject requests without this User-Agent (HTTP 400).
+constexpr const char* kUserAgent = "AzureAiStudio";
```
skottmckay, line 32: Is now a good time to try a different value here to differentiate traffic?

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -0,0 +1,411 @@
...
+
+/// Strip leading/trailing whitespace.
+std::string Trim(const std::string& s) {
```
skottmckay: Should we put all string utils in a single file under `src/util`? Could move `to_lower` from `src/utils.h` and maybe MakeString from `src/util/make_string.h` into that as well.

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -0,0 +1,411 @@
...
+        // models are committed atomically below, a later page failure cannot leak a partial filter-set result.
+        logger_.Log(LogLevel::Warning,
+                    "catalog: filter set failed (" + http::DescribeFailure(response) + "); skipping this filter set.");
+        return std::nullopt;
+      }
+
+      const std::string url = regional && pinned_region
+                                  ? BuildRegionalUrl(url_prefix_, url_suffix_, *pinned_region)
+                                  : BuildRequestUrl(base_url_, regional, region_, url_prefix_, url_suffix_);
+      FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL,
```
skottmckay: Should we have a different status for this? I would typically use 'internal' for some implementation logic failure (i.e. our code is broken or did something unexpected), and I wonder if this failure is more likely to be network related.

sdk_v2/cpp/src/download/download_manager.cc
```diff
@@ -148,13 +151,33 @@ std::string SanitizeForPathSegment(std::string_view name) {
...
+  return kDefaultRegistryRegion;
+}
+
 }  // anonymous namespace
 
 DownloadManager::DownloadManager(std::string cache_directory, std::string catalog_region, int max_concurrency,
-                                 ILogger& logger)
+                                 ILogger& logger, bool disable_region_fallback)
     : cache_directory_(std::move(cache_directory)),
+      config_region_(NormalizeConfiguredRegion(catalog_region)),
```
skottmckay, line 177: nit: If we're not `std::move`'ing `catalog_region` now should we take it by reference or maybe as a std::string_view given we're calling `to_lower` on it?

sdk_v2/cpp/test/sdk_api/shared_test_env.h
skottmckay, line 1: Should we have 'real' tests in test/sdk_api for any of the new changes?

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
baijumeswani: Unfortunately, no. If we change this to anything other than AzureAiStudio, we receive HTTP 400 response. The server rejects our requests. This is probably by design since they don't requests from random User-Agents.

Will work with the catalog team and update this in a separate commit.

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
baijumeswani: Added a string_utils.h and moved the string utility functions there.

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
baijumeswani: Added a new error code FOUNDRY_LOCAL_ERROR_NETWORK.

sdk_v2/cpp/test/sdk_api/catalog_live_test.cc
```diff
@@ -0,0 +1,157 @@
...
+  for (const auto& model : list) {
+    auto info = model->GetInfo();
+    std::cout << "  - alias=" << info.Alias() << "  id=" << info.Id() << "  name=" << info.Name() << "\n";
+    EXPECT_FALSE(info.Id().empty()) << "every catalog model must expose a non-empty id";
+  }
+  std::cout << "=== end catalog ===\n";
+}
+
+// Downloads a real model end-to-end through the public API. Disabled in CI (network + large I/O).
+TEST(CatalogLiveTest, DownloadRealModel) {
```
skottmckay: nit: Should this also use the DISABLED_ prefix so it doesn't run every time a developer runs locally?

skottmckay (dismissed):

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -0,0 +1,400 @@
...
+namespace {
+
+constexpr const char* kEntitiesPath = "/entities/crossRegion";
+constexpr int kPageSize = 50;
+
+// Region detection probe.
+constexpr const char* kRegionProbeUrl = "https://api.catalog.azureml.ms/asset-gallery/v1.0/models";
+constexpr const char* kRegionProbeBody = R"({"filters":[],"pageSize":1})";
+constexpr const char* kServedByClusterHeader = "azureml-served-by-cluster";
+constexpr const char* kDefaultRegion = "eastus";
```
bmehta001: Nit: should this be centralus as well?

bmehta001: There are also a few other small comments (some in files not even covered by the PR) that mention eastus, like `sdk_v2/cs/src/Configuration.cs`

sdk_v2/cpp/src/manager.cc
```diff
@@ -55,10 +53,20 @@ bool IsGenAIVerboseLoggingEnabled() {
     return false;
   }
 
-  std::string lowered = to_lower(*env);
+  std::string lowered = ToLower(*env);
```
bmehta001: Nit: Should this just use `IsTruthyConfigValue(*env)`?

sdk_v2/cpp/src/http/http_client.h
bmehta001: Nit: FOUNDRY_LOCAL_ERROR_INTERNAL -> FOUNDRY_LOCAL_ERROR_NETWORK

sdk_v2/cpp/src/util/region_fallback.cc
```diff
@@ -0,0 +1,216 @@
...
+      failure_summary += ", ";
+    }
+
+    failure_summary += region + "(" + DescribeStatus(response.status) + ")";
+    logger_.Log(LogLevel::Warning,
+                "RegionFallback: region '" + region + "' unhealthy (" +
+                    DescribeStatus(response.status) + "); trying next candidate.");
+  }
+
+  FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL,
```
bmehta001: Should this be FOUNDRY_LOCAL_ERROR_NETWORK as well?

bmehta001 (dismissed):

bmehta001 (approved):

PR #793, opened 2026-06-09
[SDKv2] resumable downloads (cross-process lock + per-chunk state + linked cancellation)

bmehta001:
## Resumable downloads (C++ port)

Ports the resumable-download flow from neutron-server's `AzureExtensions.cs` / `BlobDownloadState.cs` / `CrossProcessFileLock.cs` to the C++ SDK in `sdk_v2/cpp`. Two commits, one per increment.

### Increment 1 — cross-process lock + skip-existing (commit `45c99d5d`)

- `CrossProcessFileLock`: RAII, `<dir>/.download.lock` exclusive lock. Win: `CreateFileW FILE_SHARE_NONE FILE_FLAG_DELETE_ON_CLOSE`; POSIX: `flock LOCK_EX|LOCK_NB`. PIMPL (`struct State` defined only in the .cc) keeps platform headers out of the public include.
- `WaitForLockForDirectory`: polls every 1.25 s with a 100 ms cancellation slice and a 3 h ceiling, driven by a `std::function<bool()>` predicate so cancellation can also serve as a heartbeat (the progress callback returning non-zero).
- `DownloadManager::DownloadModel`: takes the lock right after `create_directories`, re-checks the cache once acquired, then proceeds with the existing download flow. Stores `ILogger&` so the lock and resume layers can log uniformly.
- `DownloadBlobsToDirectory`: blobs already present at the expected size are skipped; their bytes are credited to the initial progress callback so the percentage stays accurate on resume. Emits 100% immediately if everything is cached.
- 9 `CrossProcessFileLock` tests + 4 new skip-existing tests in `download_test.cc`.

### Increment 2 — per-chunk resume + linked cancellation (commit `b6d98db5`)

- `BlobDownloadState` (new): compact binary `<file>.dlstate` sidecar. ~45-byte LE header (magic `FLDS` + version + sizes + counters) followed by the truncated bitmap suffix. The prefix of fully-completed chunks is implicit — `SaveState` advances `bitmap_byte_aligned_start` past every all-1s word so the file stays proportional to the unfinished tail. `LoadState` rejects on magic, version, or layout mismatch and restarts the download in that case. Atomic save via tmp + rename, with remove-then-rename fallback.
- `AzureBlobDownloader::DownloadBlob` rework: protected virtuals `GetBlobSize` and `DownloadChunkToBuffer` (against an opaque `ChunkContext`) form a test seam. Worker pool uses an atomic queue index over the pending-chunks list; workers claim, fetch, write to file (under a single `file_mutex`), mark complete, and periodically `SaveState` (every `max(10, num_chunks/50)` chunks). Pre-allocation is skipped if the file already matches blob size, so resume doesn't discard valid bytes.
- Linked cancellation: a single shared `Azure::Core::Context` + a single `std::atomic<bool> internal_cancel` per call. The first chunk failure (or external cancel signal) flips the flag and calls `ctx.Cancel()`. Other in-flight chunks see either signal and exit fast — a chunk failing while 9 others are mid-flight drains in ~300 ms in the new `ChunkFailureCancelsInFlightPeersFast` test.
- `IsDownloadNeeded` now also returns true if a `.dlstate` sidecar exists (the data file may be pre-allocated with holes).
- Sidecar is persisted on any failure and deleted on full success.

### Open decisions (resolved)

| Question | Resolution |
|---|---|
| `BlobDownloadState` design | Port the C# truncated-bitmap shape directly. |
| Sidecar filename | `.dlstate` (cross-language compat with C# `.download.state` was explicitly out of scope). |
| `OnDownloadComplete` telemetry | Deferred per spec; C++ already signals via 100% progress callback. |
| Mockability | Both: `IBlobDownloader` mock for orchestrator tests + `AzureBlobDownloader` subclass via protected virtuals for chunk-level injection. |

### Tests

- 9 new `CrossProcessFileLockTest` cases (Inc 1)
- 4 new skip-existing cases in `download_test.cc` (Inc 1)
- 15 new `BlobDownloadStateTest` cases (Inc 2): create/mark/save/load/delete, gap enumeration, partial final chunk math, byte-aligned-start advancement, rejection of magic / size / total_chunks mismatches.
- 5 new `AzureBlobDownloaderResumeTest` cases (Inc 2): resume from sidecar, fresh download, sidecar persists on chunk failure, stale sidecar cleaned up for empty blobs, cancel-cascade timing.

Targeted `BlobDownloader|DownloadManager|CrossProcessFileLock|AzureBlobDownloader` suite: **60/60 in 13 s**. Full gtest run is 789 passed / 54 skipped; the 3 pre-existing `ModelLoadManagerUnloadTest` failures are environmental (missing local test model file `qwen2.5-0.5b-instruct-generic-cpu-4`), not caused by this PR.

### Out of scope

Region-based download (Increment 3 in the spec) — tracked separately.

### Notes for reviewers

- C ABI is unchanged; this PR is purely internal to `sdk_v2/cpp`.
- Reference implementation in C# lives in `neutron-server/src/Downloader/` (private repo); see Increment 2's bitmap-truncation logic in `BlobDownloadState.cs:237-281` for the closest analog.


sdk_v2/cpp/src/download/blob_downloader.cc
```diff
@@ -20,10 +25,30 @@
...
+/// Per-blob shared state passed to the protected virtuals. Cancellation is
+/// observed through `azure_ctx`: the orchestrator calls `Cancel()` on it after
+/// the first chunk failure or on external cancellation, which interrupts every
+/// in-flight chunk read.
+struct AzureBlobDownloader::ChunkContext {
+  Azure::Storage::Blobs::BlobClient* blob_client;
+  Azure::Core::Context* azure_ctx;
+};
+
+AzureBlobDownloader::AzureBlobDownloader(ILogger* logger) : logger_(logger) {}
```
skottmckay: nit: logger should be required so pass as a reference. can probably store as a reference (I assume this class doesn't need to be copyable which would require a pointer for the member) as well. #Closed

sdk_v2/cpp/src/download/blob_downloader.cc
```diff
@@ -20,10 +25,30 @@
...
 // AzureBlobDownloader — real Azure Storage SDK implementation
 // ========================================================================
 
+/// Per-blob shared state passed to the protected virtuals. Cancellation is
+/// observed through `azure_ctx`: the orchestrator calls `Cancel()` on it after
+/// the first chunk failure or on external cancellation, which interrupts every
+/// in-flight chunk read.
+struct AzureBlobDownloader::ChunkContext {
+  Azure::Storage::Blobs::BlobClient* blob_client;
+  Azure::Core::Context* azure_ctx;
```
skottmckay: Use `const T&` for both these. 

In general, use `const` and `&` wherever possible as they tighten the usage.  #Closed

sdk_v2/cpp/src/download/blob_downloader.cc
```diff
@@ -45,6 +70,61 @@ std::vector<BlobItemInfo> AzureBlobDownloader::ListBlobs(const std::string& sas_
...
+  auto& body_stream = *result.Value.BodyStream;
+
+  if (scratch.size() < kStreamingBufferBytes) {
+    scratch.resize(kStreamingBufferBytes);
+  }
+
+  int64_t remaining = size;
+  while (remaining > 0) {
+    size_t to_read =
+        static_cast<size_t>(std::min<int64_t>(remaining, static_cast<int64_t>(scratch.size())));
```
skottmckay: nit: AI needs to frequently be reminded it can use 120 chars. this could be a single line. #Closed

sdk_v2/cpp/src/download/blob_downloader.cc
skottmckay, line 146: Unrelated to this review, but out of interest should we be doing this? Is it just going to make a bad situation worse?

sdk_v2/cpp/src/download/blob_downloader.cc
```diff
@@ -65,155 +145,226 @@ void AzureBlobDownloader::DownloadBlob(const std::string& sas_uri,
...
+      if (data_ec || data_size != static_cast<uintmax_t>(blob_size)) {
+        if (logger_) {
+          logger_->Log(LogLevel::Information,
+                       "Resume sidecar for '" + local_path +
+                           "' has no matching full-size data file; starting fresh");
+        }
+        state.reset();
+      }
+    }
+    if (!state) {
```
skottmckay, line 193: nit: whitespace between separate blocks is more readable. might need to tell the agent to read .github\instructions\cpp-formatting.instructions.md. if I see it doing this I would typically say "read this and apply to the changes on this branch". #Closed

sdk_v2/cpp/test/sdk_api/download_test.cc
```diff
@@ -131,3 +131,69 @@ TEST_F(DISABLED_DownloadFixture, DownloadAlreadyCachedModelIsNoOp) {
...
+  // already on disk (well above 0) rather than restarting from scratch.
+  std::vector<float> resume_progress;
+  target->Download([&resume_progress](float pct) -> int {
+    resume_progress.push_back(pct);
+    return 0;
+  });
+
+  ASSERT_FALSE(resume_progress.empty());
+  EXPECT_GT(resume_progress.front(), 0.0f)
+      << "resume should start from the partial progress already on disk, not 0%";
```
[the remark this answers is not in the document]
bmehta001: Good catch — fixed in 47214852. DownloadModel emits the 0% heartbeat first, so I now skip leading zeros and assert the first *non-zero* sample (the on-disk reflection from DownloadBlobsToDirectory) lands well above a fresh-download first-chunk fraction, with a lenient, documented threshold to absorb the ~16 MB sidecar-save granularity.

sdk_v2/cpp/src/download/blob_download_state.cc
```diff
@@ -0,0 +1,396 @@
...
+
+// Serialize a scalar field in host byte order. Every target we build for
+// (x64 / arm64) is little-endian, so the on-disk layout is little-endian in
+// practice.
+template <typename T>
+void WriteNative(std::ostream& out, T value) {
+  static_assert(std::is_trivially_copyable_v<T>);
+  unsigned char buf[sizeof(T)];
+  std::memcpy(buf, &value, sizeof(T));
+  out.write(reinterpret_cast<const char*>(buf), sizeof(T));
```
skottmckay: why do we need to go via a temporary buffer and use a template? that's adding quite a bit of code and it's not clear to me if there's a need for that complexity and binary size cost given the file is only written and read locally. #Closed

sdk_v2/cpp/src/download/blob_download_state.cc
```diff
@@ -0,0 +1,396 @@
...
+}
+
+void BlobDownloadState::DeleteState(const std::filesystem::path& local_file_path, ILogger* logger) {
+  auto state_path = GetStateFilePath(local_file_path);
+  std::error_code ec;
+  std::filesystem::remove(state_path, ec);
+  if (ec && logger) {
+    logger->Log(LogLevel::Warning,
+                "Failed to delete download state file: " + state_path.string() + " (" +
+                    ec.message() + ")");
```
skottmckay: Should some of these be Error level? If the user is going to have to potentially manually address an issue a Warning may be too low.  #Closed

sdk_v2/cpp/src/download/blob_downloader.cc
```diff
@@ -65,155 +145,226 @@ void AzureBlobDownloader::DownloadBlob(const std::string& sas_uri,
...
+    // observer that watches the data file sees a fully-closed handle.
+    writer.Close();
+
+    if (first_error || (cancelled && cancelled->load(std::memory_order_relaxed))) {
+      // Persist what we have so the next attempt resumes from here.
+      {
+        std::lock_guard<std::mutex> lock(state->mutex());
+        state->SaveState(logger_);
+      }
+      if (cancelled && cancelled->load(std::memory_order_relaxed)) {
```
skottmckay: do we need to do a `load` here?  #Closed

sdk_v2/cpp/src/download/blob_downloader.cc
```diff
@@ -304,25 +483,62 @@ void DownloadBlobsToDirectory(IBlobDownloader& downloader,
...
+      options.progress(100.0f);
+    }
+    return;
+  }
+
   if (options.progress) {
-    int result = options.progress(0.0f);
+    float initial_percent = total_size > 0
+                                ? static_cast<float>(skipped_bytes) /
+                                      static_cast<float>(total_size) * 100.0f
```
skottmckay: nit: could be single line #Closed

sdk_v2/cpp/src/download/cross_process_file_lock.h
```diff
@@ -0,0 +1,62 @@
...
+};
+
+/// Returning true aborts WaitForDirectoryLock with FOUNDRY_LOCAL_ERROR_OPERATION_CANCELLED.
+using CancellationPredicate = std::function<bool()>;
+
+/// Polls TryAcquireForDirectory until the lock is acquired, `is_cancelled()`
+/// returns true, or `timeout` elapses.
+/// Throws FOUNDRY_LOCAL_ERROR_OPERATION_CANCELLED on cancellation, or
+/// FOUNDRY_LOCAL_ERROR_INTERNAL on timeout.
+std::unique_ptr<CrossProcessFileLock> WaitForDirectoryLock(
```
skottmckay: nit: make static member of class #Closed

sdk_v2/cpp/src/download/blob_downloader.cc
```diff
@@ -65,155 +145,226 @@ void AzureBlobDownloader::DownloadBlob(const std::string& sas_uri,
...
+          return;
+        }
 
-    // Mutex protects concurrent writes to different offsets in the same file.
-    // Each chunk opens the file, seeks, and writes — the mutex prevents interleaved I/O.
-    std::mutex file_mutex;
+        int64_t new_total = bytes_completed.fetch_add(size, std::memory_order_relaxed) + size;
+        if (bytes_written_cb) {
+          bytes_written_cb(new_total);
+        }
```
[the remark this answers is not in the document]
bmehta001: Done in e4f9dfca — moved the byte-accounting + `bytes_written_cb` call inside the `try` so a user-cancel throw routes through the same catch that runs `azure_ctx.Cancel()`, matching the chunk-failure path. Added `UserCancelDrainsInFlightPeersFast`, which parks every peer inside its chunk before the cancel fires so only the Azure-context cascade can drain them — verified ~120 ms with the change vs ~7.3 s without (i.e. one full chunk per peer), so the test genuinely guards the fix.

sdk_v2/cpp/src/download/file_writer.cc
```diff
@@ -0,0 +1,164 @@
...
+  if (f.fail()) {
+    FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL,
+             "failed to pre-allocate blob file: " + path.string() +
+                 " (size=" + std::to_string(expected_size) + ")");
+  }
+}
+
+}  // namespace
+
+#ifdef _WIN32
```
skottmckay, line 72: nit: would be nice to structure it so that platform specific code is in src/platform like we do with ORT as that makes it easy to find such differences. could be follow up PR. #Closed

sdk_v2/cpp/src/download/blob_downloader.cc
```diff
@@ -65,155 +145,231 @@ void AzureBlobDownloader::DownloadBlob(const std::string& sas_uri,
...
+
+          // Account for this chunk and fire the progress callback within the same
+          // try as the download: on user cancellation bytes_written_cb throws, and
+          // the catch below runs azure_ctx.Cancel() so peers blocked mid-chunk are
+          // interrupted immediately rather than only noticing the cancel flag when
+          // they finish their current chunk.
+          int64_t new_total = bytes_completed.fetch_add(size, std::memory_order_relaxed) + size;
+          if (bytes_written_cb) {
+            bytes_written_cb(new_total);
+          }
```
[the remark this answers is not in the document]
bmehta001: Fixed in fdfa2674 — you're right, this was a regression from this PR's per-chunk rework (8225ce78). Restored the pre-rework behavior of reporting `bytes_completed.load()` (the global running total, always non-decreasing) after the add, instead of each worker's pre-add snapshot. Kept inside the `try` so the user-cancel fast-drain fix is unaffected. Verified the pre-rework code did exactly `bytes_completed += total_read; bytes_written_cb(bytes_completed.load());`.

sdk_v2/cpp/src/download/blob_download_state.cc
bmehta001: My mistake - I had a way more unnecessarily complicated process earlier, and though I simplified most of it, this was left over. Thanks.

sdk_v2/cpp/src/download/blob_downloader.cc
bmehta001: Good point - I will discuss with Baiju and remove it in another PR.

sdk_v2/cpp/src/download/file_writer.cc
bmehta001: Ok, will do in follow-up PR

sdk_v2/cpp/src/download/blob_download_state.cc
bmehta001: Done

sdk_v2/cpp/src/download/file_writer.cc
```diff
@@ -0,0 +1,164 @@
...
+    }
+    offset += static_cast<int64_t>(written);
+    data += written;
+    len -= written;
+  }
+}
+
+void FileWriter::Close() {
+  if (handle_ != nullptr) {
+    ::CloseHandle(handle_);
```
skottmckay: Should we be checking the return value of this? Can't do much except log the failure, but might be good to know if it does fail.

skottmckay (dismissed):

sdk_v2/cpp/src/download/file_writer.cc
bmehta001: Fixed

sdk_v2/cpp/test/internal_api/download_test.cc
```diff
@@ -40,14 +47,27 @@ using namespace fl;
 
 namespace {
 
-/// Create a temporary directory for test isolation.
+/// Create a temporary directory for test isolation. The name comes from MakeUniqueTempPath, so it is
+/// unique both across the separate processes CTest launches per test and across multiple TempDirs
+/// within one test. create_directory must succeed — the directory must not already exist — so a
+/// residual collision (e.g. a directory leaked by an earlier process that reused this pid) advances
+/// to the next name and retries instead of silently sharing an existing directory.
 class TempDir {
```
skottmckay, line 55: nit: Multiple tests have a TempDir class and we also have a TempPath. can we consolidate the temp dir/path logic into a shared helper in test/utils?

skottmckay (approved):

sdk_v2/cpp/test/internal_api/download_test.cc
bmehta001: Ok, will do in a separate PR when I also clean up the platform-specific code into separate files

PR #804, opened 2026-06-15
js: add foundry-local-sdk-winml SKU mirroring the python wheel pattern

bmehta001:
## Summary

Adds a Windows-only `foundry-local-sdk-winml` npm package so JS callers can install WinML EP support out of the box, mirroring the existing python wheel pattern (`foundry-local-sdk` vs `foundry-local-sdk-winml`).

## Why

The standard `foundry-local-sdk` npm tarball ships `foundry_local.dll` alone. The WinML 2.x bootstrapper's `LoadLibraryW("Microsoft.Windows.AI.MachineLearning.dll")` therefore failed silently on JS callers — they only ever saw CUDA + WebGPU, never OpenVINO / NvTensorRTRTX. The vitest `discoverEps` test only asserts shape (`Array.isArray(eps) === true`), not contents, so this gap was never caught in CI.

The python wheel has had this split for a while (`foundry-local-sdk` vs `foundry-local-sdk-winml`); this PR brings the JS SDK to parity.

## What changes

| File | Change |
|------|--------|
| `steps-build-js.yml` | New `isWinML` parameter. When true, stages `Microsoft.Windows.AI.MachineLearning.dll` next to `foundry_local.dll` in `prebuilds/<plat>-<arch>/`. WinML DLL is signed by Microsoft via NuGet — intentionally excluded from the ESRP signing pattern (no re-sign). |
| `steps-pack-js.yml` | New `isWinML` parameter. Rewrites the package name to `foundry-local-sdk-winml` via `npm pkg set name=...` before `npm pack`; restricts the prebuilds-source list to win-x64 + win-arm64. |
| `stages-js.yml` | Refactored to take `variant: base \| winml`. Base = existing 4-platform stages (unchanged). WinML = new `js_build_winml_win_x64`, `js_build_winml_win_arm64`, `js_test_winml_win_x64`, `js_pack_winml` consuming `cpp-native-win-*-winml` artifacts, producing `js-prebuild-winml-win-*` + `js-sdk-winml`. |
| `stages-sdk-v2.yml` | Invoke `stages-js.yml` twice (base + winml), matching the C# and Python coordinators. |
| `sdk_v2/js/README.md` | Document the two SKUs at the top, mirroring the python README's variant table. |

## Distribution flow

| | Standard SKU | WinML SKU |
|---|---|---|
| Package name | `foundry-local-sdk` | `foundry-local-sdk-winml` |
| Platforms | win-x64, win-arm64, linux-x64, osx-arm64 | win-x64, win-arm64 (build-only on arm64) |
| Bundled DLLs (per platform) | `foundry_local.dll` (size unchanged) | `foundry_local.dll` + `Microsoft.Windows.AI.MachineLearning.dll` (~922 KB) |
| Install-time downloads (`install-native.cjs`) | ORT + GenAI from NuGet (~91 MB) | ORT + GenAI from NuGet (~91 MB) |
| EPs discovered out of the box | CUDA + WebGPU | CUDA + WebGPU + OpenVINO + NvTensorRTRTX |

The WinML DLL is small enough (~922 KB) to bundle inline — no install-time network hop for WinML. ORT/GenAI are still NuGet-downloaded at install time for both SKUs.

## Pairs with

This PR is stacked on top of #788 (WinML 2.x EP downloads). Specifically the fix at 01a27256 (`winml: load WinML DLL from foundry_local.dll's own directory`), which makes the WinML bootstrapper resolve `Microsoft.Windows.AI.MachineLearning.dll` relative to `foundry_local.dll`'s own directory via `GetModuleHandleExW` + `LoadLibraryExW(LOAD_WITH_ALTERED_SEARCH_PATH)`. Without that fix the WinML DLL would not be found at runtime even when bundled next to `foundry_local.dll`.

When #788 merges, this PR's base should be re-pointed to `main`.

## Validation

- All four YAML files parse via `yaml.safe_load`.
- `npm pkg set name=foundry-local-sdk-winml` + `npm pkg get name` verified locally (mechanism is idempotent; package.json round-trips cleanly back to `foundry-local-sdk`).
- The new stage layout mirrors the existing `stages-python.yml` variant pattern exactly — same `base | winml` if-block split and same artifact-name convention.

End-to-end verification (the WinML JS tarball returns 4 EPs) will happen on the first pipeline run after this PR's CI. Locally, the same code path was verified in #788 via `npm run copy-native:dev` (which already bundles the WinML DLL) — the new pipeline staging matches the dev script's layout byte-for-byte.

## Out of scope

- No public API changes.
- `install-native.cjs` unchanged. The WinML DLL is bundled at pack time, not downloaded at install time.
- `copy-native.mjs` (dev-only) unchanged — it already includes the WinML DLL for local testing.
- `pack-prebuilds.mjs` (not used by the pipeline) unchanged.


bmehta001:
Superseded by the WinML/non-WinML flavor unification work (branch bhamehta/flcore/unify-winml-sku), which removes the separate WinML SKU entirely rather than adding one per SDK. Closing in favor of that approach.

PR #816, opened 2026-06-18
Integrate "get model versions" and "download specific model version" into cpp Core with max_versions

selenayang888:
Porting the changes "get model versions" and "download specific model version" from C# into C++ Core now:

1. List all available versions of a model (e.g., to see what versions of phi-4 are published).
2. Download and use a specific older version (e.g., for reproducibility, compatibility, or regression testing).


sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -869,6 +877,22 @@ struct flCatalogApi {
   FL_API_STATUS(GetCachedModels, _In_ const flCatalog* catalog, _Outptr_ flModelList** out_models);
   FL_API_STATUS(GetLoadedModels, _In_ const flCatalog* catalog, _Outptr_ flModelList** out_models);
 
+  /// Get all versions of a model, optionally narrowed to a specific variant.
+  /// @param model_alias Alias of the model (e.g. "phi-4-mini"). May be NULL to return
+  ///        all versions of all models (subject to device/EP filtering).
```
skottmckay: Is there a use-case where someone needs to find all versions of all models? That seems like a problematic thing to enable given how much data could be retrieved. 

If we require the alias I assume we have no need to expose pagination which is adding a _lot_ of complexity to the code. 

If a user really really needs to get all versions for all models they can iterate each alias themselves. #Closed

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -869,6 +877,22 @@ struct flCatalogApi {
...
+  /// @param variant_name Optional variant name (e.g. "Phi-4-generic-gpu"). NULL returns
+  ///        every variant.
+  /// @param max_versions Soft upper bound on the number of variants returned.
+  ///        Pass 0 (or any negative value) for no cap.
```
skottmckay: Is the max on the versions or the variants? This feels like a really confusing value as I can think of various permutations between version and variant.

e.g. say I have model A with CPU (v1, 2, 3) and GPU variants (v 2, 3, 4).

'latest 2 versions' could be anything that is v3 or v4 (per alias), or it could be v2 and v3 for CPU and v3 and v4 for GPU (per variant).

I'm not sure which is best though. If versioning tends to be per variant, selecting the latest X for the variant might make more sense.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -869,6 +877,22 @@ struct flCatalogApi {
...
+  /// @param continuation_token Opaque cursor returned by a previous call used to resume
+  ///        pagination from the underlying catalog source. NULL or empty starts from
+  ///        the beginning. Implementations that do not paginate ignore this argument.
+  /// Returned list contains existing flModel handles owned by the catalog; releasing
+  /// the list does not invalidate the underlying model handles.
+  FL_API_STATUS(GetModelVersions, _In_ const flCatalog* catalog, _In_opt_ const char* model_alias,
+                _In_opt_ const char* variant_name, int32_t max_versions,
```
skottmckay: nit: would 'model_name' be more consistent given this value comes from Info_GetName (ModelInfo.Name) doesn't it?

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -203,6 +207,32 @@ std::vector<std::vector<CatalogFilter>> BuildSearchFilters(const IEpDetector& ep
   return filter_sets;
 }
 
+/// Build per-device filter sets for an all-versions query.
+/// Same shape as `BuildSearchFilters` minus the `labels=latest` filter
+/// (so older versions are included). When `model_alias` is non-empty an
+/// `annotations/tags/alias` filter scopes the result to that one alias; when
+/// empty, no alias filter is added and every versioned model the local hardware
+/// can run is returned across all its versions.
+std::vector<std::vector<CatalogFilter>> BuildAllVersionsFilters(const IEpDetector& ep_detector,
```
skottmckay: Can we add a param to BuildSearchFilters for 'all versions' or 'latest only' and linclude the 'labels' filter based on that instead of duplicating the logic here?

That said, shouldn't the model name be passed in here and used in the filter if provided so all filtering is server side? #Closed

sdk_v2/cpp/src/catalog/base_model_catalog.cc
```diff
@@ -86,6 +87,52 @@ bool CompareModelsForSort(const Model& m1, const Model& m2) {
   return created1 > created2;
 }
 
+int CompareCaseInsensitive(const std::string& lhs, const std::string& rhs) {
```
skottmckay: nit: might be good to put this in utils/string_utils.h #Closed

sdk_v2/cpp/src/catalog/catalog_client.h
```diff
@@ -27,6 +34,29 @@ class ICatalogClient {
...
+  /// previous call used to resume pagination; empty starts from the beginning.
+  /// On return, `PagedModelInfos::next_continuation_token` is set when more
+  /// data is available; an empty value means the underlying source has been
+  /// fully walked. Implementations that cannot list older versions return
+  /// whatever they have locally with an empty token.
+  ///
+  /// Provided with a default `{}` body so an implementation that has not yet
+  /// overridden it still compiles.
+  virtual PagedModelInfos FetchAllVersionsByAlias(
+      const std::string& /*model_alias*/,
```
skottmckay, line 45: Would have expected the model name to be passed in here if it can be used in the filter. #Closed

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -255,57 +356,68 @@ AzureCatalogClient::AzureCatalogClient(const std::string& base_url,
...
         logger_.Log(LogLevel::Warning,
                     std::string("catalog: filter set failed across all regions: ") + ex.what());
-        return std::nullopt;
+        result.aborted = true;
+        return result;
       }
     } else if (regional) {
-      // Subsequent pages are pinned to the region that served page 1 — a filter set
-      // never mixes regions (continuation tokens are region-specific).
       response = http_post_response_(BuildRegionalUrl(url_prefix_, url_suffix_, *pinned_region), body);
```
skottmckay, line 312: AI likes randomly deleting existing comment. Was it intentional to do so?  #Closed

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: If we can remove the complexity from public pagination the implementation for fetching all versions of an alias or model name should be pretty similar to AzureCatalogClient::FetchModelsByIds, just with a slightly different filter.

https://github.com/microsoft/Foundry-Local/blob/01939b17248324dd90f9cb1ac67352081cfc78af/sdk_v2/cpp/src/catalog/azure_catalog_client.cc#L373-L385

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
selenayang888: I eliminated duplicate code through parametrization `bool latest_only`; Added server-side model name filtering (new model_name parameter); Consolidated both filters into one unified implementation

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
selenayang888: Restored the deleted comment. It was not intentional — the AI accidentally dropped it.

sdk_v2/cpp/src/catalog/catalog_client.h
selenayang888: After model_name is used in the filter, I've added the model_name parameter here and throughout the SDK.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
selenayang888: Made the major change here: the `model_alias` is required, throw exception when it is empty or null.
Besides, removed the arg 'continuation token` for less complexity.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
selenayang888: Made the corresponding changes. Currently, the arg 'max_versions' is for selecting the latest X versions for the given model alias per variant.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
selenayang888: Thanks. I changed it to 'model_name'.

sdk_v2/cpp/src/catalog/base_model_catalog.cc
selenayang888: "CompareCaseInsensitive" function is properly moved to string_utils.h as an inline function.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -597,6 +597,12 @@ typedef struct flApi {
   size_t FL_API_T(ModelList_Size, _In_ const flModelList* models);
   flModel* FL_API_T(ModelList_GetAt, _In_ const flModelList* models, size_t idx);
 
+  /// Returns the continuation token associated with this model list. The pointer
+  /// is owned by the list and remains valid until the list is released. NULL
+  /// when no continuation token was set (e.g. the list came from GetModels(),
+  /// or GetModelVersions has walked the underlying source to exhaustion).
+  const char* FL_API_T(ModelList_GetContinuationToken, _In_ const flModelList* models);
```
skottmckay: Can this be removed along with next_continuation_token in flModelList?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -869,6 +875,17 @@ struct flCatalogApi {
...
+  /// Get all versions of a model alias, optionally narrowed to a specific model name.
+  /// @param model_alias Alias of the model (e.g. "phi-4-mini"). Must be non-NULL and non-empty.
+  /// @param model_name Optional model name (ModelInfo.Name, e.g. "Phi-4-generic-gpu"). NULL returns
+  ///        every model name.
+  /// @param max_versions Select latest X versions per model name. Pass 0 (or any
+  ///        negative value) for no per-model-name cap.
+  /// Returned list contains existing flModel handles owned by the catalog; releasing
+  /// the list does not invalidate the underlying model handles.
+  FL_API_STATUS(GetModelVersions, _In_ const flCatalog* catalog, _In_ const char* model_alias,
+                _In_opt_ const char* model_name, int32_t max_versions, _Outptr_ flModelList** out_models);
```
skottmckay, line 969: Should we note that (IIUC) these models will be returned by catalog operations in the current instance (e.g. list models) but are not saved in the model cache on disk so won't be present on restart? I _think_ that behavior is fine but it could be a little surprising.  #Closed

sdk_v2/cpp/src/catalog.h
```diff
@@ -9,6 +9,11 @@
 
 namespace fl {
 
+/// Model variants returned by `ICatalog::GetModelVersions`.
+struct ModelVersionsPage {
+  std::vector<Model*> models;
+};
```
skottmckay: do we need this struct anymore?

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -255,57 +275,70 @@ AzureCatalogClient::AzureCatalogClient(const std::string& base_url,
...
+  result.region = region_in;
+  std::optional<std::string> pinned_region;
+  if (!region_in.empty()) {
+    pinned_region = region_in;  // resuming a previously-pinned filter set
+  }
 
   while (true) {
-    const std::string body = BuildRequestBody(filters, skip, continuation_token);
+    int requested_page_size = kPageSize;
+    if (max_count > 0) {
```
skottmckay: Does anything set max_count? If not can we remove the code related to it?  #Closed

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -255,57 +275,70 @@ AzureCatalogClient::AzureCatalogClient(const std::string& base_url,
...
+    const std::string body = BuildRequestBody(filters, skip, inner_token, requested_page_size);
 
     http::HttpResponse response;
     if (regional && !pinned_region) {
-      // Page 1: run through region fallback starting from the sticky region (last known-good) or the active region.
-      // Exhaustion means every candidate had a retryable region-health failure, so fail just this filter set.
-      const std::string start =
-          region_fallback_.StickyRegion().value_or(region_);
+      // Page 1 fresh start: run region fallback starting from the sticky/active region.
+      const std::string start = region_fallback_.StickyRegion().value_or(region_);
```
skottmckay, line 295: nit: removed comments that seemed like they were more descriptive of the processing here and lines 296, 302-303. #Closed

sdk_v2/cpp/src/catalog/azure_catalog_client.h
```diff
@@ -52,18 +52,42 @@ class AzureCatalogClient : public ICatalogClient {
...
-  /// pinned to the serving region; a retryable region-health failure fails just this filter set (others continue).
-  std::optional<FetchedFilterSet> FetchFilterSet(const std::vector<CatalogFilter>& filters);
+  /// Walks pages of one filter set with explicit pagination state.
+  /// - `skip` / `inner_token`: starting cursor (nullopt for a fresh start).
+  /// - `region_in`: pinned region for resuming (empty triggers page-1 region fallback).
+  /// - `max_count`: soft cap; 0 means walk to exhaustion.
+  /// On success, returns `done=true` if the stream was exhausted, or `done=false`
+  /// with `next_skip`/`next_inner_token` populated when stopped by the cap.
+  /// On a retryable region failure, returns `aborted=true` (partial results discarded).
+  FilterSetWalk FetchFilterSetWithState(const std::vector<CatalogFilter>& filters,
```
skottmckay: Do we need new code here? The existing code was capable of fetching all the latest models, so naively I would have expected that infrastructure to be sufficient to do the fetch for all versions given it's the same process just with a different filter and some post-fetch processing to apply the max versions isn't it? 

AI summary when asked 

> compare to base and determine what of the new code is actually required. do we need FilterSetWalk or was the original FetchedFilterSet sufficient given we're doing pagination internally and there doesn't seem to be any actual usage of max_count.
we do need to build a new filter for the 'fetch all versions', and we do need to post-process to apply max_versions, but I'm wondering if we can do that with more minimal changes

```
Confirmed — I've now traced the whole chain. Here's the architectural read.

## What the new code actually added to the client

The branch replaced the original `FetchedFilterSet` / `FetchFilterSet` pair with a heavier `FilterSetWalk` / `FetchFilterSetWithState`. The new surface adds:

- `FilterSetWalk` fields: `next_skip`, `next_inner_token`, `done`, `aborted`
- `FetchFilterSetWithState` params: `skip`, `inner_token`, `region_in`, `max_count`
- `BuildRequestBody` gained a `page_size` parameter (to honor `max_count`)

## The key finding: it's all dead weight

I checked every call site of `FetchFilterSetWithState`:

| Caller | `max_count` | resumes a cursor? |
|--------|-------------|-------------------|
| `FetchAllFilterSets` | `0` | no |
| `FetchModelsByIds` | `0` | no |
| `FetchAllVersionsByAlias` | `0` | no |

**Every call passes `max_count=0` and a fresh start (`nullopt, nullopt, ""`).** Nothing ever resumes a walk. So:

- `next_skip` / `next_inner_token` are written but never read.
- `done` is always `true` on a non-aborted return when `max_count=0`, making it redundant with `!aborted`.
- `region_in` / `skip` / `inner_token` resume params are always defaulted.
- `BuildRequestBody`'s `page_size` is always `kPageSize`.

And critically — `ModelVersionsPage` is just `{ std::vector<Model*> models; }`. **There is no continuation token at the versions layer.** `GetModelVersions` always walks the source to exhaustion (your earlier `ModelList_GetContinuationToken` comment says exactly this). So there is no pagination consumer anywhere that the resumable-cursor machinery could serve.

There's a second piece of dead surface: `ICatalogClient::FetchAllVersionsByAlias` takes `max_versions`, but the implementation ignores it (`/*max_versions*/`) — the per-variant cap is correctly applied up in `base_model_catalog.cc::GetModelVersions`. So `max_versions` shouldn't be on the client method at all.

## What's genuinely required

1. **`BuildSearchFilters` gaining `latest_only` / `model_alias` / `model_name`** — yes, this is the real new capability: building the "drop `labels=latest`, add alias filter" query. Keep it.
2. **`FetchAllVersionsByAlias` (client + `ICatalogClient`)** — the new feature. Keep it, but it can just loop the per-device filter sets calling the *original* `FetchFilterSet`, exactly like `FetchModelsByIds` already does. Drop the unused `max_versions` param.
3. **Base-catalog plumbing** — `ModelVersionsPage`, `FetchModelVersions` virtual, `GetModelVersions`, `IntegrateVariantsLocked`, and the `max_versions` post-processing. All required and correctly placed.

## Recommendation

Revert the client back to `FetchedFilterSet` + `FetchFilterSet(filters)` (the `std::optional`-returning form — `nullopt` already expresses "aborted" more cleanly than an `aborted` bool), and revert the `BuildRequestBody` `page_size` change. Then `FetchAllVersionsByAlias` becomes:

```cpp
std::vector<ModelInfo> AzureCatalogClient::FetchAllVersionsByAlias(
    const std::string& model_alias, const std::string& model_name) {
  const auto filter_sets = BuildSearchFilters(ep_detector_, model_filter_,
                                              /*latest_only=*/false, model_alias, model_name);
  std::vector<ModelInfo> result;
  for (const auto& filters : filter_sets) {
    if (auto walk = FetchFilterSet(filters)) {
      auto batch = ToModelInfos(walk->models, walk->region);
      result.insert(result.end(), std::make_move_iterator(batch.begin()),
                    std::make_move_iterator(batch.end()));
    }
  }
  return result;
}
```

That keeps the genuinely-new filter construction and the new public method, drops ~all the state-machine churn, and shrinks the azure_catalog_client.cc diff substantially. The only reason to keep `FilterSetWalk` would be a near-term plan to expose incremental version pagination through `ModelVersionsPage` — and since that struct has no token field, there's no such plan in this PR.

Want me to make these reductions? It's a mechanical, behavior-preserving revert of the client layer — clean `@CppCoder` work, and I'd have `@Tester` confirm the existing `c_api_test.cc` version tests still pass since they exercise `FetchAllVersionsByAlias` end-to-end.
```
 #Closed

File: sdk_v2/cpp/src/catalog/base_model_catalog.cc [hunk 14]
```diff
@@ -329,4 +473,94 @@ std::vector<Model*> BaseModelCatalog::GetLoadedModels() const {
...
+  // underlying source produced them — important for stable pagination.
+  std::vector<std::string> fetched_ids;
+  fetched_ids.reserve(fetched.models.size());
+  for (const auto& m : fetched.models) {
+    fetched_ids.push_back(m.Info().model_id);
+  }
+
+  if (!fetched.models.empty()) {
+    std::lock_guard<std::mutex> lock(mutex_);
+    IntegrateVariantsLocked(std::move(fetched.models));
```
skottmckay: nit: Can we lock the mutex in IntegrateVariants (and drop the 'Locked' from the name) instead of duplicating locking logic at all call sites?

sdk_v2/cpp/src/catalog/azure_model_catalog.cc
```diff
@@ -138,4 +141,121 @@ std::vector<Model> AzureModelCatalog::FetchModels() const {
...
+std::vector<std::pair<std::string, std::optional<std::string>>> EnumerateEndpoints(
+    const std::vector<std::pair<std::string, std::optional<std::string>>>& configured,
+    const char* default_url,
+    const char* default_filter) {
+  if (!configured.empty()) {
+    return configured;
+  }
+
+  return {{default_url, std::optional<std::string>(default_filter)}};
+}
```
skottmckay: given catalog_urls_ is only set at construction time, can we default that value to kDefaultCatalogUrl and  kDefaultCatalogFilter if not specified so we don't need to do anything here or on line 110 (lose the special case there)?  #Closed

sdk_v2/cpp/src/catalog/azure_model_catalog.cc
```diff
@@ -138,4 +141,121 @@ std::vector<Model> AzureModelCatalog::FetchModels() const {
...
+  FetchedModelVersions out;
+  if (cache_only_) {
+    // In cache-only mode we have no remote source to query for older versions.
+    logger_.Log(LogLevel::Debug,
+                "FetchModelVersions skipped: catalog is in cache-only mode.");
+    return out;
+  }
+
+  // Scan local models so any version already on disk is reported as cached.
+  auto local_models = ScanLocalModels(cache_dir_, logger_);
```
skottmckay: Do we need to scan local models here and in FetchModelsByIds? That seems unrelated to fetching the latest versions/ids. We want to do it when we do the general fetch, but I don't think it's required in these targeted fetches. #Closed

sdk_v2/cpp/src/catalog/base_model_catalog.h
```diff
@@ -46,6 +49,31 @@ class BaseModelCatalog : public ICatalog {
...
+  /// Maps to C# `BaseModelCatalog.GetModelVersionsAsync` -> derived overrides.
+  struct FetchedModelVersions {
+    std::vector<Model> models;
+  };
+
+  virtual FetchedModelVersions FetchModelVersions(
+      const std::string& /*model_alias*/,
+      const std::string& /*model_name*/ = "") const {
+    return {};
+  }
```
skottmckay: Do we need the FetchedModelVersions struct? #Closed

sdk_v2/cpp/src/catalog/base_model_catalog.cc
```diff
@@ -329,4 +473,94 @@ std::vector<Model*> BaseModelCatalog::GetLoadedModels() const {
...
+    return {};
+  }
+
+  // Capture fetch order before moving so we can return models in the order the
+  // underlying source produced them — important for stable pagination.
+  std::vector<std::string> fetched_ids;
+  fetched_ids.reserve(fetched.models.size());
+  for (const auto& m : fetched.models) {
+    fetched_ids.push_back(m.Info().model_id);
+  }
```
skottmckay: Is this still needed?

sdk_v2/cpp/src/catalog/base_model_catalog.cc
```diff
@@ -86,6 +91,28 @@ bool CompareModelsForSort(const Model& m1, const Model& m2) {
   return created1 > created2;
 }
 
+// Deterministic API output order: alias alpha, then name alpha, then version asc.
+bool CompareModelPointersForVersionList(const Model* lhs, const Model* rhs) {
```
skottmckay: Should this have the same ordering as CompareModelsForSort? Currently the device is ignored so things like 'cuda' and 'generic-gpu' will come before 'npu' which would differ from the model list output.

When we add to the model variants in IntegrateVariantsLocked do we end up with a weird ordering given AddVariant appends to the original list, which was sorted using CompareModelsForSort?

This branch off yours has an alternative approach where the variant sorting is pushed down to Model given we're now adding new variants in multiple places: skottmckay/get-model-versions-with-variant-sort 

It also uses that ordering when returning values from `GetModelVersions` for consistency.

sdk_v2/cpp/src/manager.cc
```diff
@@ -365,6 +365,22 @@ Manager& Manager::Create(const Configuration& config) {
...
+  // the C# FoundryLocalCore IS_WINML path. Only meaningful in WinML builds; outside that
+  // configuration TryInitializeWindowsAppSdk is a no-op stub.
+#if defined(FOUNDRY_LOCAL_USE_WINML) && FOUNDRY_LOCAL_USE_WINML
+  {
+    if (IsAdditionalOptionEnabled(config, "Bootstrap")) {
+      StderrLogger bootstrap_logger;
+      TryInitializeWindowsAppSdk(bootstrap_logger);
+    }
+  }
+#endif
```
skottmckay: Assuming this is unintentional

sdk_v2/cpp/src/catalog.h
selenayang888: Removed this struct `ModelVersionsPage`.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
selenayang888: Yes, I removed this API since the  next_continuation_token was already removed.

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
selenayang888: I reverted this function `FetchFilterSetWithState` back to `FetchFilterSet`, since the `continuation_token` is already removed from `getModelVersions`.

sdk_v2/cpp/src/catalog/base_model_catalog.h
selenayang888: Since it is only a thin wrapper around a vector, so removing and replacing it with `std::vector<Model>` in the base virtual method.

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
selenayang888: As the function `FetchFilterSetWithState` is reverted, no `max_count_` anymore.

sdk_v2/cpp/src/catalog/azure_catalog_client.h
selenayang888: The function FetchFilterSetWithState and all related things are removed and reverted back to FetchFilterSet.

sdk_v2/cpp/src/manager.cc
selenayang888: It was unintentional and reverted the changes.

sdk_v2/cpp/src/catalog/azure_model_catalog.cc
selenayang888: Implemented the refactor by normalizing `catalog_urls` at construction time to always contain the default (kDefaultCatalogUrl, kDefaultCatalogFilter) when empty, then remove the now-redundant empty checks in `FetchModels` and endpoint enumeration.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
selenayang888: I think this is how design doc specified for caching strategy. `get_model_versions` results are not cached in `BaseModelCatalog`. Each call to `get_model_versions` does a fresh catalog query. However, to support `download` an older version", the resolved ModelInfo for that version is temporarily added to the catalog's modelIdToInfo index.

The link below is the part in the design doc for Caching Strategy for All-Versions Data:

 [model-versions-design.md](https://microsofteur-my.sharepoint.com/personal/samkemp_microsoft_com/_layouts/15/onedrive.aspx?id=%2Fpersonal%2Fsamkemp%5Fmicrosoft%5Fcom%2FDocuments%2FDocuments%2Fspecs%5Fmd%2Fmodel%2Dversions%2Fmodel%2Dversions%2Ddesign%2Emd&parent=%2Fpersonal%2Fsamkemp%5Fmicrosoft%5Fcom%2FDocuments%2FDocuments%2Fspecs%5Fmd%2Fmodel%2Dversions&ct=1778779862828&or=Teams%2DHL&ga=1#:~:text=4.1.7%20Caching%20Strategy%20for%20All%2DVersions%20Data)

skottmckay: should we explain that behavior to the user in the comments here as it's a little opaque?

sdk_v2/cpp/src/catalog/azure_model_catalog.cc
skottmckay: Has this change been pushed?

selenayang888: Just pushed the commit. Thanks for your reminder.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
selenayang888: Added the explanation of the behavior in the comments now.
Moreover, based on the design doc requirements for caching strategy, I updated the `GetModelVersions` in `base_model_catalog.cc:475` no longer calls `IntegrateVariants`. It now stores the fetched models in transient per-catalog query storage and returns those handles without adding them to the main alias/id indices.

sdk_v2/cpp/src/catalog/base_model_catalog.h
```diff
@@ -79,13 +105,23 @@ class BaseModelCatalog : public ICatalog {
...
   /// Build lookup indices from the current models_ collection.
   /// Builds a complete new ModelIndex locally, then atomically swaps it into index_.
   void RebuildIndex() const;
 
   /// Thread-safe access: ensures catalog is populated, refreshes if allowed and stale.
   void EnsurePopulated(bool allow_refresh = false) const;
 
+  /// Transient storage for the most recent GetModelVersions query. Replaced on each call.
+  /// These models are intentionally not integrated into the main lookup indices.
+  mutable std::vector<std::unique_ptr<Model>> version_query_models_;
```
skottmckay, line 127: Replacing on each call creates potential issues on the client side if they call this for multiple models as it would invalidate all info from the previous fetch and they have no way to keep that info alive.

Could we store in an unordered_map<string_view, unique_ptr<Model>> that uses the model id as the key?

sdk_v2/cpp/src/catalog/base_model_catalog.cc
```diff
@@ -329,4 +472,89 @@ std::vector<Model*> BaseModelCatalog::GetLoadedModels() const {
...
+    }
+
+    result.reserve(version_query_models_.size());
+    for (auto& model : version_query_models_) {
+      if (!variant_name.empty() && model->Info().name != variant_name) {
+        continue;
+      }
+
+      result.push_back(model.get());
+    }
```
skottmckay, line 427: My concern with this is that we're using a different structure to the Model's returned by the default list, which have a parent (a.k.a. 'container') Model instance for the alias and the variants are accesses from that. Here we're returning individual variants and no parent Model instance is created. 

That means usage could differ in the user code, and ideally we can avoid that. 

Given we're fetching a single alias, could we create the parent Model and instead use AddVariant to add each individual result and return the parent? That would mean the API returns a single Model though. #Closed

skottmckay: Actually we could do both. Internally create the same structure (parent container Model with leaf variants), and return a flModelList of the variants at the C API level like we do in Model_GetVariantsImpl. 

That creates a consistent setup internally whilst returning the expected list of individual variant Model instances.

sdk_v2/cpp/src/catalog/base_model_catalog.cc
selenayang888: The local scan removal from `FetchModelVersions` and the transient `version_query_models_` approach are both solid and fully address the first and main concern in the comments. 
Then, the three major changes:
1. Added container re-sorting support in `model.h` and `model.cc`.
2. Applied that re-sort in integration flow in `base_model_catalog.cc`
3. Fixed `max_versions` selection robustness in `base_model_catalog.cc`

selenayang888: Thanks for sharing the branch — I've merged it in (after the latest origin/main merge). All are completed as your suggested in teams chat:
1. Make Model own ordering invariant
2. Move comparator/device-priority logic from BaseModelCatalog to Model
3. Add default-selection step in Model
4. Remove duplicated ordering work in GetModelVersions and integration paths.

sdk_v2/cpp/src/catalog/base_model_catalog.h
selenayang888: Fixed it as below:
1. Each alias gets its own bucket; querying alias B no longer clears alias A's results.
2. Replaced the single `vector<unique_ptr<Model>>` with: `mutable std::unordered_map<std::string, std::vector<std::unique_ptr<Model>>> version_query_models_`;

sdk_v2/cpp/src/catalog/base_model_catalog.cc
selenayang888: I did both as you suggested: creating a consistent setup internally whilst returning the expected list of individual variant Model instances.

- `version_query_models_` now stores a single container `Model` per alias (via `MakeContainer` + `AddVariant`), mirroring the structure used by `PopulateModels`/`IntegrateVariants`.
- Variant `Model*` pointers are extracted from the container via `Variants()` — the same path `Model_GetVariantsImpl` uses.
- At the C API level, `Catalog_GetModelVersionsImpl` already wraps these into a `flModelList` (no change needed there).

sdk_v2/cpp/src/catalog/base_model_catalog.h
```diff
@@ -46,6 +51,27 @@ class BaseModelCatalog : public ICatalog {
   /// Maps to C# FetchModelInfoAsync.
   virtual std::vector<Model> FetchModels() const = 0;
 
+  /// Derived classes implement this to fetch all versions of a model from the
+  /// underlying catalog source, bypassing the "latest only" filter.
+  /// Returns the variants (in any order; the base class sorts/indexes them)
+  /// for the given alias.
```
skottmckay: These should automatically be sorted by AddVariant so should be in the same priority order as the model list output.

sdk_v2/cpp/src/catalog/base_model_catalog.cc
```diff
@@ -329,4 +369,100 @@ std::vector<Model*> BaseModelCatalog::GetLoadedModels() const {
...
+
+    // Build a container Model for this alias, mirroring the structure used by
+    // PopulateModels / IntegrateVariants. AddVariant maintains best-first order.
+    auto container = Model::MakeContainer(std::move(fetched.front()));
+    for (size_t i = 1; i < fetched.size(); ++i) {
+      container.AddVariant(std::move(fetched[i]));
+    }
+
+    container.SelectDefaultVariant();
+    version_query_models_[model_alias] = std::make_unique<Model>(std::move(container));
```
skottmckay: If someone had previously queried for this alias this line is going to invalidate the previously returned flModelList. The comment in the C and C++ code says that never happens.

sdk_v2/cpp/src/catalog/base_model_catalog.cc
```diff
@@ -329,4 +369,100 @@ std::vector<Model*> BaseModelCatalog::GetLoadedModels() const {
...
+      logger_.Log(LogLevel::Information,
+                  fmt::format("GetModelVersions: alias '{}' not found in catalog.", model_alias));
+    }
+  }
+
+  // Sort into best-first order. Within the same variant name, device priority is identical,
+  // so this also groups by name with latest version first — suitable for max_versions capping.
+  std::stable_sort(result.begin(), result.end(), [](const Model* a, const Model* b) {
+    return Model::CompareBestFirst(*a, *b);
+  });
```
skottmckay: I didn't expect sorting would be required here as Model::AddVariant sorts on insert to keep the vector of variants in the priority order, so you should be able to use the list returned from Model::Variants() as-is.

skottmckay:
Will the other language bindings have GetModelVersions exposed in this PR or a separate one? AI should be able to handle adding that fairly easily.

sdk_v2/cpp/src/catalog/base_model_catalog.cc
selenayang888: Yes, it is unnecessary. I removed `std::stable_sort` and its comment.

sdk_v2/cpp/src/catalog/base_model_catalog.h
selenayang888: Updated the docstring since ` AddVariant` handles ordering automatically.

selenayang888:
> Will the other language bindings have GetModelVersions exposed in this PR or a separate one? AI should be able to handle adding that fairly easily.

This PR for the implementation of GetModelVersion is only for C++ Core. I think we will have separate PR for other language bindings in the future.

sdk_v2/cpp/src/catalog/base_model_catalog.cc
skottmckay: I don't think we need to support this - we just need to fix the comments to call it out.

selenayang888: Made the code match the documented contract. By appending (push_back) instead of replacing, old containers are never destroyed, so all Model* pointers truly remain valid for the catalog's lifetime

skottmckay (approved):

skottmckay (dismissed):

skottmckay (approved):

PR #824, opened 2026-06-19
[SDKv2] Unify the WinML and non-WinML flavors into one package per SDK

bmehta001:
## Summary

Collapses the separate **WinML** and **non-WinML** SDK flavors into a **single package per SDK** (C#, Python, JS). Consumers no longer pick between a `.WinML` SKU and a base SKU: on Windows the reg-free **WinML 2.x EP catalog** is always available, and other platforms are unaffected.

## What changed

- **native** — always enable the WinML EP catalog on Windows; drop the build-flavor split.
- **C#** — single `Microsoft.AI.Foundry.Local` package; drop the `.WinML` SKU and the `UseWinML` switch.
- **Python** — single wheel; drop the WinML variant.
- **JS** — bundle the WinML 2.x runtime in the npm package on Windows.
- **macOS** — simplify ORT dylib staging to mirror Linux.
- **CI (`.pipelines/v2`)** — collapse the WinML/non-WinML matrix into one build per SDK.
- **samples (C#)** — use the unified package; central package management via `Directory.Packages.props`.
- **docs / www** — describe the unified package; one install command per SDK in the download dropdown.
- **JS exit fix** — clean process exit after loading a native Manager (`beforeExit`/`exit` handlers), avoiding the ORT-teardown crash on graceful Node exit.

## Notes / follow-ups

- Rebased onto `main` (picks up #813 `sdk_legacy` removal, #821 vision samples, #810 npm bump, #806).
- The #821 vision samples (C#/JS/Rust) landed after this branch and are **not yet** migrated to the unified package / central package management — follow-up needed.
- **Draft**: pending CI (multi-SDK pack + build) verification.


.pipelines/v2/templates/steps-build-js.yml
```diff
@@ -145,6 +146,16 @@ steps:
...
       Write-Host "  staged $(nativeLib)"
 
+      # WinML 2.x runtime ships next to foundry_local.dll so WinML EPs work out
+      # of the box. It is Microsoft-signed (NuGet) — excluded from ESRP signing.
+      $winmlDll = Join-Path "${{ parameters.nativeArtifactDir }}" 'Microsoft.Windows.AI.MachineLearning.dll'
+      if (Test-Path $winmlDll) {
+        Copy-Item $winmlDll -Destination $dst -Force
+        Write-Host "  staged Microsoft.Windows.AI.MachineLearning.dll"
+      } else {
+        Write-Warning "WinML runtime DLL not found in native artifact: $winmlDll"
```
skottmckay: Does a warning cause the build to fail? #Closed

samples/cs/Directory.Packages.props
```diff
@@ -5,8 +5,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageVersion Include="Microsoft.AI.Foundry.Local" Version="*-*" />
-    <PackageVersion Include="Microsoft.AI.Foundry.Local.WinML" Version="*-*" />
-    <PackageVersion Include="Betalgo.Ranul.OpenAI" Version="9.1.0" />
+    <PackageVersion Include="Betalgo.Ranul.OpenAI" Version="9.2.0" />
```
skottmckay: Does this work now? v1 samples were broken by this update previously and I had to revert to 9.1. breaks at runtime so you need to execute the sample.

See https://github.com/microsoft/Foundry-Local/pull/724 #Closed

sdk_v2/js/src/foundryLocalManager.ts
```diff
@@ -13,6 +13,41 @@ import {
...
 
+// Once a native Manager exists, ONNX Runtime's process-wide teardown races with
+// Node's environment teardown on a natural exit and crashes — a long-standing
+// ORT-on-exit issue, independent of this SDK. It is avoided only by releasing
+// the Manager and then leaving via `process.exit()`, which skips that graceful
+// teardown. So track live managers and do exactly that on the way out: dispose
+// on `beforeExit` then exit explicitly; an `exit` handler covers callers who
+// invoke `process.exit()` themselves (releasing the env before the C runtime
+// tears the ORT libraries down keeps their static destructors benign).
+const liveManagers = new Set<FoundryLocalManager>();
```
skottmckay: Only one manager can be alive at any point in time (enforced at the native level), so is a `Set` needed? #Closed

sdk_v2/js/src/foundryLocalManager.ts
```diff
@@ -13,6 +13,41 @@ import {
 } from "./detail/native.js";
 import type { EpDownloadResult, EpInfo } from "./types.js";
 
+// Once a native Manager exists, ONNX Runtime's process-wide teardown races with
+// Node's environment teardown on a natural exit and crashes — a long-standing
+// ORT-on-exit issue, independent of this SDK. It is avoided only by releasing
```
skottmckay, line 18: Was there a specific error? There was an issue with the log callback causing a crash at shutdown that https://github.com/microsoft/Foundry-Local/pull/835 should fix. Wondering if that was the root cause here and we don't need this new complexity, or whether there's something else going on.

sdk_v2/cs/src/Microsoft.AI.Foundry.Local.csproj
```diff
@@ -81,24 +73,6 @@
...
-      <!--
-        The C# layer is a pure P/Invoke wrapper over foundry_local.dll. No source
-        references any Windows.* / Microsoft.Windows.* / WinRT API, so we share
-        the cross-platform net8.0;net9.0 TFM with the non-WinML SKU. The WinML
-        OS-version floor (Win10 19H1+) is enforced in C++ by
-        winml_ep_bootstrapper.cc (delay-loaded WindowsML::Api + OSVERSIONINFOW
-        check) and by the WinML EP catalog NuGet itself.
-      -->
-      <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
-      <RuntimeIdentifiers>win-x64;win-arm64</RuntimeIdentifiers>
```
skottmckay, line 100: This previously removed `netstandard2.0` from the target RIDs. Is that RID gracefully handled by winml internals now? #Closed

samples/cs/Directory.Packages.props
bmehta001: Thanks - unintentional. Reverted.

.pipelines/v2/templates/steps-build-js.yml
bmehta001: No - will fix so the build fails

sdk_v2/js/src/foundryLocalManager.ts
bmehta001: No, it's not - fixed

sdk_v2/cs/src/Microsoft.AI.Foundry.Local.csproj
bmehta001: Yes, we fail gracefully in `winml_ep_bootstrapper.cc`

sdk_v2/js/src/foundryLocalManager.ts
bmehta001: This may not be the best way to fix it, but I keep having an issue when exiting an app using the JS SDK where the app has "crashed" according to procdump, and there is a heap corruption or access violation. From CoPilot's investigation, this seems to be a race condition with ORT/GenAI/EP being torn down with the Node process being torn down, so this is the fix came up with it. I can remove it from this PR, and we can perhaps think of a better way to address this issue.

skottmckay (approved):

PR #835, opened 2026-06-24
Fix a shutdown issue and make debugging of CUDA or WebGPU EPs easier

skottmckay:
* Handle late logging statements from static cleanup more gracefully. 
  * e.g. webgpu has a callback that is triggered on wgpu instance destruction which happens during static destruction due to globals in genai 
    * the globals in genai aren't required for plugin EPs but that's a separate issue/cleanup
* Make it possible to drop in a custom webgpu or cuda EP for debugging.
  * add env var for specifying path to load library from

### DIFF
--- a/sdk_v2/cpp/src/manager.cc
+++ b/sdk_v2/cpp/src/manager.cc
@@ -41,6 +41,7 @@ namespace fl {
 
 namespace {
 
+std::atomic<ILogger*> s_ort_logger{nullptr};
 std::atomic<ILogger*> s_oga_logger{nullptr};
 
 bool IsTruthyConfigValue(const std::string& value) {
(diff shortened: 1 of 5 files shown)


sdk_v2/cpp/src/manager.cc
```diff
@@ -304,9 +305,15 @@ Manager::Manager(const Configuration& config)
       config_.external_service_url.has_value(),
       config_.catalog_region.value_or("auto"),
       disable_region_fallback);
+
+  s_ort_logger.store(logger_.get(), std::memory_order_release);
```
edgchen1: should we set `s_ort_logger` before the call to `ort_api_->CreateEnvWithCustomLogger()` just to be sure we get any initial logs?

sdk_v2/cpp/src/manager.cc
```diff
@@ -304,9 +305,15 @@ Manager::Manager(const Configuration& config)
...
 }
 
 Manager::~Manager() {
+  // ORT may still emit late teardown logs from internal static cleanup after Manager destruction.
+  // Route those callbacks to a null logger to avoid dereferencing a dangling pointer.
+  s_ort_logger.store(nullptr, std::memory_order_release);
```
edgchen1: similarly, should we clear `s_ort_logger` after `ort_env_` is released? this code could also be located near where `s_oga_logger` is cleared.

skottmckay: The ort_env_ release isn't overly relevant as GenAI is going to keep that alive until process shutdown. Moved it down to the end though.

edgchen1 (approved):

PR #838, opened 2026-06-24
Read Tool & Reasoning Tokens from GenAI API instead of Catalog Metadata

sayanshaw24:
# Migrate tool/reasoning metadata from catalog to GenAI + optimize decode loop

## Summary

Two goals:

1. **Catalog v2 migration**: Tool-call and reasoning token metadata (`toolCallStart`, `toolCallEnd`, `reasoningStart`, `reasoningEnd`) is being removed from the catalog. This PR sources it from GenAI's `genai_config.json` instead, via the new `OgaTokenizer::GetBotTokenId()` / `GetEotTokenId()` / `GetBorTokenId()` / `GetEorTokenId()` APIs, and writes the decoded strings into `ModelInfo` so all existing downstream consumers continue working.

2. **Decode loop optimization**: The current per-token detection of special tokens (double-decode + `string.find()`) is replaced with a single integer comparison against cached token IDs - the same pattern used for EOS detection.

## Naming Convention

This PR adopts the new **bot/eot/bor/eor** naming convention introduced in this [GenAI PR](https://github.com/microsoft/onnxruntime-genai/pull/2215):

| Abbreviation | Expansion | Purpose |
|---|---|---|
| **bot** | beginning of tool (call) | marks start of tool-call content |
| **eot** | end of tool (call) | marks end of tool-call content |
| **bor** | beginning of reasoning | marks start of reasoning/thinking content |
| **eor** | end of reasoning | marks end of reasoning/thinking content |

Mirrors `bos`/`eos`/`pad` in both naming and access pattern (same Tokenizer class).

## Performance

**Before (per token):** 2 tokenizer stream decodes + 2 `string.find()` + string comparison
**After (per token):** 1 integer comparison + 1 normal-stream decode

The updated path **eliminates the special tokenizer stream from decode** and keeps only the normal stream.
There is no runtime double-decode fallback in `Decode()`.

## Storage (follows EOS pattern)

| Data | Storage | Consumer |
|------|---------|----------|
| `eos_token_ids_` | `GenAIModelInstance` | `Decode()` - filters EOS tokens |
| `tag_info_` (IDs + strings) | `GenAIModelInstance` | `Decode()` - fast-path detection |
| String copies | `ModelInfo.string_properties` | `BuildToolCallContext()` -> `ToolCallStreamAccumulator` |

### `GenAIModelInstance` (`genai_model_instance.h` / `.cc`)
- New `TagInfo` struct with `bot_id`/`eot_id`/`bor_id`/`eor_id` + decoded strings
- `GetTagInfo()` - lazy-cached via `std::call_once` (same pattern as `GetEosTokenIds()`)
  - Calls `tokenizer_->GetBotTokenId()` etc. (4 getter calls, one-time)
  - Decodes each valid ID through `tokenizer_with_special_->Decode(&id, 1)` to get the string (4 decodes, one-time)

### `OnnxChatGenerator::Decode()` (`onnx_chat_generator.cc`)
- Always uses single-stream decode
- Compares token IDs against cached bot/eot/bor/eor IDs
- Returns cached marker strings for tag tokens
- Decodes normal tokens once via the normal tokenizer stream

### `Model::Load()` (`model.cc`)
- After loading the GenAI model, enriches `ModelInfo` string properties from `GetTagInfo()` cached strings
- Only writes if the property isn't already set (catalog metadata takes precedence)
- Does **not** infer `supports_tool_calling` / `supports_reasoning` from tag ID presence

### No changes to `ToolCallContext`, `ToolCallStreamAccumulator`, or `BuildToolCallContext()`
- They continue consuming strings from `ModelInfo` as before - single source of truth

## Dependency

Requires updated `onnxruntime-genai` nuget with `OgaTokenizerGetBotTokenId` / `GetEotTokenId` / `GetBorTokenId` / `GetEorTokenId` APIs.
- GenAI PR: [Add tool-calling and reasoning token IDs to GenAI Tokenizer (bot/eot/bor/eor)](https://github.com/microsoft/onnxruntime-genai/pull/2215)
- Foundry Local uses the newer ORT GenAI nightly containing these APIs.

### Before (per token, every single generated token):

```
Decode Loop:
  token_id = generator.GetNextTokens()[0]

  // Decode through BOTH streams (expensive)
  token_text    = stream_->Decode(token_id)              // normal stream decode
  special_text  = stream_with_special_->Decode(token_id) // special stream decode

  // String operations to detect markers
  if (special_text != token_text):
    is_tool  = special_text.find("tool_call")            // string search
    is_think = special_text.find("think")                // string search
    if (is_tool || is_think): return special_text
  return token_text
```

**Cost per token:** 2 tokenizer stream decodes + 2 `string.find()` + string comparison - on *every* token, even though markers appear maybe 2-4 times per generation.

### After (per token):

```
Model Load (one-time, 4 tokenizer getter calls + 4 Decode calls):
  GenAI:  tokenizer->GetBotTokenId() -> 151657   (from config or fallback vocab lookup)
  FL:     Decode(151657) -> "<tool_call>"           (one tokenizer decode, cached in TagInfo)
  FL:     ModelInfo["tool_call_start"] = "<tool_call>"

Decode Loop:
  token_id = generator.GetNextTokens()[0]

  // Integer comparison (virtually free)
  if (token_id == tag_info.bot_id) { stream_->Decode(token_id); return tag_info.bot_str; }
  if (token_id == tag_info.eot_id) { stream_->Decode(token_id); return tag_info.eot_str; }
  // ... bor/eor ...

  // Normal token - single decode, no special stream needed
  return stream_->Decode(token_id)
```

**Cost per token:** 1 integer comparison + 1 normal decode. No special stream. No string search.

### Estimated performance gain

For a typical 500-token generation:
- **Before:** 500 x (2 decodes + 2 string searches) = 1000 decode calls + 1000 string ops
- **After:** 500 x (1 decode + 1 int compare) = 500 decode calls + 500 trivial comparisons

~**50% reduction in tokenizer decode calls** and **complete elimination of string search operations** in the hot loop. The special tokenizer stream is never even created/maintained for models with tag IDs configured.

## Tag names supported by the GenAI API

- `"tool_call_start"`, `"tool_call_end"` - tool call delimiters
- `"reasoning_start"`, `"reasoning_end"` - chain-of-thought reasoning delimiters

## Testing

Validated with local build and focused integration tests on the updated branch:

- C++ build succeeded (`foundry_local_tests` target)
- Tool-calling integration filter passed (9 tests)
- Response-format web service integration test passed (`WebServiceIntegrationTest.ChatCompletionsWithResponseFormat`)

The hot decode path now runs single-stream decode for all tokens.


sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -117,51 +117,34 @@ ToolCallContext ChatSession::BuildToolCallContext(const Request& request) const
...
-  // Fall back to model info properties if not specified in the request
-  const auto& info = CatalogModel().Info();
-
-  // Check if the model supports tool calling
-  const auto* tool_calling_val = info.GetPropertyInt(FOUNDRY_LOCAL_MODEL_PROP_SUPPORTS_TOOL_CALLING_INT);
-  if (tool_calling_val && *tool_calling_val == 1) {
-    tool_ctx.supports_tool_calling = true;
-  }
-
+  // Fall back to GenAI model API (reads genai_config.json + model-family fallback map)
```
skottmckay: This feels a little clunky to have a bunch of metadata coming from the catalog and some from the genai_config.json but we only merge those when creating a session.

Should we instead when populating the catalog models update the ModelInfo for a cached model to incorporate metadata from genai_config.json at that point? Same process post-download of a new model. That way all the other code doesn't need to know/care about the multiple sources of metadata - the just use the ModelInfo as-is.

sdk_v2/cpp/src/inferencing/generative/genai_model_instance.cc
```diff
@@ -111,6 +111,15 @@ OgaModel& GenAIModelInstance::GetOgaModel() {
   return *oga_model_;
 }
 
+std::string GenAIModelInstance::GetGenerationTag(const char* tag_name) const {
```
skottmckay: nit: `GetTag` would be more generic unless there's an expected use-case where we would need to separate 'generation' tags from others. I assume the tag name could provide any separation required though.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
sayanshaw24: Good call. Refactored in the latest commit. `Model::Load()` now populates `ModelInfo` with tool/reasoning tags from the GenAI model immediately after loading. `BuildToolCallContext()` is unchanged; it just reads from `ModelInfo` as before. Downstream code doesn't need to know about multiple metadata sources anymore.

sdk_v2/cpp/src/inferencing/generative/genai_model_instance.cc
sayanshaw24: agreed, renamed to `GetTag` in both [GenAI ](https://github.com/microsoft/onnxruntime-genai/pull/2215) and FL. just need to update the GenAI version here in FL once the PR is merged and included.

sdk_v2/cpp/src/model.h
```diff
@@ -197,7 +197,7 @@ class Model {
   // cleared by RemoveFromCache(). Its mutation is guarded by state_mutex_; the reader-safety
   // contract is that the path is published before cached_ flips true (and cleared after cached_
   // flips false), so any reader that gates on IsCached() observes a complete path.
-  std::unique_ptr<const ModelInfo> info_;
+  std::unique_ptr<ModelInfo> info_;
```
baijumeswani: This is the part I'd most like to sort out before merging.

Until now `info_` was `const` and never changed after the model was built, so any thread could safely read it through `Info()` without locking. This change makes it writable and `Model::Load()` now writes into `info_->string_properties` after construction.

The problem: other threads can be reading `Info()` at the same time `Load()` is writing to it (for example `DownloadModel(Info())`, catalog listings, or a variant reading its parent's info). Writing to a `std::map` while another thread reads it is a data race and can crash.

`Load()` takes `lifecycle_mutex_`, but the readers don't take any lock, so that lock doesn't actually protect them.

A couple of ways to fix it:
1. Fill in these tag strings once, before the model info is first published/visible to other threads, so nobody is reading it while you write - then it can stay effectively read-only.
2. Or, if it really must change after load, make every reader and writer of `string_properties` go through the same lock.

Option 1 is usually the cleaner fit for how this class is designed.

sdk_v2/cpp/src/model.cc
```diff
@@ -513,6 +513,33 @@ void Model::Load(ExecutionProvider ep) {
   if (result.status == ModelLoadManager::LoadStatus::kModelNotFound) {
     FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL, "model not found at path: " + local_path_);
   }
+
+  // Enrich ModelInfo with metadata from the loaded GenAI model (genai_config.json + fallback map).
+  // This makes tool/reasoning tags available via ModelInfo regardless of whether they came from
+  // catalog metadata, so downstream code doesn't need to know about multiple metadata sources.
+  // Runs on both kSuccess (fresh load) and kModelAlreadyLoaded (model already in memory from
+  // another code path) to ensure this Model's info_ is always enriched.
+  if (result.model && (result.status == ModelLoadManager::LoadStatus::kSuccess ||
```
baijumeswani: This is where the write happens (see my note on `info_` in model.h). This block runs under `lifecycle_mutex_`, but the threads that read `Info()` elsewhere don't hold that lock, so they aren't protected from this write. Enriching before the info becomes visible to other threads, or locking all readers/writers of `string_properties` with one mutex, would make this safe.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_generator.cc
```diff
@@ -85,31 +81,30 @@ std::string OnnxChatGenerator::Decode() {
...
 
-    if (!is_eos && (is_tool_call_token || is_think_token)) {
-      return special_str;
-    }
+  if (tag_info.bot_id.has_value() && token_id == *tag_info.bot_id) {
+    stream_->Decode(token_id);
+    return tag_info.bot_str;
+  }
+  if (tag_info.eot_id.has_value() && token_id == *tag_info.eot_id) {
+    stream_->Decode(token_id);
```
baijumeswani, line 93: Small correctness-vs-description note (the code itself is fine).

The PR description says the fast path "eliminates the special tokenizer stream entirely" and cuts decode calls by about half. But this fast path still calls both `stream_->Decode()` and `stream_with_special_->Decode()` - which is correct, because the streams are stateful and have to stay in sync. So the number of decode calls is actually unchanged; what you save is the two `string.find()` calls and the string comparison per token.

That's still a real (if smaller) win. Could you tweak the PR description so it matches what the code does? Otherwise a reader might expect a bigger speedup than is actually delivered.

sdk_v2/cpp/src/model.cc
```diff
@@ -110,7 +110,7 @@ Model::Model(ModelInfo info,
              DownloadManager& download_manager,
              ModelLoadManager& model_load_manager,
              bool external_registration)
-    : info_(std::make_unique<const ModelInfo>(std::move(info))),
+  : info_(std::make_unique<ModelInfo>(std::move(info))),
```
baijumeswani: Tiny formatting nit: the `:` here is indented 2 spaces, but the members just below (`cached_`, `local_path_`, ...) are indented 6. Lining this up with them keeps the initializer list tidy.

baijumeswani (commented):
Thanks for this - the overall approach is clean and I like that it mirrors the existing EOS token pattern with `call_once` caching, keeps the slow path for backward compatibility, and deliberately avoids guessing the tool/reasoning support flags from the family-wide fallback map. Nice.

I left a few inline comments. The main one worth resolving before merge is the thread-safety of making `ModelInfo` writable and updating it inside `Load()`. The rest are smaller: the performance description doesn't quite match what the code does, a couple of description/code mismatches, and a small formatting nit.

A few non-blocking notes:
- **Dependency:** this needs the newer onnxruntime-genai package that has `GetBotTokenId` / `GetEotTokenId` / `GetBorTokenId` / `GetEorTokenId`. `deps_versions.json` isn't bumped in this PR, so CI won't build until that lands. Worth calling out in the PR so it isn't merged early.
- **Tests:** none here (understandable, since it needs the new package). Once the package is available, a small test that a model with these tokens fills in the tag strings, and a model without them still works via the old path, would give good coverage.
- **Description mismatch:** the "Changes -> Model::Load()" section says it "Infers supports_tool_calling / supports_reasoning from tag ID presence". The code intentionally does NOT do this (and the code comment explains why - good call). Please update the description so it matches.

baijumeswani (dismissed):

sdk_v2/cpp/src/model.cc
baijumeswani: Will address in a follow up.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_generator.cc
baijumeswani: Will address in a follow up.

sdk_v2/cpp/src/model.cc
baijumeswani: Will address in a follow up.

sdk_v2/cpp/src/model.h
baijumeswani: Will address in a follow up.

sayanshaw24:
> Thanks for this - the overall approach is clean and I like that it mirrors the existing EOS token pattern with `call_once` caching, keeps the slow path for backward compatibility, and deliberately avoids guessing the tool/reasoning support flags from the family-wide fallback map. Nice.
> 
> I left a few inline comments. The main one worth resolving before merge is the thread-safety of making `ModelInfo` writable and updating it inside `Load()`. The rest are smaller: the performance description doesn't quite match what the code does, a couple of description/code mismatches, and a small formatting nit.
> 
> A few non-blocking notes:
> 
> * **Dependency:** this needs the newer onnxruntime-genai package that has `GetBotTokenId` / `GetEotTokenId` / `GetBorTokenId` / `GetEorTokenId`. `deps_versions.json` isn't bumped in this PR, so CI won't build until that lands. Worth calling out in the PR so it isn't merged early.
> * **Tests:** none here (understandable, since it needs the new package). Once the package is available, a small test that a model with these tokens fills in the tag strings, and a model without them still works via the old path, would give good coverage.
> * **Description mismatch:** the "Changes -> Model::Load()" section says it "Infers supports_tool_calling / supports_reasoning from tag ID presence". The code intentionally does NOT do this (and the code comment explains why - good call). Please update the description so it matches.

Thanks for the detailed review. Addressed the comments in the latest update:

- Thread safety: Restored ModelInfo immutability and removed the post-load mutation of info_->string_properties. Dynamically discovered GenAI marker strings are now read from the loaded model’s cached TagInfo and merged into the request-scoped ToolCallContext by ChatSession::BuildToolCallContext(). This avoids publishing mutable metadata and removes the data-race risk.
- Decode path: The implementation now uses a single tokenizer stream for every generated token. Tag IDs are compared directly, and matching tokens return their cached marker strings after advancing the normal stream. The special-token stream and double-decode path were removed.
- Capability flags: Support flags remain catalog-authored; tag-token presence is not used to infer tool-calling or reasoning support because the GenAI fallback map can apply to model families more broadly than the actual capability.
- PR description: Updated to match the immutable metadata ownership and single-stream decode behavior.
- Validation: The Windows C++ build passed, the focused tool-calling suite passed 9/9, and WebServiceIntegrationTest.ChatCompletionsWithResponseFormat passed.

The changes are pushed in commit d55c5ee2 (Avoid mutating published model metadata).

sdk_v2/cpp/src/model.h
sayanshaw24: Addressed in d55c5ee2. ModelInfo is immutable again, and Load() no longer writes to info_->string_properties. ChatSession::BuildToolCallContext() now falls back to the loaded GenAI model’s cached TagInfo for dynamically discovered marker strings, keeping the dynamic data request-scoped and avoiding concurrent reads/writes to published metadata.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_generator.cc
sayanshaw24: Noticed this as well, addressed. The latest implementation removes stream_with_special_ entirely. Decode() compares the token ID against the cached bot/eot/bor/eor IDs and advances only the normal tokenizer stream once per token. Matching tags return their cached marker string; normal tokens return the normal-stream decode result. The PR description now reflects this single-stream behavior.

baijumeswani (approved):

sdk_v2/cpp/src/model.cc
sayanshaw24: Addressed in d55c5ee2. Removed the post-load mutation of info_->string_properties and restored immutable ModelInfo ownership. BuildToolCallContext() now falls back to the loaded GenAI model’s cached TagInfo for dynamically discovered marker strings, avoiding concurrent reads/writes to published metadata.

sdk_v2/cpp/src/model.cc
sayanshaw24: Fixed!

PR #845, opened 2026-06-29
Mark OpenAI direct clients as deprecated in V2 SDK

skottmckay:
Users should move to the new Session API for direct requests.
The web service OpenAI endpoints are unchanged.


sdk_v2/cs/src/Detail/Model.cs
```diff
@@ -115,6 +115,7 @@ await Utils.CallWithExceptionHandlingAsync(
             $"Error removing model {Id} from cache", _logger, ct).ConfigureAwait(false);
     }
 
+    [System.Obsolete("Use new ChatSession(model) instead.", error: false)]
```
prathikr, line 118: Should we add a timeframe for when these will be ultimately removed as part of the deprecation notice? I fear people will disregard the note on its own unless we present a timeframe

baijumeswani (approved):

PR #848, opened 2026-06-29
fetches test models from azure storage blob

prathikr:


sdk_v2/DEVELOPMENT.md
```diff
@@ -87,6 +87,24 @@ pwsh ./build_and_test_all.ps1 -ContinueOnError
 
 See `pwsh ./build_and_test_all.ps1 -?` for the full parameter list.
 
+## Model cache for tests
+
+Model-dependent sdk_v2 tests require `FOUNDRY_TEST_DATA_DIR` to point to a
+local model cache directory.
+
+In CI, sdk_v2 pipelines download model data from blob storage into this path.
+For local runs, set it explicitly before invoking test commands.
```
baijumeswani: How should developers download these models for testing?
Should we instead just download the model from FL SDK for local runs and use the cache for CI runs?
Or perhaps can we share instructions that developers should do to run these tests locally?

skottmckay: And instructions for how to update the test models available.

prathikr: Yes, that’s the plan. I will add more docs.

sdk_v2/DEVELOPMENT.md
```diff
@@ -87,6 +87,70 @@ pwsh ./build_and_test_all.ps1 -ContinueOnError
...
+foundry model download qwen3-embedding-0.6b
+foundry model download nemotron-speech-streaming-en-0.6b
+```
+
+Then set env var `FOUNDRY_TEST_DATA_DIR` to the cache path you want tests to use.
+
+Example:
+
+```powershell
+$env:FOUNDRY_TEST_DATA_DIR = 'C:\path\to\model-cache'
```
baijumeswani: Since we are recommending that users use the cli to download the test models, let's also add the step that copies the models fl cli downloaded into a structure that foundry local sdk tests accepts?

```
mkdir test-data-shared
$env:FOUNDRY_TEST_DATA_DIR = C:\path\to\test-data-shared
$FOUNDRY_CLI_CACHE_DIR = "$env:USERPROFILE\.foundry\cache\models\Microsoft"
Copy-Item -Path $FOUNDRY_CLI_CACHE_DIR\qwen2.5-0.5b-instruct-generic-cpu-4\v4 -Destination $env:FOUNDRY_TEST_DATA_DIR\qwen2.5-0.5b --Recurse
```

Alternatively, we can build a small exe as part of the build_and_test_all.ps1 that auto downloads the necessary models and places them in a folder structure that the tests expect.

baijumeswani: Or a python utility might be simpler.

prathikr: Not sure I follow. In both the old and new setup, the expected cache structure is `$FOUNDRY_CLI_CACHE_DIR\qwen2.5-0.5b-instruct-generic-cpu-4\v4`. Neither setup uses `$env:FOUNDRY_TEST_DATA_DIR\qwen2.5-0.5b`. In fetch-test-data-from-blob.yml we build the model filepath from the alias + version to match the former filepath, not the latter.

baijumeswani (approved):

PR #851, opened 2026-06-30
Consolidate test temp helpers and move file I/O + lock platform code into src/platform (#793 follow-up)

bmehta001:
Follow-up to #793 addressing two review nits from @skottmckay that were deferred to a separate PR.

Four unit tests (`download_test`, `cross_process_file_lock_test`, `blob_download_state_test`, `file_writer_test`) each defined a near-identical local `TempDir`/`TempPath` RAII class, and `test_helpers.h` carried the `CurrentPid`/`MakeUniqueTempPath` helpers. These now live in a single `test/utils/temp_dir.h` (`fl::test` namespace); the tests use it via the existing `test_helpers.h` umbrella include.

Both `file_writer.cc` and `cross_process_file_lock.cc` carried Win32/POSIX branches behind in-file `#ifdef _WIN32`. Following the existing `src/platform/{windows,posix}/path.cc` pattern:

- New `src/platform/file_io.h` declares `fl::platform` `OpenWritableFile` / `WriteFileAt` / `CloseFile` primitives over an integer handle, implemented per-OS in `src/platform/{windows,posix}/file_io.cc`. `FileWriter` stores an `intptr_t` handle and delegates — no more `#ifdef` member or branches.
- `CrossProcessFileLock` is split: the platform-specific `State`, its releasing destructor, `FormatProcessInfo`, the ctor/dtor, and `TryAcquireForDirectory` move **verbatim** into `src/platform/{windows,posix}/cross_process_file_lock.cc`; the neutral `src/download/cross_process_file_lock.cc` keeps only the cross-platform `WaitForDirectoryLock` orchestration.
- The four new sources are wired into `FOUNDRY_LOCAL_PLATFORM_SOURCES`.

## Validation
- **No behavior change** — pure refactor.
- Windows `RelWithDebInfo` build green; 83 download / lock / file-writer / state unit tests pass (incl. all 9 `CrossProcessFileLockTest`).
- POSIX sources syntax-clean under `g++ -std=c++20 -Wall -Wextra`; full POSIX link validated by CI.


sdk_v2/cpp/src/platform/cross_process_file_lock.cc
skottmckay, line 1: Should we put this file and the header in src/platform? Feels a little confusing to have parts of the implementation here and parts in src/platform. This would match the ORT setup where the generic part of a class with platform specific implementation is in the `platform` directory.

sdk_v2/cpp/test/utils/temp_dir.h
```diff
@@ -0,0 +1,101 @@
...
+        throw std::runtime_error("TempDir: failed to create '" + candidate.string() + "': " +
+                                 ec.message());
+      }
+      // candidate already existed — try the next name.
+    }
+  }
+
+  ~TempDir() {
+    std::error_code ec;
+    std::filesystem::remove_all(path_, ec);
```
skottmckay: nit: `remove_all` can throw so should add a try/catch and maybe a stderr output of the exception reason.

sdk_v2/cpp/test/utils/temp_dir.h
```diff
@@ -0,0 +1,101 @@
...
+  const std::filesystem::path& path() const { return path_; }
+  std::string string() const { return path_.string(); }
+
+ private:
+  std::filesystem::path path_;
+};
+
+/// RAII unique temporary file path. The path is not created here — callers create the file — and
+/// it is removed on destruction so a flaky test never leaks files into the system temp dir.
+class TempPath {
```
skottmckay: nit: can we combine these two classes given they're essentially doing the same thing? 

maybe call it TempPath (as a dir is a path), private ctor with a flag for dir/path, static CreateTempDir/CreateTempFile methods returning a TempPath instance.

sdk_v2/cpp/src/platform/cross_process_file_lock.cc
bmehta001: Yes, moved

sdk_v2/cpp/test/utils/temp_dir.h
bmehta001: Fixed

sdk_v2/cpp/test/utils/temp_dir.h
bmehta001: Yes, done

skottmckay (approved):

PR #860, opened 2026-07-01
Improve Linux/macOS debug symbol setup

skottmckay:
Separate/create symbols file for linux and macos to replicate Windows setup where the release library binary is small but the symbols are available for debugging issues.

**Linux** — three-step GNU toolchain sequence:
1. `objcopy --only-keep-debug` → `libfoundry_local.so.dbg` (DWARF only)
2. `strip --strip-debug` → removes DWARF sections from the `.so`, keeps the exported symbol table (needed for `dlopen`)
3. `objcopy --add-gnu-debuglink=libfoundry_local.so.dbg` → embeds the filename + CRC into a `.gnu_debuglink` section. GDB/LLDB read this and auto-discover the `.dbg` file when it's in the same directory. Zero developer friction.

**macOS** — two-step Apple toolchain sequence (order matters):
1. `dsymutil` first — harvests DWARF from all `.o` files (still present at POST_BUILD time) into `libfoundry_local.dylib.dSYM/` — a self-contained relocatable bundle
2. `strip -S` second — removes the now-redundant N_OSO stab entries. LLDB/Xcode find the `.dSYM` automatically by UUID when it sits alongside the `.dylib`.


baijumeswani (dismissed):

prathikr:
Do we want to include these symbols/debug files in the tgz files of the cpp sdk? Will need some changes under `.pipelines\v2\templates\steps-pack-cpp-sdk.yml` to add symbols into the tgz files.

skottmckay:
> Do we want to include these symbols/debug files in the tgz files of the cpp sdk? Will need some changes under `.pipelines\v2\templates\steps-pack-cpp-sdk.yml` to add symbols into the tgz files.

I think so. We don't want the symbols in the packages, but we do want them to be available for direct download. We also need to publish the pdb to the windows symbol server I believe. 

Can you add the required changes to this branch or point me to what needs updating?

skottmckay:
> > Do we want to include these symbols/debug files in the tgz files of the cpp sdk? Will need some changes under `.pipelines\v2\templates\steps-pack-cpp-sdk.yml` to add symbols into the tgz files.
> 
> I think so. We don't want the symbols in the packages, but we do want them to be available for direct download. We also need to publish the pdb to the windows symbol server I believe.
> 
> Can you add the required changes to this branch or point me to what needs updating?

Added changes to include in tgz

baijumeswani (approved):

PR #866, opened 2026-07-08
fix(cpp): honor SSL_CERT_FILE via CURLOPT_CAINFO for TLS on Android

sheetalarkadam:
## Problem

TLS verification on Android failed for every HTTPS request to Azure, with two different misleading errors:

- `self-signed certificate in certificate chain` — catalog requests
- `unable to get local issuer certificate` — blob downloads

Both have the same root cause: **the trust store was empty**, because neither of the two mechanisms that should have supplied it works in this build.

1. **The compiled-in default is wrong.** OpenSSL's `--openssldir` is baked in at build time as a string constant. vcpkg cross-compiles for Android on the host, so the baked path does not exist on a device (Android's store lives at `/system/etc/security/cacerts/` as hash-named `.0` files).

2. **`SSL_CERT_FILE` never gets read.** It is an OpenSSL-level fallback consulted only by `SSL_CTX_set_default_verify_paths()` — the "no explicit CA configured" path. libcurl *always* configures an explicit CA (its own build-time default, via `SSL_CTX_load_verify_locations()`), so that fallback never runs and the environment variable is silently ignored.

The `self-signed certificate in certificate chain` wording is a red herring: with an empty store the chain walks up to a root, and every root *is* self-signed — with nothing to validate it against, that is exactly what OpenSSL reports.

Isolated on-device: passing the device's own trust store via `SSL_CERT_FILE` failed, while the **same bundle** passed explicitly as a CA file (`openssl s_client -CAfile`, equivalent to `CURLOPT_CAINFO`) verified `ai.azure.com` with `Verify return code: 0 (ok)`. Same bytes, same device, same endpoint — only the delivery mechanism differed.

## Fix

Keep the familiar `SSL_CERT_FILE` interface, but stop relying on OpenSSL to read it. A shared `http::CaBundleFile()` helper reads the variable, and each libcurl transport receives it explicitly as `CurlTransportOptions.CAInfo` (`CURLOPT_CAINFO`), which libcurl always honors.

The third is the non-obvious one: **the Azure Storage SDK builds its own libcurl transport internally** rather than reusing ours, so fixing the first two left blob downloads still failing — with the different `unable to get local issuer certificate` message. `MakeBlobClientOptions()` injects a preconfigured `CurlTransport` into `options.Transport.Transport`.

Every site is guarded by `#if !defined(FOUNDRY_LOCAL_USE_WINHTTP_TRANSPORT)`, so **desktop Windows is unaffected** — it uses WinHTTP and the OS trust store.

Callers still set `SSL_CERT_FILE` exactly as before; its effect now comes from the Core forwarding it to `CAInfo`, not from OpenSSL auto-reading the environment.

## Validation

- Verified on-device: catalog browse + model download over HTTPS to Azure now succeed.

### DIFF
--- a/sdk_v2/cpp/src/download/blob_downloader.cc
+++ b/sdk_v2/cpp/src/download/blob_downloader.cc
@@ -29,6 +29,8 @@
 // does not honor SSL_CERT_FILE. On non-Windows builds we install a CurlTransport preconfigured with
 // CAInfo (see MakeBlobClientOptions below). Desktop Windows uses the default WinHTTP transport.
 #if !defined(FOUNDRY_LOCAL_USE_WINHTTP_TRANSPORT)
+#include "http/curl_transport.h"
+
 #include <azure/core/http/curl_transport.hpp>
 #endif
 
@@ -50,10 +52,11 @@ constexpr size_t kStreamingBufferBytes = 64 * 1024;
 Azure::Storage::Blobs::BlobClientOptions MakeBlobClientOptions() {
   Azure::Storage::Blobs::BlobClientOptions options;
 #if !defined(FOUNDRY_LOCAL_USE_WINHTTP_TRANSPORT)
-  if (std::string ca_bundle = fl::http::CaBundleFile(); !ca_bundle.empty()) {
-    Azure::Core::Http::CurlTransportOptions curl_opts;
-    curl_opts.CAInfo = std::move(ca_bundle);
-    options.Transport.Transport = std::make_shared<Azure::Core::Http::CurlTransport>(curl_opts);
+  // Only override the Storage SDK's default transport when a CA bundle is configured; the options are
+  // cached and shared with our other curl transports (see http/curl_transport.h).
+  if (!fl::http::CaBundleFile().empty()) {
+    options.Transport.Transport =
+        std::make_shared<Azure::Core::Http::CurlTransport>(fl::http::CachedCurlTransportOptions());
   }
 #endif
   return options;
--- a/sdk_v2/cpp/src/download/blob_downloader.cc
+++ b/sdk_v2/cpp/src/download/blob_downloader.cc
@@ -52,11 +52,11 @@ constexpr size_t kStreamingBufferBytes = 64 * 1024;
 Azure::Storage::Blobs::BlobClientOptions MakeBlobClientOptions() {
   Azure::Storage::Blobs::BlobClientOptions options;
 #if !defined(FOUNDRY_LOCAL_USE_WINHTTP_TRANSPORT)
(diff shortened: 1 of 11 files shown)


skottmckay:
Suggestions in #955

sheetalarkadam:
Merged @skottmckay's suggestions from #955 into this branch (`f1e18314`, `5383d24e`, `5bc6af8a`) — applied as-is, no modifications.

**Device-validated.** Built arm64-v8a and confirmed the fix works on a physical Android device

**On the process-lifetime cache:** I raised a concern that a failed CA export would memoize an empty path for the process lifetime, defeating a retry. Having dug into it, I'm dropping that — Scott is right that it's rare, and it doesn't warrant the change:

- On Android the export failing is unlikely, and the exporter already logs at ERROR when it does.
- Process death is the common case and it *self-heals*: the static, the environment variable, and the loaded library all die together, so a bad cached value can never outlive the process that produced it.
- Across an AIDL disconnect where the process survives, the cached value stays correct — the cache holds a **path**, not certificate bytes, and that path is a pure function of package + UID, so it is identical before and after.

Keeping the cache as written. It also lets the function keep returning `const std::string&` safely, since the static is immutable.

skottmckay:
Should we limit the new code that is reading the env var and setting up options to only ANDROID builds?

sheetalarkadam:
> Should we limit the new code that is reading the env var and setting up options to only ANDROID builds?

Do you mean `SSL_CERT_FILE` and the `CAInfo` population? It's no-op on Linux/macOS unless someone sets `SSL_CERT_FILE` in which case we should use it?

sdk_v2/cpp/src/download/blob_downloader.cc
```diff
@@ -32,6 +43,25 @@ namespace {
...
 
+/// Builds BlobClientOptions with the CA bundle wired into the transport. The Azure Storage SDK
+/// constructs its own libcurl transport internally, which does not consult SSL_CERT_FILE, so blob
+/// downloads would otherwise fail TLS verification on Android with "unable to get local issuer
+/// certificate". On non-Windows builds we install a CurlTransport preconfigured with CAInfo so
+/// verification uses the caller-provided trust store; on desktop Windows the default WinHTTP
+/// transport uses the OS store and is left untouched.
+Azure::Storage::Blobs::BlobClientOptions MakeBlobClientOptions() {
+  Azure::Storage::Blobs::BlobClientOptions options;
+#if !defined(FOUNDRY_LOCAL_USE_WINHTTP_TRANSPORT)
```
skottmckay: Can this be `#if defined(ANDROID)` in various places to narrow it down?

skottmckay (approved):

PR #885, opened 2026-07-15
support file-based ASR for nemotron streaming model

rui-ren:
## Summary

This PR completes file-based audio transcription support for Nemotron streaming ASR models in  sdk_v2/cpp  by finishing the C++  AudioSession  implementation and aligning behavior with the existing OpenAI-JSON transcription flow.

For more details, please refer to https://github.com/github/copilot-cli/issues/4024#issuecomment-4951658941

## What changed

• Implemented/finished Nemotron file transcription path in:
```
 sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc 
 sdk_v2/cpp/src/inferencing/generative/audio/audio_session.h 
```
• Added WAV file parsing for file-based transcription input:
• RIFF/WAVE validation
•  fmt / data  chunk handling with bounds checks
• 16kHz enforcement
• PCM16 and float32 decoding
• Added Nemotron-specific language runtime option mapping ( lang_id ) and temperature handling.
• Fixed decode flow to keep a single generator across incremental audio processing (100ms chunks + flush), which resolves truncated transcription output for file-based runs.

## Behavior

• OpenAI-JSON file transcription for Nemotron models now returns full transcript text from WAV input in SDK v2.
• Existing streaming and non-Nemotron paths remain unchanged.

## End-to-end scenario covered

• Model alias:  nemotron-3.5-asr-streaming-0.6b 
• Input:  sample-speech-1m-16k.wav 
• Path exercised: JS SDK v2  AudioClient.transcribe(...)  → C++ SDK v2 audio transcription pipeline

### DIFF
--- a/sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
@@ -579,7 +579,7 @@ void AudioSession::DecodeNemotronTokens(OgaGenerator& generator, OgaTokenizerStr
                                         int& completion_tokens) const {
   const bool is_streaming = (streaming_callback != nullptr);
 
-  while (!generator.IsDone() && !original_request.canceled) {
+  while (!generator.IsDone() && !generator.IsSessionTerminated() && !original_request.canceled) {
     generator.GenerateNextToken();
     auto next_tokens = generator.GetNextTokens();
     if (next_tokens.empty()) {
@@ -667,8 +667,10 @@ void AudioSession::ProcessNemotronFileTranscription(const AudioTranscriptionRequ
     RunNemotronDecodePass(processor->Process(samples.data() + offset, count), *generator, *tokenizer_stream, text,
                           streaming_callback, response_id, original_request, completion_tokens);
   }
-  RunNemotronDecodePass(processor->Flush(), *generator, *tokenizer_stream, text, streaming_callback, response_id,
-                        original_request, completion_tokens);
+  if (!original_request.canceled) {
+    RunNemotronDecodePass(processor->Flush(), *generator, *tokenizer_stream, text, streaming_callback, response_id,
+                          original_request, completion_tokens);
+  }
 
   response.finish_reason = original_request.canceled ? FOUNDRY_LOCAL_FINISH_NONE : FOUNDRY_LOCAL_FINISH_STOP;
   // Nemotron file-transcription path feeds audio tensors directly and does not expose prompt token accounting.
--- a/sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
@@ -19,9 +19,15 @@
 #include "utils.h"
 
 #include <filesystem>
+#include <algorithm>
+#include <cctype>
+#include <cstdint>
+#include <cstring>
+#include <fstream>
 #include <fmt/format.h>
 #include <nlohmann/json.hpp>
 #include <ort_genai.h>
+#include <unordered_map>
 
 namespace fl {
 
@@ -71,41 +77,45 @@ std::string JoinTokens(const std::vector<std::string>& token_texts) {
   return out;
 }
 
+const std::unordered_map<std::string, std::string>& NemotronLanguageIdMap() {
+  static const std::unordered_map<std::string, std::string> kMap = {
+      {"en", "0"},       {"en-us", "0"},    {"en-gb", "1"},    {"es-es", "2"},   {"es", "3"},
+      {"es-us", "3"},    {"zh-cn", "4"},    {"hi", "6"},       {"hi-in", "6"},   {"ar", "7"},
+      {"ar-ar", "7"},    {"fr", "8"},       {"fr-fr", "8"},    {"fr-ca", "100"}, {"de", "9"},
+      {"de-de", "9"},    {"ja", "10"},      {"ja-jp", "10"},   {"ru", "11"},     {"ru-ru", "11"},
+      {"pt-br", "12"},   {"pt", "13"},      {"pt-pt", "13"},   {"ko", "14"},     {"ko-kr", "14"},
+      {"it", "15"},      {"it-it", "15"},   {"nl", "16"},      {"nl-nl", "16"},  {"pl", "17"},
+      {"pl-pl", "17"},   {"tr", "18"},      {"tr-tr", "18"},   {"uk", "19"},     {"uk-ua", "19"},
+      {"ro", "20"},      {"ro-ro", "20"},   {"el", "21"},      {"el-gr", "21"},  {"cs", "22"},
+      {"cs-cz", "22"},   {"hu", "23"},      {"hu-hu", "23"},   {"sv", "24"},     {"sv-se", "24"},
+      {"da", "25"},      {"da-dk", "25"},   {"fi", "26"},      {"fi-fi", "26"},  {"sk", "28"},
+      {"sk-sk", "28"},   {"hr", "29"},      {"hr-hr", "29"},   {"bg", "30"},     {"bg-bg", "30"},
+      {"lt", "31"},      {"lt-lt", "31"},   {"th", "32"},      {"th-th", "32"},  {"vi", "33"},
+      {"vi-vn", "33"},   {"et", "60"},      {"et-ee", "60"},   {"lv", "61"},     {"lv-lv", "61"},
+      {"sl", "62"},      {"sl-si", "62"},   {"he", "64"},      {"he-il", "64"},  {"auto", "101"},
+      {"mt", "102"},     {"mt-mt", "102"},  {"nb", "103"},     {"nb-no", "103"}, {"nn", "104"},
+      {"nn-no", "104"},
+  };
+  return kMap;
+}
+
+std::string ToLowerAscii(std::string s) {
+  std::transform(s.begin(), s.end(), s.begin(),
+                 [](unsigned char c) { return static_cast<char>(std::tolower(c)); });
+  return s;
+}
+
+bool IsLanguageToken(const std::string& token) {
+  size_t start = token.find_first_not_of(" \t\r\n");
+  if (start == std::string::npos) {
+    return false;
+  }
+
+  size_t end = token.find_last_not_of(" \t\r\n");
+  return end >= start && token[start] == '<' && token[end] == '>';
+}
+
 }  // namespace
-
-namespace {
-  const std::unordered_map<std::string, std::string> NemotronLangIdMap = {
-      ["en"] = "0", ["en-US"] = "0", ["en-GB"] = "1",
-      ["es-ES"] = "2", ["es"] = "3", ["es-US"] = "3",
-      ["zh-CN"] = "4",
-      ["hi"] = "6", ["hi-IN"] = "6", ["ar"] = "7", ["ar-AR"] = "7",
-      ["fr"] = "8", ["fr-FR"] = "8", ["fr-CA"] = "100",
-      ["de"] = "9", ["de-DE"] = "9",
-      ["ja"] = "10", ["ja-JP"] = "10", ["ru"] = "11", ["ru-RU"] = "11",
-      ["pt-BR"] = "12", ["pt"] = "13", ["pt-PT"] = "13",
-      ["ko"] = "14", ["ko-KR"] = "14", ["it"] = "15", ["it-IT"] = "15",
-      ["nl"] = "16", ["nl-NL"] = "16", ["pl"] = "17", ["pl-PL"] = "17",
-      ["tr"] = "18", ["tr-TR"] = "18", ["uk"] = "19", ["uk-UA"] = "19",
-      ["ro"] = "20", ["ro-RO"] = "20",
-      ["el"] = "21", ["el-GR"] = "21", ["cs"] = "22", ["cs-CZ"] = "22",
-      ["hu"] = "23", ["hu-HU"] = "23", ["sv"] = "24", ["sv-SE"] = "24",
-      ["da"] = "25", ["da-DK"] = "25",
-      ["fi"] = "26", ["fi-FI"] = "26", ["sk"] = "28", ["sk-SK"] = "28",
-      ["hr"] = "29", ["hr-HR"] = "29", ["bg"] = "30", ["bg-BG"] = "30",
-      ["lt"] = "31", ["lt-LT"] = "31", ["th"] = "32", ["th-TH"] = "32",
-      ["vi"] = "33", ["vi-VN"] = "33",
-      ["et"] = "60", ["et-EE"] = "60", ["lv"] = "61", ["lv-LV"] = "61",
-      ["sl"] = "62", ["sl-SI"] = "62", ["he"] = "64", ["he-IL"] = "64",
-      ["auto"] = "101",
-      ["mt"] = "102", ["mt-MT"] = "102", ["nb"] = "103", ["nb-NO"] = "103",
-      ["nn"] = "104", ["nn-NO"] = "104",
-    };
-
-    std::string ToLowerAscii(std::string s) {
-      std::transform(s.begin(), s.end(), s.begin(), [](unsigned char c) { return std::tolower(c); });
-      return s;
-    }
-  }  // namespace fl
 
 
 AudioSession::AudioSession(const fl::Model& catalog_model, GenAIModelInstance& model,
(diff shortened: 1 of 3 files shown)


nenad1002:
Why? Parakeet can execute faster and more accurate file-based transcription. I wouldn't want customers to rely on streaming models for non-streaming scenarios, and we're encouraging them with this.

rui-ren:
> Why? Parakeet can execute faster and more accurate file-based transcription. I wouldn't want customers to rely on streaming models for non-streaming scenarios, and we're encouraging them with this.

This is not a way to encourage them to do `file-base` transcription with `RNNT streaming` model, as you can see here, 

https://github.com/github/copilot-cli/issues/4024#issuecomment-4951658941

they are using streaming model, however they do file-based transcription and get error raised. IMO, we better to cover this edge case for them, instead of switch models?

What do you think? @sylvanc

nenad1002:
> > Why? Parakeet can execute faster and more accurate file-based transcription. I wouldn't want customers to rely on streaming models for non-streaming scenarios, and we're encouraging them with this.
> 
> This is not a way to encourage them to do `file-base` transcription with `RNNT streaming` model, as you can see here,
> 
> [github/copilot-cli#4024 (comment)](https://github.com/github/copilot-cli/issues/4024#issuecomment-4951658941)
> 
> they are using streaming model, however they do file-based transcription and get error raised. IMO, we better to cover this edge case for them, instead of switch models?
> 
> What do you think? @sylvanc



> > Why? Parakeet can execute faster and more accurate file-based transcription. I wouldn't want customers to rely on streaming models for non-streaming scenarios, and we're encouraging them with this.
> 
> This is not a way to encourage them to do `file-base` transcription with `RNNT streaming` model, as you can see here,
> 
> [github/copilot-cli#4024 (comment)](https://github.com/github/copilot-cli/issues/4024#issuecomment-4951658941)
> 
> they are using streaming model, however they do file-based transcription and get error raised. IMO, we better to cover this edge case for them, instead of switch models?
> 
> What do you think? @sylvanc

Hmm, I haven't realized they're using file-based transcription. Ok, but there are better models for this going forward.

rui-ren:
@skottmckay can you please review this PR ? Thanks.

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -502,4 +554,278 @@ void AudioSession::ProcessAudioTranscriptionJson(const std::string& request_json
...
+
+  auto streaming_callback = CreateCallbackHandler(original_request);
+  const bool is_streaming = (streaming_callback != nullptr);
+  std::string response_id = ResponseConverter::GenerateId("audio");
+
+  std::string text;
+  text.reserve(512);
+  int completion_tokens = 0;
+
+  auto decode_all_tokens = [&]() {
```
kunal-vaishnavi: Can we put this in a named helper function rather than use a lambda function?

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -502,4 +554,278 @@ void AudioSession::ProcessAudioTranscriptionJson(const std::string& request_json
...
+        AudioTranscriptionResponse chunk;
+        chunk.id = response_id;
+        chunk.text = token;
+        streaming_callback->PushItem(std::make_unique<TextItem>(nlohmann::json(chunk).dump(),
+                                                                FOUNDRY_LOCAL_TEXT_ITEM_TYPE_OPENAI_JSON));
+      }
+    }
+  };
+
+  auto run_one_pass = [&](std::unique_ptr<OgaNamedTensors> tensors) {
```
kunal-vaishnavi: Same comment here

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -422,6 +467,13 @@ void AudioSession::ProcessAudioTranscriptionJson(const std::string& request_json
     FL_LOG_AND_THROW(logger_, FOUNDRY_LOCAL_ERROR_INVALID_USAGE, fmt::format("Audio file not found: '{}'", req.filename));
   }
 
+  // If the model is a Nemotron speech model, process the transcription using the Nemotron-specific method.
+  // e.g. RNNT models require Nemotron-specific transcription handling.
+  if (IsNemotronSpeechModel()) {
```
kunal-vaishnavi, line 475: I presume that the v2 SDK already supported Nemotron when it was merged into main. Why do we now need an early return to support it?

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -502,4 +554,278 @@ void AudioSession::ProcessAudioTranscriptionJson(const std::string& request_json
...
+  }
+
+  auto samples = LoadPcmWavAsFloatSamples(req.filename);
+  auto& oga_model = Model().GetOgaModel();
+  auto processor = OgaStreamingProcessor::Create(oga_model);
+  auto tokenizer = OgaTokenizer::Create(oga_model);
+  auto tokenizer_stream = OgaTokenizerStream::Create(*tokenizer);
+  auto generator_params = OgaGeneratorParams::Create(oga_model);
+  if (temperature.has_value()) {
+    generator_params->SetSearchOption("temperature", *temperature);
```
kunal-vaishnavi, line 657: Can we simplify this `if-else` branching to the following?

```cpp
generator_params->SetSearchOption("temperature", temperature.value_or(0.0f));
```

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -502,4 +554,278 @@ void AudioSession::ProcessAudioTranscriptionJson(const std::string& request_json
...
+  response.usage.total_tokens = completion_tokens;
+
+  AudioTranscriptionResponse output;
+  output.id = response_id;
+  output.text = std::move(text);
+  response.items.push_back(std::make_unique<TextItem>(nlohmann::json(output).dump(),
+                                                      FOUNDRY_LOCAL_TEXT_ITEM_TYPE_OPENAI_JSON));
+}
+
+std::vector<float> AudioSession::LoadPcmWavAsFloatSamples(const std::string& audio_file_path) {
```
kunal-vaishnavi, line 694: From Copilot:

Good addition overall — the Nemotron transcription path is clean, the WAV parser is thorough, and keeping a single generator across chunks is the right fix for truncated output. A few things worth addressing before merging:

---

**🔴 Blocking issues**

1. **WAV chunk bounds check is subtly wrong** (`LoadPcmWavAsFloatSamples`, chunk-walking loop)
   The bounds guard `chunk_start + chunk_size > file_size - (chunk_size % 2 == 1 ? 1 : 0)` mixes the padding-byte adjustment into the main bounds check, which can undercount available space and let a crafted file slip through. The padding byte lives *after* the chunk data, so it shouldn't reduce the window used to validate the chunk itself. Suggested fix:
   ```cpp
   // Before: chunk_start + chunk_size > file_size - (chunk_size % 2 == 1 ? 1 : 0)
   // After:
   if (chunk_start < 0 ||
       chunk_start + static_cast<std::streamoff>(chunk_size) > file_size) {
     FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_USAGE, "Corrupted WAV chunk.");
   }
   ```
   The odd-byte padding seek that follows handles alignment correctly as-is.

2. **Format validation happens after full data load** (`LoadPcmWavAsFloatSamples`, `data` chunk branch)
   `data.resize(chunk_size)` allocates the full audio buffer before sample rate and format have been validated. For a large or malformed file, this could allocate hundreds of MB only to throw immediately afterward. Consider moving the `sample_rate`, `audio_format`, and `bits_per_sample` checks to right after the `fmt ` chunk is parsed, before the `data` chunk is even encountered.

---

**🟡 Suggestions**

3. **Gap at language ID 5 in `NemotronLanguageIdMap`** — IDs jump from `4` (zh-cn) to `6` (hi). If this is intentional (reserved/unsupported language), a brief comment would prevent future contributors from assuming it's a typo.

4. **`prompt_tokens` hardcoded to `0`** in `ProcessNemotronFileTranscription` — the existing path computes this from the processor. If it's not meaningful for Nemotron, a `// TODO` explaining why would help maintain parity expectations with the OpenAI usage spec.

5. **`decode_all_tokens` lambda and thread-safety** — the lambda captures `generator` by reference and is safe as used today (same-thread, same scope). A short comment noting it must not be moved off the creation thread would protect against future refactors.

rui-ren:
@kunal-vaishnavi can you please review again?  We are hoping to add this to next release. Thank you so much .

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -502,4 +554,278 @@ void AudioSession::ProcessAudioTranscriptionJson(const std::string& request_json
...
+  if (req.language.has_value()) {
+    language = *req.language;
+  } else {
+    auto session_lang_it = SessionOptions().find("language");
+    if (session_lang_it != SessionOptions().end()) {
+      language = session_lang_it->second;
+    }
+  }
+
+  auto samples = LoadPcmWavAsFloatSamples(req.filename);
```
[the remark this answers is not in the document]
kunal-vaishnavi: This is a valid issue. We need unit tests to validate these changes such as:

- Happy path (`nemotron-3.5-asr-streaming-0.6b` with a valid 16kHz WAV)
- Any regression fixes in this change
- Parser validation cases like malformed headers and unsupported sample rates
- PCM16 and float32 input handling

kunal-vaishnavi:
> @kunal-vaishnavi can you please review again? We are hoping to add this to next release. Thank you so much .

I reviewed the PR again. Can you add unit tests to validate your changes and ensure there's no breaking change?

rui-ren:
@kunal-vaishnavi can you please review it again? Thanks

sdk_v2/cpp/test/internal_api/audio/audio_session_test.cc
```diff
@@ -431,6 +515,145 @@ TEST_F(AudioSessionTest, OpenAIJsonWithInvalidJsonThrows) {
   EXPECT_THROW(session.ProcessRequest(request, response), nlohmann::json::parse_error);
 }
 
+TEST_F(AudioSessionTest, NemotronOpenAIJsonWithInvalidRiffThrows) {
```
kunal-vaishnavi, line 654: From Copilot:

Could we add a regression test that uses a multi-chunk WAV and verifies we get the full transcript back? The main bug this PR is fixing is truncation during file-based Nemotron transcription, so having an end-to-end case that proves the full output survives chunked decoding would give us much stronger confidence here.

sdk_v2/cpp/test/internal_api/audio/audio_session_test.cc
```diff
@@ -431,6 +515,145 @@ TEST_F(AudioSessionTest, OpenAIJsonWithInvalidJsonThrows) {
   EXPECT_THROW(session.ProcessRequest(request, response), nlohmann::json::parse_error);
 }
```
kunal-vaishnavi, line 653: From Copilot:

Could we add an explicit unit test for the float32 WAV path in `LoadPcmWavAsFloatSamples`? The parser has separate decoding logic for PCM16 and float32, so having coverage for float32 would help make sure that branch keeps working as expected.

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -502,4 +547,303 @@ void AudioSession::ProcessAudioTranscriptionJson(const std::string& request_json
...
+                                         int& completion_tokens) const {
+  if (!tensors || original_request.canceled) {
+    return;
+  }
+
+  generator.SetInputs(*tensors);
+  DecodeNemotronTokens(generator, tokenizer_stream, text, streaming_callback, response_id, original_request,
+                       completion_tokens);
+}
+
```
kunal-vaishnavi, line 629: From Copilot:

Could we consider moving `decode_all_tokens` into a named private helper instead of keeping it as a local lambda? This logic is already central to the new Nemotron path, and giving it a dedicated method would make the control flow a bit easier to follow, test, and extend if we need to change the decode behavior later.

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -71,8 +77,48 @@ std::string JoinTokens(const std::vector<std::string>& token_texts) {
   return out;
 }
 
+const std::unordered_map<std::string, std::string>& NemotronLanguageIdMap() {
+  // Language id 5 is intentionally missing because upstream Nemotron lang_id assignments skip it.
+  static const std::unordered_map<std::string, std::string> kMap = {
```
kunal-vaishnavi, line 84: What is the language id for 5? I see it goes from 4 to 6 directly.

rui-ren: @kunal-vaishnavi  That is removed by Nvidia, you can check here.
https://microsoft.visualstudio.com/windows.ai.toolkit/_git/neutron-server/pullrequest/15742356

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -502,4 +547,303 @@ void AudioSession::ProcessAudioTranscriptionJson(const std::string& request_json
...
+  }
+}
+
+void AudioSession::DecodeNemotronTokens(OgaGenerator& generator, OgaTokenizerStream& tokenizer_stream, std::string& text,
+                                        const std::unique_ptr<CallbackHandler>& streaming_callback,
+                                        const std::string& response_id, const Request& original_request,
+                                        int& completion_tokens) const {
+  const bool is_streaming = (streaming_callback != nullptr);
+
+  while (!generator.IsDone() && !original_request.canceled) {
```
kunal-vaishnavi: From Copilot:

Could we also stop decoding when the generator has entered a terminated session? The existing streaming path checks both `IsDone()` and `IsSessionTerminated()` before calling `GenerateNextToken()`, and this Nemotron loop only checks `IsDone()`. If the generator terminates early, we can end up continuing to call `GenerateNextToken()` without making progress, which may leave the request hanging.

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -502,4 +547,303 @@ void AudioSession::ProcessAudioTranscriptionJson(const std::string& request_json
...
+  int completion_tokens = 0;
+
+  constexpr size_t kNemotronSamplesPerChunk = 1600;  // 100ms at 16kHz
+  for (size_t offset = 0; offset < samples.size() && !original_request.canceled;
+       offset += kNemotronSamplesPerChunk) {
+    size_t count = std::min(kNemotronSamplesPerChunk, samples.size() - offset);
+    RunNemotronDecodePass(processor->Process(samples.data() + offset, count), *generator, *tokenizer_stream, text,
+                          streaming_callback, response_id, original_request, completion_tokens);
+  }
+  RunNemotronDecodePass(processor->Flush(), *generator, *tokenizer_stream, text, streaming_callback, response_id,
```
kunal-vaishnavi: From Copilot:

Could we skip the final `Flush()` call when the request has already been canceled? The streaming path avoids flushing after cancellation, and doing it here still forces the processor to do extra work even though the request is already being torn down.

rui-ren:
@kunal-vaishnavi , can you please review this one?

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -279,9 +325,7 @@ void AudioSession::ProcessStreamingAudio(const AudioItem& format_item, ItemQueue
   auto effective_kvp = MergedOptions(request.options);
   SearchOptions options = SearchOptions::FromParameters(effective_kvp);
 
-  if (options.temperature.has_value()) {
-    gen_params->SetSearchOption("temperature", *options.temperature);
```
kunal-vaishnavi, line 283: Thinking about this some more, let's revert this change and keep the original `if` branching. Having a default of `0.0f` could change standard behavior.

sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -502,4 +549,307 @@ void AudioSession::ProcessAudioTranscriptionJson(const std::string& request_json
...
+        FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_USAGE,
+                 fmt::format("Unsupported WAV format: audioFormat={}, bitsPerSample={}.", audio_format,
+                             bits_per_sample));
+      }
+    } else if (chunk_id == "data") {
+      if (channels <= 0 || sample_rate <= 0 || bits_per_sample <= 0) {
+        FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_USAGE, "Missing WAV fmt chunk before data.");
+      }
+
+      data.resize(chunk_size);
```
kunal-vaishnavi, line 786: From Copilot:

`chunk_size` comes directly from the WAV file, so this path can allocate an arbitrarily large buffer with `data.resize(chunk_size)` before applying any resource limit. A valid multi-gigabyte WAV could therefore cause excessive memory usage, and the subsequent decoded `samples` allocation increases the peak further.

Please enforce a maximum data size/sample count before allocating, or decode the data chunk incrementally instead of buffering the entire chunk. For example:

```cpp
if (chunk_size > kMaxWavDataBytes) {
  FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_USAGE,
           "WAV data chunk exceeds the maximum supported size.");
}
```

Please also add a regression test covering an oversized but otherwise structurally valid `data` chunk.

kunal-vaishnavi (dismissed):

kunal-vaishnavi (approved):

PR #894, opened 2026-07-21
Add Linux arm64 support to the SDK v2 packaging pipelines

baijumeswani:
Adds first-class Linux arm64 support across the SDK v2 build, test, and packaging pipelines. Every package format the pipeline already produces (C++ tgz, Runtime NuGet, Python wheel, npm tarball) now includes a `linux-arm64` payload, built and tested natively on the `onnxruntime-linux-ARM64-CPU-2019` pool.

Initial Linux ARM64 support is CPU-only.


.pipelines/v2/templates/stages-js.yml
```diff
@@ -136,6 +136,43 @@ stages:
         includeArtifactDir: '$(Pipeline.Workspace)/cpp-native-include'
         outputDir: '$(Build.ArtifactStagingDirectory)/js-prebuild'
 
+- stage: js_build_linux_arm64
+  displayName: 'JS SDK: Build Linux ARM64'
+  dependsOn:
+  - compute_version
+  - cpp_build_linux_arm64
+  - cpp_build_win_x64
```
prathikr, line 144: why does this stage depend on win x64?

.pipelines/v2/templates/stages-python.yml
```diff
@@ -1,14 +1,18 @@
...
+# tests win-x64 + linux-x64 + linux-arm64 + osx-arm64
+# (win-arm64 is build-only, matching the C# matrix).
 #
 # Each build stage depends on the matching native build stage from
 # stages-build-native.yml and consumes the `cpp-native-<rid>` pipeline artifact
 # directly (no NuGet — the wheel bundles the native lib at
 # src/foundry_local_sdk/_native/<rid>/). On Windows that native artifact carries
 # the reg-free WinML 2.x runtime, so the single wheel includes WinML support.
 #
+# linux-arm64 is CPU-only; it uses onnxruntime + onnxruntime-genai (plain
```
prathikr: can you include why this is the case and what needs to happen in order to enable linux-arm64+gpu support?

.pipelines/v2/templates/stages-python.yml
```diff
@@ -110,6 +114,78 @@ stages:
...
+    steps:
+    - checkout: self
+      clean: true
+    - template: steps-build-python.yml
+      parameters:
+        nativeArtifactDir: '$(Pipeline.Workspace)/cpp-native-linux-arm64'
+        rid: 'linux-arm64'
+        useSystemPython: true
+        outputDir: '$(Build.ArtifactStagingDirectory)/python-sdk'
+    # Assert the produced wheel has the correct platform tag and bundles the
```
prathikr: Is this something we should be doing for all platforms?

sdk_v2/cpp/cmake/FindOnnxRuntime.cmake
```diff
@@ -132,41 +136,47 @@ else()
...
-        FetchContent_MakeAvailable(ort_gpu_linux)
+            FetchContent_Declare(ort_gpu_linux URL ${ORT_GPU_LINUX_FETCH_URL} DOWNLOAD_EXTRACT_TIMESTAMP TRUE DOWNLOAD_NAME ort_gpu_linux.zip)
+            FetchContent_MakeAvailable(ort_gpu_linux)
 
-        set(_ORT_LIB_DIR "${ort_gpu_linux_SOURCE_DIR}/runtimes/${_ORT_PLATFORM}/native")
-        message(STATUS "OnnxRuntime GPU Linux package: ${ort_gpu_linux_SOURCE_DIR}")
+            set(_ORT_LIB_DIR "${ort_gpu_linux_SOURCE_DIR}/runtimes/${_ORT_PLATFORM}/native")
+            message(STATUS "OnnxRuntime GPU Linux package: ${ort_gpu_linux_SOURCE_DIR}")
+        else()
+            # linux-arm64: libonnxruntime.so is included directly in the Microsoft.ML.OnnxRuntime.Foundry
```
prathikr, line 176: Will this impact our future process for removing the .Foundry package and relying on the plain ort cpu package (once all eps are plugin)?

sdk_v2/cpp/nuget/test_pack.py
```diff
@@ -0,0 +1,276 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""Focused unit tests for pack.py staging behaviour.
+
+Run with:
+    python -m unittest test_pack        (from sdk_v2/cpp/nuget/)
+    python test_pack.py                 (direct invocation)
```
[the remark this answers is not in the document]
prathikr: OR, include more comments/details on when this should be run and what the purpose of this testing script is

sdk_v2/cpp/src/ep_detection/webgpu_ep_bootstrapper.cc
```diff
@@ -75,13 +75,18 @@ const std::unordered_map<std::string_view, WebGpuPackageMetadata> kPackageMetada
 #endif
 
 // Platform-specific package metadata is baked into the binary to keep
-// verification inputs fixed at build time.
+// verification inputs fixed at build time. linux-arm64 is intentionally absent
```
prathikr: Is the CDN package updated? Might need to do this before merging this PR

prathikr (commented):
Can you add to the PR description all the new unit tests that have been introduced and their purposes?

.pipelines/v2/templates/stages-python.yml
baijumeswani: I'll add a comment. To enable linux-arm64 + GPU, we need to move to the plugin cuda ep. Moreover, the onnxruntime-genai linux-arm64 should also build the libonnxruntime-genai-cuda.so which is missing right now.

.pipelines/v2/templates/stages-python.yml
baijumeswani: I'll make sure that other platforms also follow this pattern.

.pipelines/v2/templates/stages-js.yml
baijumeswani: It appears that the cpp_build_win_x64 is the only stage that packages the native cpp include package and makes it available to all other js-addon jobs. I think it can be simplified in the future. But I won't update this in this PR as it touches all platforms not just linux-arm64.

sdk_v2/cpp/cmake/FindOnnxRuntime.cmake
baijumeswani: No, it should not. The cpu onnxruntime package also contains the libonnxruntime.so for linux-arm64.

sdk_v2/cpp/src/ep_detection/webgpu_ep_bootstrapper.cc
baijumeswani: I think this comment is misleading. I'll address this.

prathikr (dismissed):

prathikr (approved):

baijumeswani:
Thank you for the review! :)

PR #908, opened 2026-07-26
Preserve HTTP errors when loading JS models

fallintoplace:
## Summary

- limit network error wrapping to failures thrown by `fetch`
- preserve unsuccessful HTTP responses as server errors with the status code
- add focused coverage for HTTP and transport failures

## Validation

- `npm test -- --grep "ModelLoadManager external service errors"`
- `npm run build`
- isolated TypeScript check for the changed source and test


baijumeswani:
sdk v1 is not under active development at this time. we will only be doing sdk v2 release going forward (except for addressing critical issues).

Issue #905, opened 2026-07-25
Streaming chat crashes with IndexError on the final chunk (empty choices) - Python SDK


dipeshpandit12:
# Streaming chat crashes with `IndexError` on the final chunk (empty `choices`) — Python SDK

## Summary
When consuming `complete_streaming_chat(...)` from the Python SDK, the **final stream chunk
carries an empty `choices` list**. The SDK's own documented usage example indexes
`chunk.choices[0]` unconditionally, so any consumer following the canonical pattern crashes with
`IndexError: list index out of range` **after** the model has already produced its full answer.

- **Component:** `foundry-local-sdk` (Python)
- **Severity:** High (canonical usage crashes; the example in the code itself is affected)
- **Type:** Correctness / API-contract
- **Status:** Reproduced

## Environment
| | |
|---|---|
| SDK | `foundry-local-sdk` 1.2.3 (PyPI) |
| Python | 3.13.5 |
| OS / HW | macOS (Darwin 25.5.0), Apple silicon (arm64) |
| Execution provider | WebGpuExecutionProvider |
| Model | `qwen2.5-0.5b` (`qwen2.5-0.5b-instruct-generic-gpu-4`, v4) |

## Where the bug lives
`sdk/python/src/openai/chat_client.py`

- `_on_chunk` (~line 234) correctly iterates `raw.get("choices", [])` and **can yield a
  `ChatCompletionChunk` whose `choices` list is empty** (the terminating chunk).
- The docstring/example for `complete_streaming_chat` (~lines 273–274) demonstrates:
  ```python
  for chunk in client.complete_streaming_chat(messages):
      if chunk.choices[0].delta.content:   # <-- IndexError on the empty final chunk
          ...
  ```
The library does not filter the empty chunk and does not document that it occurs, so it pushes the
failure onto every caller — including its own example, which the sample apps copy.

## Steps to reproduce
1. Install the SDK and load a small model:
   ```bash
   python3 -m venv .venv && source .venv/bin/activate
   pip install foundry-local-sdk
   ```
2. Run the shipped sample `samples/python/tutorial-chat-assistant/src/app.py` (its streaming loop
   uses `chunk.choices[0].delta.content`), or the minimal repro below.

### Minimal reproduction
```python
from foundry_local_sdk import Configuration, FoundryLocalManager

FoundryLocalManager.initialize(Configuration(app_name="repro"))
manager = FoundryLocalManager.instance
[...]

PR #919, opened 2026-07-27
fix(python): guard empty choices on final streaming chunk

dipeshpandit12:
Related issue: https://github.com/microsoft/Foundry-Local/issues/905

## Problem
The final chunk yielded by `complete_streaming_chat` can carry an empty `choices` list. The documented example in `chat_client.py` and the streaming samples indexed `chunk.choices[0]` unconditionally, raising `IndexError` **after** the full response had already been produced.

## Fix
Guard with `if not chunk.choices: continue` in:
- `sdk/python/src/openai/chat_client.py` (documented example in the docstring)
- `samples/python/native-chat-completions/src/app.py`
- `samples/python/tutorial-chat-assistant/src/app.py`

## Testing
Reproduced on macOS (arm64), Python 3.13, `foundry-local-sdk` 1.2.3, model `qwen2.5-0.5b` via the WebGPU EP. After the fix, streaming completes cleanly with no exception — verified with the interactive `app.py` over a 3-turn conversation and a single-prompt run.

### DIFF
--- a/samples/python/native-chat-completions/src/app.py
+++ b/samples/python/native-chat-completions/src/app.py
@@ -51,6 +51,8 @@ def main():
     # Stream the response token by token
     print("Assistant: ", end="", flush=True)
     for chunk in client.complete_streaming_chat(messages):
+        if not chunk.choices:  # the final chunk can carry no choices
+            continue
         content = chunk.choices[0].delta.content
         if content:
             print(content, end="", flush=True)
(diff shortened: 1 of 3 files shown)


prathikr:
Please rename the branch as our CI pipelines cannot find branches using the `:` character

dipeshpandit12:
Renamed the branch to fix-streaming-empty-choices per your request (dropped the / and the fork : notation). GitHub closed this PR when the old ref disappeared, so I've reopened it as #922 . Same commit.

PR #922, opened 2026-07-28
fix(python): guard empty choices on final streaming chunk

dipeshpandit12:
Related issue: #905
Supersedes #919 (closed after renaming the branch to remove the `/` so Azure Pipelines can resolve it).

## Problem
The final chunk yielded by `complete_streaming_chat` can carry an empty `choices` list. The documented example in `chat_client.py` and the streaming samples indexed `chunk.choices[0]` unconditionally, raising `IndexError` after the full response had already been produced.

## Fix
Guard with `if not chunk.choices: continue` in:
- `sdk/python/src/openai/chat_client.py` (documented example in the docstring)
- `samples/python/native-chat-completions/src/app.py`
- `samples/python/tutorial-chat-assistant/src/app.py`

## Testing
Reproduced on macOS (arm64), Python 3.13, foundry-local-sdk 1.2.3, model qwen2.5-0.5b via the WebGPU EP. After the fix, streaming completes cleanly with no exception — verified with the interactive `app.py` over a 3-turn conversation and a single-prompt run.

### DIFF
--- a/samples/python/native-chat-completions/src/app.py
+++ b/samples/python/native-chat-completions/src/app.py
@@ -51,6 +51,8 @@ def main():
     # Stream the response token by token
     print("Assistant: ", end="", flush=True)
     for chunk in client.complete_streaming_chat(messages):
+        if not chunk.choices:  # the final chunk can carry no choices
+            continue
         content = chunk.choices[0].delta.content
         if content:
             print(content, end="", flush=True)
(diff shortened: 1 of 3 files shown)


prathikr:
Still not working, ported your changes in https://github.com/microsoft/foundry-local/pull/926

PR #928, opened 2026-07-30
The Implementation for Bring Your Own Local Model (BYOM) API

selenayang888:
The implementation to register a local ONNX model from an arbitrary filesystem path and make it available through the standard Catalog/Model/Inference api.


sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -60,7 +60,7 @@
  * Incremented with each release.
  * Used to request the API function table via FoundryLocalGetApi.
  * ----------------------------------------------------------------------- */
-#define FOUNDRY_LOCAL_API_VERSION 1
+#define FOUNDRY_LOCAL_API_VERSION 3
```
prathikr: why 1 --> 3 and not 1 --> 2?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -202,6 +202,12 @@ typedef enum flDeviceType {
   FOUNDRY_LOCAL_DEVICE_NPU = 3
 } flDeviceType;
 
+typedef enum flCatalogType {
+  FOUNDRY_LOCAL_CATALOG_PUBLIC = 0,
+  FOUNDRY_LOCAL_CATALOG_LOCAL = 1,
+  FOUNDRY_LOCAL_CATALOG_PRIVATE = 2,
```
prathikr: is private catalog ready yet? if not, I don't think we should allow this option at all.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -299,14 +302,33 @@ struct Runtime {
   std::optional<std::string_view> execution_provider;
 };
 
+enum class CatalogType {
+  Public = FOUNDRY_LOCAL_CATALOG_PUBLIC,
+  Local = FOUNDRY_LOCAL_CATALOG_LOCAL,
+  Private = FOUNDRY_LOCAL_CATALOG_PRIVATE,
```
prathikr: is private catalog ready?

sdk_v2/cpp/src/catalog/base_model_catalog.cc
```diff
@@ -360,15 +376,50 @@ std::vector<Model*> BaseModelCatalog::GetLoadedModels() const {
...
+  std::lock_guard<std::mutex> lock(mutex_);
+  auto container = std::make_unique<Model>(Model::MakeContainer(std::move(model)));
+  container->SelectDefaultVariant();
+  auto* result = container.get();
+  models_.push_back({std::move(container), true});
+  RebuildIndex();
+  return result;
+}
+
+bool BaseModelCatalog::DeactivateModel(const std::string& alias_or_model_id) {
```
prathikr: regarding naming, AddModel and DeactivateModel don't intuitively seem like opposites of each other. either AddModel & RemoveModel or ActivateModel & DeactivateModel make sense.

prathikr: followup, perhaps we need all 4 instead of just one of the pairs...

sdk_v2/cpp/test/internal_api/azure_catalog_test.cc
prathikr, line 609: this test makes a model named "custom-model" but the below assert seems to expect "phi-4-mini"?

sdk_v2/cpp/test/internal_api/local_model_catalog_test.cc
```diff
@@ -0,0 +1,391 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
prathikr, line 1: can we add some tests that actually run a local model to make sure it works e2e?

sdk_v2/cpp/test/internal_api/azure_catalog_test.cc
selenayang888: `custom-model:0` is not created by this test; it is passed as a cached model ID for catalog lookup. The first mocked request returns `phi-4-mini:3`, while the second request cannot resolve `custom-model:0`. Because unresolved custom models must not be synthesized as public catalog entries, the final result should contain only `phi-4-mini:3`. Therefore, the assertion is intentional.

sdk_v2/cpp/src/catalog/base_model_catalog.cc
selenayang888: These are internal storage operations rather than a public lifecycle pair.The public API has remained `RegisterModel()` and `UnregisterModel()`. Regarding to renaming the protected helpers to something implementation-oriented, the safest rename is:

`AddModel()` → `AppendActiveModel()`
`DeactivateModel()`→ `RetireModel()`

This communicates the actual asymmetric behavior:
`AppendActiveModel()` appends a newly registered active model.
`RetireModel()` hides a model from catalog lookup while retaining its object to preserve outstanding pointers.
Neither `ActivateModel()` nor `RemoveModel()` should be introduced because reactivation and physical removal are unsupported.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
selenayang888: Thanks. I changed it back to version 2.

sdk_v2/cpp/test/internal_api/local_model_catalog_test.cc
selenayang888: Added a local model `qwen2.5-coder-0.5b-instruct-generic-cpu-3` in `FOUNDRY_TEST_DATA_DIR` for a BYOM e2e test.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -781,6 +806,15 @@ class ICatalog {
   virtual ModelList GetModelVersions(const std::string& model_alias,
                                      const std::string& variant_name = {},
                                      int max_versions = 50) = 0;
+  virtual std::unique_ptr<IModel> RegisterModel(const ModelInfo&) {
+    throw Error("models can only be registered in a local catalog", FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT);
+  }
+  virtual void UnregisterModel(const std::string&) {
```
natke: Suggest "Deregister"

natke (commented):
Confirming that the SDKs also need adding to this PR

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,549 @@
...
+        *assets_inspected = true;
+      }
+      SetModelInfoIntProperty(resolved, FOUNDRY_LOCAL_MODEL_PROP_FILESIZE_BYTES_INT, DirectorySize(model_path));
+      resolved.int_properties.erase(FOUNDRY_LOCAL_MODEL_PROP_CONTEXT_LENGTH_INT);
+      if (config.model && config.model->context_length > 0) {
+        SetModelInfoIntProperty(resolved, FOUNDRY_LOCAL_MODEL_PROP_CONTEXT_LENGTH_INT, config.model->context_length);
+      }
+      // Keep the default provider in genai_config.json authoritative when the caller did not supply one. Some OGA
+      // providers such as DML are not represented by the SDK's explicit ExecutionProvider enum and use kDefault.
+      std::string task = "chat-completion";
```
natke: This will overwrite user supplied metadata?

sdk_v2/cpp/src/c_api.cc
```diff
@@ -715,6 +753,57 @@ FL_API_STATUS_IMPL(Catalog_GetModelVersionsImpl, const flCatalog* catalog,
...
+  list->items.reserve(models.size());
+  for (auto* model : models) {
+    list->items.push_back(AsHandle<flModel>(model));
+  }
+  *out_models = list.release();
+  return nullptr;
+  API_IMPL_END
+}
+
+static const flCatalogApi g_catalog_api_v1 = {
```
baijumeswani: Why do we need `g_catalog_api_v1`? The API changes are purely additive and do not break ABI. So we should be ok just expanding `g_catalog_api`?

sdk_v2/cpp/src/c_api.cc
```diff
@@ -970,6 +1053,102 @@ static int64_t FL_API_CALL Info_GetIntPropertyImpl(const flModelInfo* info,
...
+  if (!info) {
+    return MakeStatus(FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT, "info must not be null");
+  }
+
+  *out_info = AsHandle<flModelInfo>(new fl::ModelInfo(*AsImpl(info)));
+  return nullptr;
+  API_IMPL_END
+}
+
+static const flModelApi g_model_api_v1 = {
```
baijumeswani: Same here. The api changes should be purely additive. We should not need to define `g_model_api_v1`.

sdk_v2/cpp/src/c_api.cc
```diff
@@ -1888,6 +2082,40 @@ static const flApi g_api_v1 = {
     Manager_IsEpDownloadInProgressImpl,
     Manager_ShutdownImpl,
     Manager_IsShutdownRequestedImpl,
+  };
+
+  static const flApi g_api_v2 = {
```
baijumeswani: Same here. We do not need this. Any reason we need a new `g_api_v2`?

baijumeswani (commented):
Will add more comments soon.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -256,6 +262,16 @@ typedef enum flTensorDataType {
...
+#define FOUNDRY_LOCAL_MODEL_PROP_DEVICE_TYPE_STR "device_type"                  ///< CPU, GPU, or NPU
+#define FOUNDRY_LOCAL_MODEL_PROP_EP_STR "execution_provider"                    ///< optional execution provider
+#define FOUNDRY_LOCAL_MODEL_PROP_ENTITY_TYPE_STR "entity_type"                  ///< fixed to "Model" for BYOM
+#define FOUNDRY_LOCAL_MODEL_PROP_AUTHOR_STR "author"                            ///< optional
+#define FOUNDRY_LOCAL_MODEL_PROP_QUANTIZATION_STR "quantization"                ///< optional
+#define FOUNDRY_LOCAL_MODEL_PROP_CREATION_TIME_STR "creation_time"              ///< ISO-8601 UTC timestamp
+
+/* flModelInfo registration properties */
+#define FOUNDRY_LOCAL_REG_MODEL_PATH "model_path"
+#define FOUNDRY_LOCAL_REG_ALIAS "alias"
```
skottmckay: Should these two values be provided via metadata properties or directly to the `RegisterModel` call? 

Instead of 'alias' should the user provide 'id' in format `<alias>:<version>`? Simplifies the input and maintains consistency with our catalog entries where id must be unique and in that format.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -265,6 +281,9 @@ typedef enum flTensorDataType {
 #define FOUNDRY_LOCAL_MODEL_PROP_CREATED_AT_UNIX_INT "created_at_unix"              ///< Unix timestamp. default=0
 #define FOUNDRY_LOCAL_MODEL_PROP_IS_TEST_MODEL_INT "is_test_model"                  ///< bool (0=false, 1=true)
 #define FOUNDRY_LOCAL_MODEL_PROP_CONTEXT_LENGTH_INT "context_length"                ///< optional int64_t
+#define FOUNDRY_LOCAL_MODEL_PROP_VERSION_INT "version"                              ///< optional non-negative integer
+#define FOUNDRY_LOCAL_MODEL_PROP_FILESIZE_BYTES_INT "file_size_bytes"                ///< optional int64_t
```
skottmckay: Why bytes if the 'real' catalog property is fileSizeMb?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -706,6 +725,12 @@ typedef struct flApi {
   bool FL_API_T(Manager_IsShutdownRequested, _In_ const flManager* manager);
 
   // End V1
+  FL_API_STATUS(Manager_GetCatalogByType, _In_ const flManager* manager, flCatalogType catalog_type,
+                _Outptr_ flCatalog** out_catalog);
+  FL_API_STATUS(Manager_GetCatalogByName, _In_ const flManager* manager, _In_ const char* catalog_name,
+                _Outptr_ flCatalog** out_catalog);
```
skottmckay: Is this something the user needs to know about?

Internally we could have a local catalog that we automatically use for RegisterModel, but that seems like it could stay an implementation detail and there's just a single user facing flCatalog. 

Ideally the internals of FL aggregate across the various catalogs available (public/private/local) and the user doesn't need to care about that detail. V1 had a catalog aggregator class that I didn't port across as it hadn't been used yet. We could potentially add that concept back in.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -969,6 +994,15 @@ struct flCatalogApi {
                 _In_opt_ const char* model_name, int32_t max_versions, _Outptr_ flModelList** out_models);
 
   // End V1
+  /// Register a model in a local catalog. The input ModelInfo is copied.
+  FL_API_STATUS(RegisterModel, _In_ flCatalog* catalog, _In_ const flModelInfo* model_info,
+                _Outptr_ flModel** out_model);
+  /// Unregister by alias or model ID without deleting model assets.
+  FL_API_STATUS(UnregisterModel, _In_ flCatalog* catalog, _In_ const char* alias_or_model_id);
+  /// List models explicitly registered in this local catalog.
+  FL_API_STATUS(GetLocalModels, _In_ const flCatalog* catalog, _Outptr_ flModelList** out_models);
```
skottmckay: Is this needed? Does a user need to know/care that a model was added directly?

We generally want to keep the public API as minimal as possible. Anything we add here is permanent so it must have a clear requirement to justify expanding the API surface.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -1038,6 +1072,17 @@ struct flModelApi {
...
   // End V1
+  /// Create a caller-owned mutable ModelInfo. Release it with ReleaseModelInfo.
+  FL_API_STATUS(CreateModelInfo, _Outptr_ flModelInfo** out_info);
+  void FL_API_T(ReleaseModelInfo, _Frees_ptr_opt_ flModelInfo* info);
+  FL_API_STATUS(Info_SetStringProperty, _In_ flModelInfo* info, _In_ const char* key, _In_ const char* value);
+  FL_API_STATUS(Info_SetIntProperty, _In_ flModelInfo* info, _In_ const char* key, int64_t value);
+  FL_API_STATUS(Info_SerializeToFile, _In_ const flModelInfo* info, _In_ const char* file_path);
+  FL_API_STATUS(Info_DeserializeFromFile, _In_ const char* file_path, _Outptr_ flModelInfo** out_info);
+  /// Create a caller-owned deep copy. Release it with ReleaseModelInfo.
+  FL_API_STATUS(Info_Clone, _In_ const flModelInfo* info, _Outptr_ flModelInfo** out_info);
```
skottmckay: Naively I would have expected the user adding a model to create the model info in memory and provide that in a RegisterModel call. Why does the user need to serialize/deserialize or clone?

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -299,14 +302,33 @@ struct Runtime {
...
  public:
-  explicit ModelInfo(const flModelInfo& info) noexcept : info_(&info) {}
+  ModelInfo();
+  explicit ModelInfo(const flModelInfo& info) noexcept : handle_(&info) {}
+
+  /// Create an independent, owning, mutable deep copy, including when the source is a borrowed view.
+  ModelInfo(const ModelInfo& other);
+  ModelInfo& operator=(const ModelInfo& other);
+  ModelInfo(ModelInfo&&) noexcept = default;
+  ModelInfo& operator=(ModelInfo&&) noexcept = default;
```
skottmckay: Do we need copy and assignment?

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,549 @@
...
+#define NOMINMAX
+#include <windows.h>
+#endif
+
+namespace fl {
+namespace {
+
+constexpr const char* kRegistrationIdProperty = "_local_registration_id";
+
+std::string UtcTimestamp(int64_t unix_time) {
```
skottmckay: Generic helpers could go in utils so they can be re-used if needed.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,549 @@
...
+bool HasParentTraversal(const std::filesystem::path& path) {
+  for (const auto& component : path) {
+    if (component == "..") {
+      return true;
+    }
+  }
+  return false;
+}
+
+int64_t DirectorySize(const std::filesystem::path& path) {
```
skottmckay: Do we need this? There's a property where the user can specify the size. The more code we have the more things to test and maintain and that can break at runtime, so I'm wondering if this is providing enough value.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
skottmckay: We need to define a set of known task values, and the user needs to specify. The task is used to determine which session types are valid for the model, so needs to match the model.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,549 @@
...
+}
+
+void WriteOptionalYamlInt(std::ostream& stream, const ModelInfo& info, const char* key, const char* yaml_key) {
+  const auto* value = info.GetPropertyInt(key);
+  if (value) {
+    stream << yaml_key << ": " << *value << '\n';
+  }
+}
+
+nlohmann::json RegistrationToJson(const LocalModelCatalog::Registration& registration) {
```
skottmckay: Is there a need to have both yaml and json formats being used?

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,549 @@
...
+  }
+  if (!alias_value || alias_value->empty()) {
+    FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT, "alias is required");
+  }
+  if (!std::regex_match(*alias_value, std::regex("[A-Za-z0-9][A-Za-z0-9._-]*"))) {
+    FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT, "alias must match [a-zA-Z0-9][a-zA-Z0-9._-]*");
+  }
+
+  std::filesystem::path supplied_path(*model_path_value);
+  if (HasParentTraversal(supplied_path)) {
```
skottmckay: If there's delayed download (hopefully we can avoid that though) how does this work? User provides a path that we download to in the future?

Is it ok that there's a hard requirement that this path must be writable by FL in order to create the inference_model.json there? 

Would it be better to require the path to be under our model cache (e.g. we create `.<appname>/cache/models/local` and a user should place a BYOM model there prior to calling RegisterModel)?

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,549 @@
...
+  Registration registration;
+  {
+    std::lock_guard<std::mutex> guard(registration_mutex_);
+    FileLock file_lock(lock_path_);
+    auto registrations = LoadRegistrations();
+    for (const auto& existing : registrations) {
+      if (existing.info.alias == *alias_value) {
+        FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT,
+                 "a model with alias '" + *alias_value + "' is already registered");
+      }
```
skottmckay: Would have expected this to be based on id not alias.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,549 @@
...
+    }
+
+    bool assets_inspected = false;
+    registration = {ResolveMetadata(model_info, nullptr, model_path, *alias_value, &assets_inspected), model_path,
+            assets_inspected};
```
skottmckay: nit: run clang-format to fix alignment

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,549 @@
...
+  } catch (...) {
+    std::lock_guard<std::mutex> guard(registration_mutex_);
+    FileLock file_lock(lock_path_);
+    auto registrations = LoadRegistrations();
+    registrations.erase(std::remove_if(registrations.begin(), registrations.end(), [&](const Registration& entry) {
+                          return entry.info.GetPropertyWithDefault(kRegistrationIdProperty, std::string{}) ==
+                                 registration.info.GetPropertyWithDefault(kRegistrationIdProperty, std::string{});
+                        }),
+                        registrations.end());
+    SaveRegistrations(registrations);
```
skottmckay: Feels like too much is entangled here. Do we want to remove the registration if the activate fails or just if the CreateModel fails? 

Or neither? if we validated the registration info before saving originally wouldn't it only be a runtime error that is likely transient here so automated unregistration isn't desirable?

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,549 @@
...
+  const auto registration_id = registration.info.GetPropertyWithDefault(kRegistrationIdProperty, std::string{});
+  return model_factory_(
+      registration.info, registration.model_path,
+      [this, registration_id](const std::string& model_id) {
+        auto* current = GetModelVariant(model_id);
+        if (!current || !current->IsActive() ||
+            current->Info().GetPropertyWithDefault(kRegistrationIdProperty, std::string{}) != registration_id) {
+          FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_USAGE, "model is no longer registered");
+        }
+        const_cast<LocalModelCatalog*>(this)->UnregisterModel(model_id);
```
skottmckay: const_cast is a design smell. Why are we calling UnregisterModel in CreateModel?

sdk_v2/cpp/src/catalog/local_model_catalog.h
```diff
@@ -0,0 +1,53 @@
...
+#include <mutex>
+#include <optional>
+
+namespace fl {
+
+/// Mutable, persistent catalog for models registered from arbitrary local directories.
+class LocalModelCatalog final : public BaseModelCatalog {
+ public:
+  using ModelFactory = std::function<Model(ModelInfo, std::string, std::function<void(const std::string&)>,
+                                           std::function<std::optional<ModelInfo>()>)>;
```
skottmckay: Define the 2 function args with separate `using` statements to provide some context as to what they represent via the name. Would also be helpful to include the arg names in the std::function usage as well. Whilst they're not strictly required it helps a user understand the function input.

sdk_v2/cpp/src/manager.cc
```diff
@@ -305,27 +306,34 @@ Manager::Manager(const Configuration& config) : config_(config) {
...
       disable_region_fallback);
+  local_catalog_ = std::make_unique<LocalModelCatalog>(
+      *config_.app_data_dir,
+      [this](ModelInfo info, std::string local_path, std::function<void(const std::string&)> unregister_callback,
+              std::function<std::optional<ModelInfo>()> prepare_callback) {
+        return CreateLocalModel(std::move(info), std::move(local_path), std::move(unregister_callback),
+                                std::move(prepare_callback));
+      },
+      *logger_);
+  local_catalog_->ListModels();
```
skottmckay: do we need to call ListModels() here?

sdk_v2/cpp/src/catalog/local_model_catalog.cc
skottmckay: I see now it's in an unregister delegate. Can we split those out. The intention of the model factory is to be able to create a Model from the catalog info. unregister seems like a separate concern that could be untangled from a model creation factory.

sdk_v2/cpp/src/model.h
```diff
@@ -158,9 +183,17 @@ class Model {
...
+  mutable std::vector<std::unique_ptr<const ModelInfo>> info_snapshots_;
+  mutable std::atomic<const ModelInfo*> current_info_{nullptr};
   std::atomic<bool> cached_{false};
+  std::atomic<bool> active_{true};
   std::string local_path_;
+  std::string runtime_model_id_;
+  bool external_registration_ = false;
+  std::function<void(const std::string&)> unregister_callback_;
+  std::function<std::optional<ModelInfo>()> prepare_callback_;
+  mutable std::atomic<bool> metadata_prepared_{false};
```
skottmckay: This feels like a lot of complexity to be adding for BYOM which conceptually doesn't seem like it should be that different to the current model usage. If we require the model to be present at registration time does that simplify a lot of this?

sdk_v2/cpp/src/model.h
```diff
@@ -130,6 +143,11 @@ class Model {
   void Unload();
   void RemoveFromCache();
 
+  /// Mark this model and its variants inactive while retaining pointer validity.
+  void Deactivate();
+  void BeginUnregister();
+  void CancelUnregister();
```
skottmckay: Do we need to support cancellation of unregister?

sdk_v2/cpp/src/model.cc
```diff
@@ -156,6 +178,21 @@ Model Model::FromModelInfo(ModelInfo info,
...
 
+Model Model::FromLocalRegistration(ModelInfo info,
+                                   std::string local_path,
+                                   DownloadManager& download_manager,
+                                   ModelLoadManager& model_load_manager,
+                                   std::function<void(const std::string&)> unregister_callback,
+                                   std::function<std::optional<ModelInfo>()> prepare_callback) {
+  auto model = FromModelInfo(std::move(info), std::move(local_path), download_manager, model_load_manager);
+  model.external_registration_ = true;
+  model.runtime_model_id_ = "local/" + model.Info().model_id;
```
skottmckay: do we need to treat this differently to any other model id? the more differences we have between how models from public/private/local are treated the more complicated the system is.

sdk_v2/cpp/src/model.cc
```diff
@@ -213,23 +250,27 @@ const std::string& Model::Id() const {
...
 const ModelInfo& Model::Info() const {
   if (Model* sv = selected_variant_.load(std::memory_order_acquire)) {
     return sv->Info();
   }
 
-  return info_;
+  const auto* info = current_info_.load(std::memory_order_acquire);
+  if (!info) {
+    FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL, "model metadata is not initialized");
+  }
```
skottmckay: This feels like something that should never happen. If we create a Model instance it should be completely valid. Having a "the info might be missing and that's an error" state complicates things for an end-user.

sdk_v2/cpp/src/c_api.cc
skottmckay: Agree. Should be exceedingly rare that we create a copy of an existing API to expand.

sdk_v2/cpp/src/model.h
```diff
@@ -101,6 +110,10 @@ class Model {
 
   bool IsCached() const;
   bool IsLoaded() const;
+  bool IsActive() const {
+    Model* selected = selected_variant_.load(std::memory_order_acquire);
+    return selected ? selected->IsActive() : active_.load();
+  }
```
skottmckay: I'm assuming deregistration to be rare so models will generally never become inactive. Given that, is this providing much value or would it be sufficient to document that unregistering a model potentially invalidates a model list the user has?

sdk_v2/cpp/src/model_info.h
```diff
@@ -87,4 +88,14 @@ ModelInfo ModelInfoFromJson(const nlohmann::json& j);
 /// Serialize a ModelInfo to JSON.
 nlohmann::json ModelInfoToJson(const ModelInfo& info);
 
+/// Set a property while keeping the typed ModelInfo fields synchronized with well-known keys.
+void SetModelInfoStringProperty(ModelInfo& info, std::string key, std::string value);
+void SetModelInfoIntProperty(ModelInfo& info, std::string key, int64_t value);
```
skottmckay: Can these be methods like the existing getters?

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,549 @@
...
+    stream << yaml_key << ": " << *value << '\n';
+  }
+}
+
+nlohmann::json RegistrationToJson(const LocalModelCatalog::Registration& registration) {
+  return {
+      {"alias", registration.info.alias},
+      {"model_path", registration.model_path},
+      {"registered_at", registration.info.GetPropertyWithDefault(FOUNDRY_LOCAL_MODEL_PROP_CREATION_TIME_STR, {})},
+      {"properties", ModelInfoToPropertyBagJson(registration.info)},
```
skottmckay: can we serialize the ModelInfo as is (we serialize ModelInfo instances to model_info.json I believe) so we avoid going back to an untyped property bag?

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -299,14 +302,33 @@ struct Runtime {
...
+
+  ModelInfo& SetStringProperty(const char* key, const char* value);
+  ModelInfo& SetIntProperty(const char* key, int64_t value);
+  void SerializeToFile(const std::string& file_path) const;
+  static ModelInfo DeserializeFromFile(const std::string& file_path);
```
skottmckay: Why does a user need to de/serialize the model info?

sdk_v2/cpp/src/inferencing/model_load_manager.cc
```diff
@@ -47,6 +47,36 @@ std::string_view RequiredEpForModelId(std::string_view model_id) {
   return {};
 }
 
+/// Returns whether the provider is DirectML, which is supplied by WinML rather than a downloadable EP bootstrapper.
+bool IsDmlProvider(std::string_view provider) {
+  return provider == "dml" || provider == "DML" || provider == "DmlExecutionProvider" ||
+         provider == "DMLExecutionProvider";
+}
```
skottmckay, line 83: Is this related to the PR?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
baijumeswani: I slightly prefer having separate catalogs per type (public, local, private - if that materializes). Because aggregate views can cause id/name conflicts and it becomes harder for consumers to tell which catalog a particular model came from. We then would need to expose some other model attributes or query-able catalog arguments.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
baijumeswani: Let's remove private until it is available.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -376,8 +398,11 @@ class ModelInfo {
   std::optional<std::string_view> Capabilities() const noexcept;
 
  private:
+  explicit ModelInfo(flModelInfo& info);
   static std::string_view safe(const char* s) noexcept { return s ? s : ""; }
-  const flModelInfo* info_;
+  detail::Base<flModelInfo> handle_;
+
+  friend class Catalog;
```
baijumeswani: Why does Catalog need to be a friend of ModelInfo?

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -781,6 +806,15 @@ class ICatalog {
   virtual ModelList GetModelVersions(const std::string& model_alias,
                                      const std::string& variant_name = {},
                                      int max_versions = 50) = 0;
+  virtual std::unique_ptr<IModel> RegisterModel(const ModelInfo&) {
+    throw Error("models can only be registered in a local catalog", FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT);
+  }
+  virtual void UnregisterModel(const std::string&) {
+    throw Error("models can only be unregistered from a local catalog", FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT);
+  }
+  virtual ModelList GetLocalModels() const {
```
baijumeswani: Why do we need `GetLocalModels` when we have `GetModels`? Assuming local and public catalogs are different. GetModels on a local catalog should return the local models.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -870,7 +909,9 @@ class Manager {
   detail::Base<flManager> handle_;
   Configuration config_;
   mutable std::unique_ptr<Catalog> catalog_;
+  mutable std::unique_ptr<Catalog> local_catalog_;
```
baijumeswani: Should we define a struct like:

```
struct CatalogCollection {
    mutable std::unique_ptr<Catalog> public_;
    mutable std::unique_ptr<Catalog> local_;
} catalogs_;
```

And then access them using catalogs_.public_ and catalogs_.local_?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
skottmckay: Separate catalogs at the public API level adds overhead to the SDK user as they can't simply deal with a unified meta-catalog. Do we think that cost is outweighed by the benefit of manually being able to resolve a clash with the model id if it occurred and a default priority of `local > private > cloud` was used? 

I could understand adding a metadata entry to the model for the catalog type. Low cost, ignored by vast majority of SDK users, available if needed using existing API.

Thinking through potential clashes:

- local with cloud
  - local is higher priority as manually added
  - if you later added that to a cloud repo and want to prefer the cloud version you can unregister the local model
    - local model instance will be the one you're getting by default, you can check the metadata for catalog type if needed (although model location is likely just as good a signal as we're not putting those in our cloud model download cache currently) to know it's a local model that can be unregistered
      - but equally the unregister can return an error saying 'not a local model' as internally we'd still have separate catalogs
- private with public
  - private is higher priority
    - if you put something in your private catalog assumably it should be preferred over public
      - remove from private catalog if that's not the case

And I would expect an id clash to be an edge case so we should try not to complicate the API for that if we have other ways of handling it.

baijumeswani: There may be scenarios where the application wants all three versions of the same model from {public, private, local} and present to their users 1 or all 3 for different scenarios.
There may also be a scenario where the application has a model in the private model catalog before we make it available in the public catalog. Which could then force application to change the name (if they wanted to use both).

> Separate catalogs at the public API level adds overhead to the SDK user as they can't simply deal with a unified meta-catalog.

I like the idea of an aggregate catalog, but only as a view of the 3 distinct catalogs. Not as a catalog itself. For example, we could create a helper/function called `manager.GetAggregateCatalogView()` which would then aggregate the 3 catalogs and present them based on some priority strategy.

This was one of the problems I thought the aggregate catalog had before in the v1 core which got surfaced when we did the private model catalog development.

baijumeswani: There is also the scenario where one could have multiple private catalogs. Then the aggregation would become more involved and conflict resolution may require configurable strategy.

sdk_v2/cpp/src/model_info.h
selenayang888: `SetPropertyStr()` and `SetPropertyInt()` are now `ModelInfo` methods beside the corresponding getters in model_info.h.

sdk_v2/cpp/src/model.h
selenayang888: Yes, IsActive() itself was not providing value, which was only used by tests, so I removed it and changed the tests to verify observable behavior.

sdk_v2/cpp/src/model.h
selenayang888: We do not support cancellation of unregister. The method was only releasing the lifecycle guard after either success or rollback, so `CancelUnregister()` was misleading. I renamed it to `EndUnregister()` and retained the rollback path needed when persistence or validation fails.

sdk_v2/cpp/src/inferencing/model_load_manager.cc
selenayang888: As we now derive the required EP from genai_config.json. DirectML must be excluded from that registration guard because it is supplied by the bundled WinML/GenAI runtime rather than through DownloadAndRegisterEps(). Otherwise a config using the full DmlExecutionProvider name would be incorrectly rejected. I updated the test to exercise the full provider name.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,488 @@
...
+      registration.info, registration.model_path,
+      [this, registration_id](const std::string& model_id) {
+        auto* current = GetModelVariant(model_id);
+        if (!current || !current->IsActive() ||
+            current->Info().GetPropertyWithDefault(kRegistrationIdProperty, std::string{}) != registration_id) {
+          FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_USAGE, "model is no longer registered");
+        }
+        const_cast<LocalModelCatalog*>(this)->UnregisterModel(model_id);
+      },
+      [this, registration]() { WriteMetadata(registration); });
```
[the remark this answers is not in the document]
selenayang888: Deferred registration was removed. Registration now rejects unavailable assets and requires complete metadata immediately.

sdk_v2/cpp/src/c_api.cc
selenayang888: One append-only catalog API table now serves all supported versions.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
selenayang888: Registration uniqueness is now checked using the full `model_id` rather than the alias. Models such as my-model:1 and my-model:2 can therefore coexist and are grouped under the derived alias `my-model`. A duplicate full ID is still rejected, with focused test coverage confirming this behavior.

sdk_v2/cpp/src/model.h
selenayang888: Agreed. We now require the model directory and a parseable `genai_config.json` to be present at registration time. This allowed us to remove the deferred metadata preparation, snapshot storage, and provisioning callbacks that were present in the original diff.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
selenayang888: > This will overwrite user supplied metadata?

Resolved. Registration no longer infers or overwrites the caller-supplied task, input/output modalities, or filesize_mb from `genai_config.json`. It preserves the supplied `ModelInfo`, including custom properties, prompt templates, and model settings. Only catalog-owned identity and bookkeeping fields are normalized.

selenayang888: > We need to define a set of known task values, and the user needs to specify. The task is used to determine which session types are valid for the model, so needs to match the model.

Resolved. `task` is now required during registration and must be one of automatic-speech-recognition, chat-completion, embeddings, or vision-language-chat. Missing or unsupported values return `FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT`.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
selenayang888: `file_size_bytes` was redundant and `FOUNDRY_LOCAL_MODEL_PROP_FILESIZE_BYTES_INT` is removed.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
selenayang888: The reviewed generic `UtcTimestamp()` helper was moved to reusable `FormatUtcTimestamp()` in `time_utils.h`.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
selenayang888: Thanks, I updated based on this comment. `RegisterModel` now takes `model_path` and `model_id` directly, while `ModelInfo` contains only descriptive metadata.

`model_id` must use the canonical `<name>:<version>` format and be unique within the local catalog. The implementation derives `Name`, `Alias`, and `Version` from that ID, removing the competing alias/version inputs and keeping local registrations consistent with catalog entries.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
selenayang888: Registration now directly calls `ModelInfoToJson()` and restores with `ModelInfoFromJson()` instead of flattening into a registration-specific property bag.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
selenayang888: Public C serialization, deserialization, and cloning operations were removed.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
selenayang888: `ModelInfo` is now move-only.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
selenayang888: Removed generation and `model_metadata.yml`; registration no longer writes into the model directory. The API reads the model's existing `<model_path>/genai_config.json` to validate it, and persists registrations only in `<app_data_dir>/catalogs/local/local_models.json`.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
selenayang888: The model directory and parseable config must exist, and registration no longer writes beside model assets.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
selenayang888: Recursive directory-size calculation was removed.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
selenayang888: Validation occurs before persistence, and transient post-save activation failures no longer erase durable registration.

sdk_v2/cpp/src/catalog/local_model_catalog.h
selenayang888: The unregister callback was eliminated by making unregistration a catalog operation, and the metadata-preparation callback was eliminated by requiring complete metadata at registration. Added explicitly names the remaining ModelFactory arguments as info, local_path, and runtime_model_id.

sdk_v2/cpp/src/manager.cc
selenayang888: Remove unnecessary manager `ListModels()` call.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
selenayang888: I removed the unregister delegate from `LocalModelCatalog::CreateModel()`. The model factory now receives only the model metadata, local path, and runtime model ID, so its responsibility is limited to constructing the model. Unregistration is now an explicit `Catalog::UnregisterModel()` operation handled by `LocalModelCatalog`, which validates that the model is unloaded, updates the persistent registration index, and retires the model from catalog lookup without deleting caller-owned assets.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
selenayang888: `SerializeToFile()` and `DeserializeFromFile()` were removed from the C++ wrapper.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
selenayang888: Friendship was removed and replaced with a narrow native-handle accessor.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
selenayang888: `GetLocalModels` API is removed; Callers use `GetModels()` on the selected catalog now.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
selenayang888: Updated it as suggested

sdk_v2/cpp/src/model.cc
selenayang888: I standardized the public local-model ID to `<alias>:<version>`. A private `local/<registration_id>` load-manager key remains to distinguish re-registration of the same ID while preserving outstanding handles. Removing it would require redesigning the model lifecycle. Would keeping identical public ID semantics with a private generation key address your concern?

sdk_v2/cpp/src/model.cc
selenayang888: Model can no longer be constructed without metadata: default construction is deleted, leaves initialize immutable metadata in their constructor, and containers require a metadata-bearing selected leaf. The old nullable current_info_ state is removed; the remaining guard only detects invalid access after an explicit move.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
selenayang888: Thanks @skottmckay and @baijumeswani. My understanding is that we agree the underlying public, local, and future private catalogs should remain distinct, but we have not agreed on the user-facing view or collision behavior.

For this PR, I propose:

- Keep `GetCatalog()` returning the existing public catalog to preserve current behavior.
- Expose the local registration catalog separately.
- Keep `RegisterModel()` and `UnregisterModel()` scoped to the local catalog.
- Do not add aggregation in this PR.
- Consider a read-only `GetAggregateCatalogView()` in a follow-up after defining provenance and collision semantics.

Could you confirm agreement on these two decisions?

1. Can this PR ship with separate public and local catalogs, with aggregation deferred to a follow-up?
2. For accessing the local catalog, should we keep `GetCatalog(CatalogType::Local)`, or use a narrower `GetLocalCatalog()` API? The explicit accessor avoids treating catalog kinds as a closed enum, which may not scale if multiple private catalogs are supported.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
selenayang888: We retained `UnregisterModel`/`UnregisterModel` pair and precisely describes removing the catalog registration without deleting the model assets. If there is a project-wide naming convention requiring `DeregisterModel`, I can apply it consistently across the C and C++ APIs before merge.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
selenayang888: I think we had a final decision(https://teams.microsoft.com/l/message/19:meeting_ZmVmYmNiYWEtMzBiMy00NGM1LTkxMmQtZTI4ODI5ZjY3M2E5@thread.v2/1787611366905?context=%7B%22contextType%22%3A%22chat%22%7D) after the discussion in a group chat, the decision is to have separate catalog API for now.

sdk_v2/cpp/src/model.cc
skottmckay: nit: the id is typically `<name>:<version>` as the name has EP/device details like 'generic-gpu' in it but `alias` does not, and when picking the latest version we need to do that on the per-name basis not per-alias.

the id should always be unique (at least within a catalog). can we use that instead of a separate registration_id?

not sure what you mean by "identical public ID semantics with a private generation key". Can you elaborate?

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,444 @@
...
+  resolved.SetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_MODEL_TYPE_STR, "ONNX");
+
+  const auto now = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+  resolved.SetPropertyInt(FOUNDRY_LOCAL_MODEL_PROP_CREATED_AT_UNIX_INT, now);
+  resolved.SetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_CREATION_TIME_STR, FormatUtcTimestamp(now));
+
+  return resolved;
+}
+
+std::vector<LocalModelCatalog::Registration> LocalModelCatalog::LoadRegistrations() const {
```
baijumeswani, line 342: If the function fails to read the local model info file (for any reason), it returns empty. Later, the registration will overwrite this file causing the index to be replaced. This makes the local index non-recoverable after that.

If the schema is unsupported, what should be the resolution? Maybe, we can document the behavior? Alternatively, distinguish not found from read failures. And for read failures, we should not mutate the index.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,444 @@
...
+        info.name = parsed_id.name;
+        info.version = parsed_id.version;
+        info.model_id = model_id;
+        info.uri.clear();
+
+        const auto duplicate = std::find_if(registrations.begin(), registrations.end(), [&](const auto& existing) {
+          return existing.info.model_id == info.model_id || existing.registration_id == registration_id;
+        });
+        if (duplicate == registrations.end()) {
+          registrations.push_back({std::move(info), model_path.string(), std::move(registration_id)});
```
baijumeswani: Do we validate the restored local model info to the same degree as fresh registrations? Ideally, both the registrations (whether new registrations or restored registrations) should share the model info validations.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.inline.h
```diff
@@ -204,9 +204,22 @@ inline ICatalog& Manager::GetCatalog() const {
...
+inline ICatalog& Manager::GetCatalog(CatalogType type) const {
+  if (type == CatalogType::Public) {
+    return GetCatalog();
+  }
+
+  std::call_once(*local_catalog_once_, [this, type]() {
+    flCatalog* catalog = nullptr;
+    Check(detail::api()->Manager_GetCatalogByType(handle_.get(), static_cast<flCatalogType>(type), &catalog));
+    catalogs_.local_ = std::make_unique<Catalog>(*catalog);
+  });
```
baijumeswani: What happens if a non-public, non-local type is passed in? We should reject the call or check before instantiating the local catalog.

On another note, why is this caching necessary? Can we not always call the C API to get the local catalog? Is there a perf implication to querying the local catalog from the C API instead of caching it here?

sdk_v2/python/src/foundry_local_sdk/_native/build_cffi.py
```diff
@@ -79,6 +79,11 @@
     FOUNDRY_LOCAL_DEVICE_NPU = 3,
 } flDeviceType;
 
+typedef enum flCatalogType {
```
baijumeswani, line 82: Any reason we need to port the python bindings changes in this PR? If yes, can you also update the api version being queried (from 1 to 2)?

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -0,0 +1,444 @@
...
+    }
+  }
+  if (!model) {
+    FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT, "model not found: " + alias_or_model_id);
+  }
+
+  model->BeginUnregister();
+  bool unregister_in_progress = true;
+  try {
+    for (const auto* variant : model->Variants()) {
```
baijumeswani: What happens if we try to register/unregister 2 local models (with the same alias/id) but different versions? Is that allowed? Will the system function correctly in that scenario?

sdk_v2/cpp/src/model.cc
selenayang888: Thanks. I applied the canonical ID `<name>:<version>`, while `alias` groups device/EP variants. I corrected the latest-version lookup to operate per `name`.

I also removed the separate `registration_id` and now use `model_id` end-to-end. Normalized paths are checked only to prevent same-ID load collisions and stale handles from affecting replacements; they are not a second identity.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
selenayang888: `LoadRegistrations()` now treats only a missing index as an empty catalog. File-inspection, open, JSON-parse, malformed-root, unsupported-schema, malformed-entry, and duplicate-ID failures throw before any mutation.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
selenayang888: Extracted shared `ValidateRegistrationMetadata()` validation for canonical model IDs, required tasks, and supported tasks. Both new registration and index restoration call it. Invalid restored metadata now blocks loading and subsequent mutation instead of being silently accepted or discarded.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.inline.h
selenayang888: `Manager::GetCatalog(CatalogType)` now explicitly accepts only Public and Local and throws `FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT` for any other value. Retained the cached C++ wrapper because the method returns a reference whose wrapper must live for the Manager lifetime; the underlying C catalog is already manager-owned and cached, so repeated C API calls would not provide a different catalog.

sdk_v2/python/src/foundry_local_sdk/_native/build_cffi.py
selenayang888: Retained the Python CFFI changes and changed `_FOUNDRY_LOCAL_API_VERSION` from 1 to 2, matching the v2 root function-table layout used by `build_cffi.py`.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
selenayang888: The canonical IDs remain unique while multiple versions can share a derived alias. Documented that unregistering by model ID removes only that version/variant, while unregistering by alias removes the entire alias group. Fixed model-ID unregistration so another loaded version does not block removal of the requested unloaded version.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -780,12 +805,23 @@ class ICatalog {
...
+  /// The catalog does not take ownership of `model_path` and never deletes its contents.
+  virtual std::unique_ptr<IModel> RegisterModel(const std::string&, const std::string&, const ModelInfo&) {
+    throw Error("models can only be registered in a local catalog", FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT);
+  }
+  /// Unregister without deleting model assets. A model ID removes only that version/variant; an alias removes all of
+  /// its registered versions and variants. Existing model wrappers and immutable metadata views remain valid, but
+  /// operations that require the retired registration, including Download and Load, fail with invalid usage.
+  virtual void UnregisterModel(const std::string&) {
+    throw Error("models can only be unregistered from a local catalog", FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT);
+  }
```
skottmckay: I would leave these as pure virtual methods here. 

It would make sense to have this error if this class implemented general catalog logic and we had a derived LocalCatalog class, but as a) this is defining the interface and b) the Catalog class is used for both I think this check should be internal to its implementation.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -840,6 +879,7 @@ class Manager {
 
   /// Get the catalog for querying models. Creates on first call, caches internally.
   ICatalog& GetCatalog() const;
+  ICatalog& GetCatalog(CatalogType type) const;
```
skottmckay:
```suggestion
  ICatalog& GetCatalog(flCatalogType type = flCatalogType::FOUNDRY_LOCAL_CATALOG_PUBLIC) const;
```

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -298,14 +301,31 @@ struct Runtime {
   std::optional<std::string_view> execution_provider;
 };
 
+enum class CatalogType {
+  Public = FOUNDRY_LOCAL_CATALOG_PUBLIC,
+  Local = FOUNDRY_LOCAL_CATALOG_LOCAL,
+};
```
skottmckay: nit: use flCatalogType instead of creating a duplicate enum.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -872,10 +912,16 @@ class Manager {
...
+    mutable std::unique_ptr<Catalog> public_;
+    mutable std::unique_ptr<Catalog> local_;
+  };
+
   detail::Base<flManager> handle_;
   Configuration config_;
-  mutable std::unique_ptr<Catalog> catalog_;
+  CatalogCollection catalogs_;
   mutable std::unique_ptr<std::once_flag> catalog_once_{std::make_unique<std::once_flag>()};
+  mutable std::unique_ptr<std::once_flag> local_catalog_once_{std::make_unique<std::once_flag>()};
```
skottmckay: nit: can we put these in CatalogCollection?

sdk_v2/cpp/src/c_api.cc
```diff
@@ -327,8 +328,9 @@ FL_API_STATUS_IMPL(Manager_CreateImpl, const flConfiguration* config, flManager*
   }
 
   auto& mgr = fl::Manager::Create(*cfg);
-  auto wrapper = std::make_unique<flManager>(flManager{mgr, nullptr, {}});
-  wrapper->catalog = std::make_unique<flCatalog>(flCatalog{mgr.GetCatalog()});
+  auto wrapper = std::make_unique<flManager>(flManager{mgr, nullptr, nullptr, {}});
+  wrapper->public_catalog = std::make_unique<flCatalog>(flCatalog{mgr.GetCatalog(fl::CatalogType::kPublic)});
+  wrapper->local_catalog = std::make_unique<flCatalog>(flCatalog{mgr.GetCatalog(fl::CatalogType::kLocal)});
```
skottmckay: Should local catalog creation be on-demand like we do at the C++ API level? 

If always having it is preferred, should the C++ API be consistent and create both on the first 'GetCatalog' call?

sdk_v2/python/src/foundry_local_sdk/_native/build_cffi.py
```diff
@@ -79,6 +79,11 @@
...
+    FOUNDRY_LOCAL_CATALOG_PUBLIC = 0,
+    FOUNDRY_LOCAL_CATALOG_LOCAL = 1,
+} flCatalogType;
```
skottmckay, line 85: Do we need something similar in the other language bindings?

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
selenayang888: Changed the public C++ `ICatalog::RegisterModel()` and `ICatalog::UnregisterModel()` methods to pure virtual. The concrete Catalog continues forwarding through the C API, where the internal public catalog rejects mutation and the local catalog implements it.

sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
selenayang888: Consolidated the overloads into `GetCatalog(flCatalogType type = FOUNDRY_LOCAL_CATALOG_PUBLIC)`. Default and explicitly public calls share one cached wrapper; local selection uses the same method.

sdk_v2/python/src/foundry_local_sdk/_native/build_cffi.py
selenayang888: I will work on the other language bindings in the follow-up PR based on this next.

sdk_v2/cpp/src/c_api.cc
selenayang888: Thanks, addressed. `Manager_CreateImpl()` no longer eagerly creates the public and local`flCatalog` wrappers. `Manager_GetCatalog()` and `Manager_GetCatalogByType()` now create the requested wrapper on first access and cache it for the manager’s lifetime. Each catalog has its own `std::once_flag`, so concurrent first access is safe and repeated calls return the same stable wrapper.

skottmckay (approved):

baijumeswani (approved):

PR #930, opened 2026-07-30
  Add samples/cs/rag/rag

Philippe-Laval:
Add the equivalent of ^python example found in the doc.


Philippe-Laval:
fix some problem reported

PR #951, opened 2026-08-05
Fix issue with language bindings caching info

skottmckay:
Don't cache current variant info at the language binding level as it can get stale for various reasons. 
Native level is source of truth.

### DIFF
--- a/sdk_v2/cs/test/FoundryLocal.Tests/CatalogTests.cs
+++ b/sdk_v2/cs/test/FoundryLocal.Tests/CatalogTests.cs
@@ -66,4 +66,44 @@ internal sealed class CatalogTests
         await Assert.That(async () => await catalog.ListModelsAsync(cts.Token).ConfigureAwait(false))
             .Throws<OperationCanceledException>();
     }
+
+    [Test]
+    public async Task SelectVariant_RefreshesReportedMetadata()
+    {
+        var catalog = await FoundryLocalManager.Instance.GetCatalogAsync();
+
+        var models = await catalog.ListModelsAsync();
+        var modelWithVariants = models.FirstOrDefault(m => m.Variants.Count > 1);
+
+        if (modelWithVariants == null)
+        {
+            // No multi-variant model in the catalog — nothing to exercise.
+            return;
+        }
+
+        var variants = modelWithVariants.Variants.ToList();
+        var defaultVariant = variants[0];
+        var otherVariant = variants.First(v => v.Id != defaultVariant.Id);
+
+        var infoBefore = modelWithVariants.Info;
+        await Assert.That(infoBefore.Id).IsEqualTo(defaultVariant.Id);
+
+        modelWithVariants.SelectVariant(otherVariant);
+
+        // Native is the source of truth: every read must reflect the selected variant.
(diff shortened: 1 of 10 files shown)


baijumeswani:
Another issue AI flagged that is not tied directly to the changes in this PR but AI seems to suggest that it makes the condition far more reachable now with these changes.


>There is an async native race condition. `selectVariant` writes `selected_variant_` with no lock while an async load / download may be reading it on a worker thread.
>
>`SelectVariant` and the readers `Info()` / `Id()` / `Alias()` / `IsCached()` / `IsLoaded()` / `GetPath()` all access `selected_variant_` without  `state_mutex_ ` yet `Variants()` does take the lock. That inconsistency signals the unguarded reads are an oversight, not a deliberate design.
>More seriously,  `GetPath()` reads `local_path_`  (a plain `std::string`) while `Download()` mutates it on a libuv worker thread. A genuine data race, not just a torn pointer.


Can be addressed separately though

sdk_v2/cs/test/FoundryLocal.Tests/CatalogTests.cs
```diff
@@ -66,4 +66,44 @@ public async Task ListModelsAsync_HonorsCancelledToken()
...
+
+    [Test]
+    public async Task SelectVariant_RefreshesReportedMetadata()
+    {
+        var catalog = await FoundryLocalManager.Instance.GetCatalogAsync();
+
+        var models = await catalog.ListModelsAsync();
+        var modelWithVariants = models.FirstOrDefault(m => m.Variants.Count > 1);
+
+        if (modelWithVariants == null)
```
baijumeswani, line 78: AI suggestion:

>C# test passes vacuously when no multi-variant model exists.
 `if (modelWithVariants == null) return;`  — a silent pass with no assertion and no skip signal. Python at least calls  pytest.skip(...) . Suggest C# emit an explicit inconclusive/skip so a catalog without multi-variant models doesn't masquerade as a green regression test.

skottmckay:
> Another issue AI flagged that is not tied directly to the changes in this PR but AI seems to suggest that it makes the condition far more reachable now with these changes.
> 
> > There is an async native race condition. `selectVariant` writes `selected_variant_` with no lock while an async load / download may be reading it on a worker thread.
> > `SelectVariant` and the readers `Info()` / `Id()` / `Alias()` / `IsCached()` / `IsLoaded()` / `GetPath()` all access `selected_variant_` without  `state_mutex_ ` yet `Variants()` does take the lock. That inconsistency signals the unguarded reads are an oversight, not a deliberate design.
> > More seriously,  `GetPath()` reads `local_path_`  (a plain `std::string`) while `Download()` mutates it on a libuv worker thread. A genuine data race, not just a torn pointer.
> 
> Can be addressed separately though

I believe I addressed the main issues by making the selected variant atomic and protecting/ordering updates to `local_path_`. 

I left the simple getters which are returning a single value from the currently variant as-is as locking felt like overkill there. `Variants()` is treated differently as it's creating a new list of values so what is returned is stable so needs a lock due to the multi-step process.

baijumeswani (approved):

PR #952, opened 2026-08-05
Add secure, transactional CUDA and WebGPU EP bootstrapping

baijumeswani:
## Summary

This change adds first-class CUDA execution-provider bootstrapping for Windows x64, Windows ARM64, and Linux x64. It also moves WebGPU onto the same reusable bundle installer so both providers follow one secure, transactional installation flow.

CUDA is treated as one compatible bundle assembled from independently reusable artifacts. The large CUDA and cuDNN packages change less frequently than the execution-provider package, so Foundry Local validates each installed artifact before downloading anything. If the CUDA or cuDNN runtime-file hashes are unchanged, those files are reused and only changed artifacts are downloaded. The reused and newly downloaded files are then assembled, verified, and atomically activated as one generation.

On Windows, provider dependencies are made discoverable at model-load time using `SetDllDirectoryW`. This avoids eagerly loading the full CUDA bundle while still supporting downstream bare-name DLL loads performed by ONNX Runtime GenAI. On Linux x64, `libonnxruntime-genai-cuda.so` is preloaded with `RTLD_NOW | RTLD_GLOBAL` and retained for the bootstrapper lifetime.

## Bundle installer responsibilities

The shared installer provides one implementation for:

- Archive and runtime-file SHA-256 validation.
- Selective artifact reuse and repair.
- Immutable bundle generations.
- Atomic activation and rollback.
- Cross-process synchronization.
- Cancellation and cleanup.
- Secure archive extraction.

## Architecture

The bootstrappers describe their bundles and handle provider-specific registration. The shared installer owns the package lifecycle.

```text
EpDetector::DownloadAndRegisterEps()
|
+-- CudaEpBootstrapper::DownloadAndRegister()
|   +-- NvmlGpuDetector::HasNvidiaGpu()
|   +-- BuildCudaEpManifest(HostCudaEpPlatform())
|   `-- EpBundleInstaller::EnsureInstalled(cuda_manifest)
|
`-- WebGpuEpBootstrapper::DownloadAndRegister()
    +-- BuildWebGpuManifest()
    `-- EpBundleInstaller::EnsureInstalled(webgpu_manifest)
```

### Shared bundle installation

```text
EpBundleInstaller::EnsureInstalled(manifest)
|
+-- ValidateManifest()
|   +-- Validate bundle and artifact identifiers
|   +-- Require HTTPS package URLs
|   +-- Validate archive and runtime-file SHA-256 values
|   `-- Validate all relative paths
|
+-- Acquire the provider's cross-process file lock
|
+-- Read the active-generation marker
|
+-- Clean stale staging and inactive generation directories
|   `-- Reject symlinked or otherwise unsafe managed directories
|
+-- If the active bundle verifies
|   `-- Reuse it without downloading
|
`-- Otherwise create a private staging generation
    |
    +-- For each artifact
    |   +-- Copy its files from the active generation when they still verify
    |   `-- Otherwise download, authenticate, extract, and verify it
    |
    +-- Verify the exact completed runtime-file set
    `-- Atomically move staging into an immutable bundle generation
```

### Activation and provider registration

The install transaction keeps the cross-process lock until registration succeeds or is rolled back.

```text
EpInstallTransaction::Activate()
|
+-- Reverify the completed generation while holding the provider lock
`-- Atomically publish the active-generation marker

Provider registration
|
+-- Register the provider library with ONNX Runtime
|
+-- Success
|   `-- EpInstallTransaction::Finalize()
|       `-- Remove stale generations
|
`-- Failure
    `-- EpInstallTransaction::Rollback()
        `-- Restore the previous active-generation marker
```

The transaction retains its logger, so destructor-triggered rollback can report recovery failures instead of failing silently.

### Provider registration and model loading

```text
Provider registration
|
+-- CUDA on Linux x64
|   +-- Load libonnxruntime-genai-cuda.so with RTLD_NOW | RTLD_GLOBAL
|   +-- Retain the handle transactionally
|   `-- Register CUDAExecutionProvider
|
+-- CUDA on Windows
|   `-- Register CUDAExecutionProvider by absolute provider path
|
`-- WebGPU
    `-- Register Foundry.WebGPU by absolute provider path


ModelLoadManager::LoadModel()
|
+-- Resolve the effective execution provider
|   +-- Explicit provider override, when supplied
|   +-- CUDA for generic-gpu when CUDA is registered
|   +-- WebGPU for generic-gpu when CUDA is unavailable and WebGPU is registered
|   `-- Model-ID requirement for provider-specific variants such as cuda-gpu
|
+-- Verify the required provider is registered
|
+-- EpDetector::PrepareForModelLoad(required_ep)
|   `-- Windows CUDA/WebGPU: SetDllDirectoryW(bundle_directory)
|
`-- Construct the GenAI model
```

`SetDllDirectoryW` is process-global, but model loads are serialized and the directory is selected immediately before GenAI model construction. This supports GenAI's bare-name DLL loads without preloading the CUDA and cuDNN bundle.

## CUDA platform support

| Platform | Package layout | Status |
|---|---|---|
| Windows x64 | CUDA toolkit, cuDNN, and CUDA EP archives | Supported |
| Windows ARM64 | CUDA toolkit, cuDNN, and CUDA EP archives | Supported |
| Linux x64 | CUDA EP and GenAI CUDA archive | Supported |
| Linux ARM64 | No published bundle | Unsupported |

## WebGPU behavior

WebGPU uses the shared bundle installer on its supported platforms.

Generic GPU models continue to prefer CUDA when CUDA is registered. If CUDA is unavailable and WebGPU is registered, the model uses WebGPU. The Windows WebGPU bundle directory is prepared with `SetDllDirectoryW` before model construction so delayed dependencies such as DXCompiler can be discovered.

WebGPU uses the same:

- Archive and runtime-file verification.
- Immutable generations and atomic activation.
- Transactional rollback.
- Cross-process synchronization.
- Cancellation and cleanup.
- Secure extraction.
- Authenticated metadata handling.

### Valid active bundle

Every declared runtime file is rehashed. If all hashes and the exact installed file set match, the existing generation is reused without downloading.

### Partially damaged bundle

Each artifact is checked independently. Valid artifact files are copied into a new staging generation, while only invalid artifacts are downloaded again. The combined generation is fully verified before activation.

### Forced installation

A forced request downloads every artifact instead of selectively reusing installed files.

### Concurrent processes

Each provider uses its own OS-backed cross-process file lock. Only one process can install and register a given EP at a time. Other processes wait for the lock and then reuse the verified active generation. Different providers use different locks and may install concurrently.

The lock is held through provider registration and transaction finalization. OS locks are released automatically when a process exits; the lock file remaining on disk does not indicate that the lock is still held.

### Failure or cancellation

Failed downloads, hash mismatches, malformed archives, cancellation, activation failures, and registration failures return failure for that provider. Temporary staging data is removed, and the previous active generation is restored when registration fails after activation.

`EpDetector` attempts the remaining requested providers after one provider fails. It stops early only when the caller cancels through the progress callback. The result reports both successfully registered and failed providers.

EP artifact downloads currently restart from the beginning after an interruption; resumable range downloads are not part of this change.

## Recovery behavior

The installer is designed to recover without manual cache deletion:

- Incomplete downloads and extraction remain in unique staging directories.
- Stale staging directories are removed by the next lock owner.
- A generation is published only after complete verification.
- Unpublished immutable generations are cleaned on the next installation attempt.
- A failed registration restores the previous active marker.
- A process crash releases the OS file lock automatically.

Manual cleanup should only be necessary for external filesystem failures such as broken permissions, antivirus interference, or directories that the operating system refuses to remove.

## Security and reliability

- HTTPS-only package manifests.
- Hard-coded SHA-256 values for archives and installed runtime files.
- Exact archive and installed file-set validation.
- Cross-process installation locks.
- Unique private staging directories.
- Immutable generation directories.
- Atomic active-marker replacement.
- Transactional activation and rollback.
- ZIP path-traversal protection.
- Symlink and special-file rejection.
- Duplicate-entry and file/directory collision rejection.
- Entry-count and uncompressed-size limits.
- HTTP body-size enforcement with or without `Content-Length`.
- Strict `Content-Length` parsing.
- Cancellation checks before provider registration.
- Logged destructor-triggered transaction rollback.
- WindowsStore-safe NVML behavior.
- No eager Windows CUDA or cuDNN DLL preload.

## NVIDIA GPU detection

CUDA eligibility is detected through NVML. Foundry Local requires at least one NVIDIA device with the supported compute capability. NVML is loaded dynamically and released after detection.


sdk_v2/cpp/src/manager.cc
```diff
@@ -339,31 +331,36 @@ Manager::~Manager() {
...
       if (status != nullptr) {
         const char* msg = ort_api_->GetErrorMessage(status);
-        logger_->Log(LogLevel::Warning,
-                     std::string("EP unregister: UnregisterExecutionProviderLibrary failed for '") +
-                         name + "': " + (msg ? msg : "unknown"));
+        try {
+          safe_log(LogLevel::Warning, std::string("EP unregister: UnregisterExecutionProviderLibrary failed for '") +
+                                          name + "': " + (msg ? msg : "unknown"));
+        } catch (...) {
+        }
```
skottmckay: nit: do you need another try/catch here if safe_log already has that? same for the "Exception while shutting down Manager subsystems during destruction" usage of safe_log. #Closed

sdk_v2/cpp/src/ep_detection/ep_bundle_manifest.h
```diff
@@ -0,0 +1,53 @@
...
+
+struct EpBundleArtifact {
+  std::string id;
+  std::string url;
+  bool is_archive = true;
+
+  std::string archive_sha256;
+  std::vector<EpBundleFile> extracted_files;
+  std::vector<std::string> ignored_archive_paths;
+  uint64_t archive_max_bytes = 0;
```
skottmckay, line 24: I assume we'll manage the manifests using AI, so if we have the SHA for each file could we also have the exact bytes for individual files, and infer the max values from that?

sdk_v2/cpp/src/http/http_download.h
```diff
@@ -3,28 +3,35 @@
...
+/// Supports progress reporting, cancellation, and an optional size cap.
 /// @param url  The URL to download from.
 /// @param destination  Local file path to write to.
 /// @param user_agent  HTTP User-Agent header.
 /// @param cancel_flag  Set to true to cancel. nullptr if not needed.
 /// @param progress_cb  Called with percent 0.0-100.0. Empty = no callback.
 /// @param logger  Logger for diagnostic output on failure.
+/// @param max_bytes  Fail closed if a Content-Length header exceeds this, and abort mid-stream
+///                   if the body exceeds it regardless of what Content-Length promised (defends
+///                   against a missing/incorrect header on chunked transfers). -1 means no cap.
```
skottmckay: Is this a known issue? are we downloading from an endpoint we don't control and we need extra protection? #Closed

sdk_v2/cpp/src/ep_detection/ep_bundle_installer.h
```diff
@@ -0,0 +1,84 @@
...
+  EpInstallTransaction& operator=(EpInstallTransaction&&) = delete;
+
+  const std::filesystem::path& bin_dir() const { return bin_dir_; }
+
+  const std::string& bundle_id() const { return manifest_.bundle_id; }
+
+  bool CommitActive(ILogger& logger);
+
+ private:
+  friend class EpBundleInstaller;
```
skottmckay: nit: using `friend` means you give up control of all your internals. as this is an internal only class that seems like a high price just to protect the ctor.  #Closed

sdk_v2/cpp/src/ep_detection/ep_utils.h
```diff
@@ -47,4 +22,32 @@ bool VerifyEpBinaries(
 /// @param dir Directory to prepend to `PATH`.
 void PrependDirToProcessPath(const std::filesystem::path& dir);
 
+/// Select the manifest-declared DLLs under @p bin_dir that should be preloaded before the EP provider
+/// library is registered, excluding the provider library itself and the core ORT runtime libraries
+/// (`onnxruntime.dll`, `onnxruntime-genai.dll`), matched case-insensitively.
+///
+/// This selection logic is platform-independent (pure path/string manipulation) so it can be unit
+/// tested on any platform, even though the actual preloading only happens on Windows.
```
skottmckay: what's the point of including/testing this on other platforms? #Closed

sdk_v2/cpp/src/ep_detection/ep_utils.cc
```diff
@@ -3,91 +3,106 @@
...
+#ifdef _WIN32
+  const auto dependencies = SelectEpBundleDependenciesToPreload(bin_dir, manifest);
+  handles_.reserve(handles_.size() + dependencies.size());
+
+  std::vector<void*> loaded;
+  loaded.reserve(dependencies.size());
+
+  for (const auto& path : dependencies) {
+    auto* handle =
+        LoadLibraryExW(path.c_str(), nullptr, LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR | LOAD_LIBRARY_SEARCH_SYSTEM32);
```
skottmckay: What's the reason to limit to LOAD_LIBRARY_SEARCH_SYSTEM32 instead of LOAD_LIBRARY_SEARCH_DEFAULT_DIRS? #Closed

sdk_v2/cpp/src/ep_detection/ep_bundle_installer.cc
```diff
@@ -0,0 +1,734 @@
...
+    logger.Log(LogLevel::Warning,
+               fmt::format("{}: refusing unsafe managed directory '{}'", ep_display_name, path.string()));
+    return false;
+  }
+
+  return true;
+}
+
+bool ValidateManagedDirectories(const std::filesystem::path& bundles_dir, const std::filesystem::path& staging_root,
+                                std::string_view ep_display_name, ILogger& logger) {
```
skottmckay, line 292: How do we recover from this failing? #Closed

sdk_v2/cpp/src/ep_detection/ep_bundle_installer.cc
```diff
@@ -0,0 +1,734 @@
...
+    logger.Log(LogLevel::Warning, fmt::format("{}: artifact '{}' archive entries do not match its manifest",
+                                              ep_display_name, artifact.id));
+    return false;
+  }
+
+  for (const auto& file : artifact.extracted_files) {
+    const auto file_path = extraction_dir / file.relative_path;
+    if (!std::filesystem::is_regular_file(file_path, ec)) {
+      logger.Log(LogLevel::Warning, fmt::format("{}: artifact '{}' is missing expected extracted file '{}'",
+                                                ep_display_name, artifact.id, file.relative_path));
```
skottmckay: nit: is this a potentially misleading message? #Closed

sdk_v2/cpp/src/ep_detection/ep_bundle_installer.cc
```diff
@@ -0,0 +1,734 @@
...
+    std::filesystem::path active_bin;
+    if (active_generation.has_value()) {
+      active_bin = bundles_dir / *active_generation / "bin";
+    }
+
+    if (policy == EpBundleInstallPolicy::ReuseVerified && !active_bin.empty() &&
+        VerifyBundleDir(active_bin, manifest, ep_display_name_, logger)) {
+      logger.Log(LogLevel::Information,
+                 fmt::format("{}: reusing verified bundle '{}'", ep_display_name_, manifest.bundle_id));
+      if (!ReportProgress(progress_cb, ep_display_name_, 90.0f)) {
```
skottmckay: Is 90% an arbitrary value for 'downloaded but not registered'? could we have a constant instead of a magic number to make that clearer if that's the case? #Closed

sdk_v2/cpp/src/ep_detection/nvml_gpu_detector.cc
```diff
@@ -0,0 +1,192 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#include "ep_detection/nvml_gpu_detector.h"
+
+#include <filesystem>
+
+#if defined(FOUNDRY_LOCAL_USE_WINHTTP_TRANSPORT)
```
skottmckay: Not clear how FOUNDRY_LOCAL_USE_WINHTTP_TRANSPORT relates to CUDA usage.  #Closed

sdk_v2/cpp/src/ep_detection/nvml_gpu_detector.cc
```diff
@@ -0,0 +1,192 @@
...
+    get_count_ = reinterpret_cast<NvmlDeviceGetCountFn>(GetSymbol(lib_, "nvmlDeviceGetCount_v2"));
+    get_handle_ = reinterpret_cast<NvmlDeviceGetHandleByIndexFn>(GetSymbol(lib_, "nvmlDeviceGetHandleByIndex_v2"));
+    get_compute_cap_ =
+        reinterpret_cast<NvmlDeviceGetCudaComputeCapabilityFn>(GetSymbol(lib_, "nvmlDeviceGetCudaComputeCapability"));
+
+    if (!init_ || !shutdown_ || !get_count_ || !get_handle_ || !get_compute_cap_) {
+      UnloadLibrary(lib_);
+      lib_ = kNullLibrary;
+      return;
+    }
```
skottmckay, line 121: should there be something logged if we fail this way? #Closed

sdk_v2/cpp/src/ep_detection/webgpu_ep_bootstrapper.cc
```diff
@@ -3,130 +3,108 @@
...
+constexpr const char* kProviderSha256 = "BE2EBCC0A96D1558D9123C04E75C2851260FE45C9DBC8959CB2CD8D11B83ABBE";
+constexpr const char* kDxCompilerSha256 = "174DBC3DF8F7AF5C32C0E39F43C0D5BC576395EDC3CCDD64119A1B63C081ED55";
+constexpr const char* kDxilSha256 = "080C02F62E90D0AB7ACC463BBC10280C37397DC6D036224D7C10F2ED9C20E13D";
+#elif defined(__APPLE__) && defined(__aarch64__)
+constexpr const char* kBundleId = "webgpu-ep-0.2.1-macos-arm64";
+constexpr const char* kDownloadUrl =
+    "https://foundrypackages-ffhrdhbxb7gpdreh.b02.azurefd.net/webgpu_ep_0.2.1_macos-arm64.zip";
+constexpr const char* kArchiveSha256 = "5F0F8378172F53EFA281328F33D1EBC793E44197FFB4CC605D02C593B891C0EA";
+constexpr const char* kProviderSha256 = "8FAC874A60F32F0127C74CB7DEF915807FCC8A6C30B77629E45F8CEE60272EAE";
 #endif
```
skottmckay: nit: do we need these vs. putting the values inline like we do in cuda_ep_manifest.cc? I found the latter a bit easier to read, but not sure if one way or the other is going to be easier to update. #Closed

sdk_v2/cpp/src/ep_detection/ep_utils.cc
```diff
@@ -3,91 +3,106 @@
...
-bool VerifyEpBinaries(
-    const std::filesystem::path& dir,
-    std::initializer_list<std::pair<std::string_view, std::string_view>> expected,
-    std::string_view ep_name,
-    ILogger& logger) {
-
-  for (const auto& [filename, expected_hash] : expected) {
-    auto file_path = dir / filename;
+std::vector<std::filesystem::path> SelectEpBundleDependenciesToPreload(const std::filesystem::path& bin_dir,
+                                                                       const EpBundleManifest& manifest) {
```
skottmckay: Why do we need to do this? Is this potentially loading files that may not be needed for a specific model? #Closed

sdk_v2/cpp/src/ep_detection/cuda_ep_bootstrapper.cc
```diff
@@ -112,89 +131,40 @@ bool CudaEpBootstrapper::DownloadAndRegister(bool force,
...
-    // Register with ORT
+    const auto provider_path = txn->bin_dir() / manifest->provider_relative_path;
 #ifdef _WIN32
-    // Permanently prepend the EP directory to PATH. The zip bundles all
-    // required CUDA/cuDNN DLLs, so no system CUDA install is needed.
-    // PATH must stay modified for the process lifetime because:
-    //   - onnxruntime_providers_cuda.dll delay-loads some dependencies
-    //   - onnxruntime-genai-cuda.dll is loaded later at model-load time
-    //   - ORT creates CUDA sessions after registration
-    PrependDirToProcessPath(ep_dir);
```
skottmckay, line 192: Something about the new setup is causing OpenCppCoverage to fail when the CUDA EP is registered. Not sure if it's this or something about the latest CUDA EP. Have you attempted to run code coverage? #Closed

sdk_v2/cpp/src/http/http_download.h
baijumeswani: The ep archives are hosted on endpoints controlled by us. The sha256 validation only happens post download. But if the url was wrong, or the endpoint compromised or misconfigured, this may protect users from downloading excessively large archives.

Will make clarifying code comment.

sdk_v2/cpp/src/ep_detection/cuda_ep_bootstrapper.cc
skottmckay: Seems like a transient issue with Windows. Went away with reboot. OpenCppCoverage FAQ mentions it and says it's non-deterministic

> This error happens when CREATE_PROCESS_DEBUG_INFO::hFile or LOAD_DLL_DEBUG_INFO::hFile is null.

sdk_v2/cpp/src/ep_detection/ep_bundle_installer.cc
baijumeswani: If this check fails, we stop the EP installation and registration attempt. Any previously downloaded CUDA installation remains can continue to be used. A future attempt will retry, but it will continue to fail if the managed folder structure is damaged or invalid.

We can add an explicit log message instructing users or applications to clear the affected EP cache directory and retry if the message below is vague. Or we could consider deleting the folder altogether and attempt a retry.

Failing hard here would prevent other eps from being registered and would make FL unusable.

sdk_v2/cpp/src/ep_detection/ep_bundle_installer.cc
baijumeswani: I'll update the message.

sdk_v2/cpp/src/ep_detection/ep_bundle_manifest.h
baijumeswani: According to AI:

```
A ZIP can be larger than the sum of the declared file sizes because:

 - Incompressible files may slightly expand when compressed.
 - ZIP headers and the central directory add overhead.
 - The archive contains ignored entries such as  version.json .
```

we could derive the max from total file sizes plus some overhead.

sdk_v2/cpp/src/ep_detection/ep_utils.cc
baijumeswani: We should not be preloading all binaries.

We need to preload `libonnxrutime-genai-cuda.so` so onnxruntime-genai can find it when it tries to load it.

On windows, we can do AddDllDirectory so cuda and cudnn and onnxruntime-genai-cuda.dll binaries be found.

sdk_v2/cpp/src/ep_detection/ep_utils.cc
baijumeswani: removed the LoadLibraryExW with the new changes.

sdk_v2/cpp/src/ep_detection/nvml_gpu_detector.cc
baijumeswani: from AI:

```
FOUNDRY_LOCAL_USE_WINHTTP_TRANSPORT means “this build uses WinHTTP for networking.” The NVML detector uses it as a proxy for “this is desktop Windows,” because CMake currently defines it only for desktop Windows and not Windows Store/UWP builds.

Windows Store/UWP applications cannot use the same native-loading approach as an unpackaged desktop application.

The detector currently:

 - Loads  nvml.dll  dynamically from  System32  or NVIDIA’s  Program Files  directory.
 - Uses desktop APIs such as  SHGetKnownFolderPath .
 - Relies on access to the machine-wide NVIDIA driver installation.

A UWP process is sandboxed and generally must load packaged native libraries through  LoadPackagedLibrary ; it cannot freely discover and load DLLs from arbitrary machine paths. The CUDA bootstrapper also downloads native DLLs at runtime, which does not fit the normal UWP package model.

So CUDA/NVML is intentionally disabled for WindowsStore builds today. Supporting it would require a separate packaged-library design, not simply changing the preprocessor condition
```

Will introduce `FOUNDRY_LOCAL_DESKTOP_WINDOWS`.

sdk_v2/cpp/src/ep_detection/ep_bundle_installer.cc
skottmckay: My concern was whether there was any path back to 'good' if it fails as we call it in `EpInstallTransaction::Activate`. 

If it's incredibly unlikely this would ever happen (seems that should be the case) I think a log message telling the user what needs to be cleaned up is ok. We can't delete the 'directory' can we as it's no longer a directory we have access to is it?

sdk_v2/cpp/src/ep_detection/ep_bundle_installer.h
```diff
@@ -0,0 +1,89 @@
...
+  EpInstallTransaction& operator=(EpInstallTransaction&&) = delete;
+
+  const std::filesystem::path& bin_dir() const { return bin_dir_; }
+  std::filesystem::path provider_path() const { return bin_dir_ / manifest_.provider_relative_path; }
+
+  const std::string& bundle_id() const { return manifest_.bundle_id; }
+
+  bool Activate(ILogger& logger);
+  void Finalize(ILogger& logger);
+  bool Rollback(ILogger& logger) noexcept;
```
skottmckay: nit: can we pass the logger into the ctor so we use the same one for the entire lifetime?  #Closed

skottmckay (dismissed):

sdk_v2/cpp/src/ep_detection/ep_utils.cc
baijumeswani: After trying out `AddDllDirectory` on windows, it did not work. I reverted to preloading all bins in the ep dir.

The problem is that genai loads `onnxruntime-genai-cuda.dll` and onnxruntime loads the `cudnn` and `cuda` bins by name. AddDllDirectory does not allow these bins to be discoverable by genai and ort if they are not loaded with some explicit flags. So, if we don't preload, we get the `126` error.

There are three other options:
1. SetDllDirectoryW: would work (it is what we did with FLC C#), but it's process-global (ep dirs having same bins would collide). Moreover, it touches the global state and AI seems to think is less sage than preloading.
2. SetDefaultDllDirectories: AddDllDirectory only applies to loads with specific flags (or after  SetDefaultDllDirectories). But this is again process-global, and wipes out the PATH.
3. Add to PATH: again process global.

On the point of files not being needed, AI thinks:

```
LoadLibraryEx  just memory-maps the DLL — it maps the file into    address space but doesn't read it into physical memory. The OS pulls in only the pages that    actually get touched, on demand. And CUDA libs defer their heavy init to the first real    API call. So preloading a cudnn a particular model never exercises costs almost nothing.
```

I can change to any of the other three options if that is preferable.

baijumeswani: SetDllDirectoryW was better. Using that now.

skottmckay (approved):
:shipit:

PR #965, opened 2026-08-07
Validate packaged SDKs with real consumer samples

baijumeswani:
## Summary

This change fixes the JavaScript SDK package installation failure and adds post-packaging validation for the C++, C#, Python, and JavaScript SDKs.

The JavaScript package install script requires `deps_versions.json` to determine which ONNX Runtime dependencies to install. The packaging pipeline previously ran `npm pack` without staging that file into the JavaScript package root, causing consumers to receive an installation error. The package now stages the shared dependency manifest automatically through npm's `prepack` lifecycle.

Each SDK packaging flow now has a Windows x64 consumer test that uses the generated package rather than source build outputs. The test installs or extracts the package, runs a checked-in basic chat sample with a clean model cache, downloads a CPU chat model, loads it, and performs inference.

### DIFF
--- a/.pipelines/v2/templates/stages-cs.yml
+++ b/.pipelines/v2/templates/stages-cs.yml
@@ -279,9 +279,11 @@ stages:
           }
           dotnet restore $project `
             --configfile (Join-Path $consumerDir 'NuGet.config') `
+            --runtime win-x64 `
             /p:FoundryLocalPackageVersion=$version
           if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
-          dotnet run --project $project --no-restore /p:FoundryLocalPackageVersion=$version
+          dotnet run --project $project --no-restore --runtime win-x64 `
+            /p:FoundryLocalPackageVersion=$version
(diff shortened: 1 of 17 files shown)


.pipelines/v2/templates/stages-cs.yml
```diff
@@ -198,3 +198,91 @@ stages:
...
+          $packages | ForEach-Object { Copy-Item $_.FullName -Destination $feedDir -Force }
+
+          $version = (Get-Content '$(Pipeline.Workspace)/version-info/sdkVersion.txt' -Raw).Trim()
+          @"
+          <?xml version="1.0" encoding="utf-8"?>
+          <configuration>
+            <packageSources>
+              <clear />
+              <add key="LocalPackages" value="$feedDir" />
+              <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
prathikr: nuget.org may not be allowed in CIs long term.

baijumeswani: will change to lotus or aifoundrylocal

prathikr (dismissed):

prathikr (approved):

PR #973, opened 2026-08-10
Add Harmony (GPT-OSS) Tool Calling Support in FL

sayanshaw24:
# Add Harmony (GPT-OSS) Tool Calling Support in FL

## Summary

Adds data-driven tool calling support for GPT-OSS (Harmony) models, which use a fundamentally different protocol from ChatML-style models (Phi, Qwen). No model-specific code branches — all format knowledge is captured in a `ToolCallConfig` struct that the accumulator interprets at runtime.

## Problem

ChatML models emit tool calls as:
```
<tool_call>{"name": "func", "arguments": {...}}</tool_call>
```

GPT-OSS models use header-based routing:
```
<|start|>assistant to=functions.get_weather<|channel|>default<|message|>{"city": "Seattle"}<|call|>
```

Key differences:
- Function name is in the **header** (`to=functions.X`), not in the JSON body
- Arguments are the raw body (no wrapping `{"name": ..., "arguments": ...}` object)
- `<|call|>` is both EOT and EOS (generation stops naturally)
- Parallel calls use `<|end|>` for intermediate calls, `<|call|>` for the final one

### New: `tool_call_config.h`
- `ToolCallConfig` struct with `Mode::kSimple` (ChatML) and `Mode::kHeaderInspection` (Harmony)
- `Harmony()` preset: message_token, header_regex, channel_token, end_tokens
- `FromModelType(model_type)` factory: `"gptoss"` → Harmony, everything else → Simple

### Modified: `tool_call_stream_accumulator.h`
- New constructor accepting `ToolCallConfig`
- Header-inspection state machine: `Idle → InHeader → InBody → Emit`
- Regex-based function name extraction from header
- Multiple end tokens supported (`<|end|>` + `<|call|>` for parallel calls)
- Simple mode path completely unchanged

### Modified: `chat_session.cc`
- Both accumulator creation sites (Responses API + ChatCompletions) now detect model type via `model_.GetGenAIConfig().model->type` and pass the appropriate `ToolCallConfig`

### New tests: `tool_call_stream_accumulator_test.cc`
- `HarmonyBasicToolCall` — single tool call in one chunk
- `HarmonyToolCallSplitAcrossChunks` — streaming token-by-token
- `HarmonyRegularTextNotToolCall` — header without `to=functions.X` is visible text
- `HarmonyParallelToolCalls` — two calls, first ends with `<|end|>`, second with `<|call|>`
- `HarmonyTextBeforeToolCall` — preamble text preserved
- `HarmonyFlushUnterminatedHeader` — graceful EOS mid-header
- `HarmonyFlushUnterminatedBody` — graceful EOS mid-body

## Design Principles

- **Data-driven**: `ToolCallConfig` captures all protocol knowledge; no if/else on model name in the accumulator
- **Backward compatible**: existing models use `Mode::kSimple` (default) — zero behavior change
- **Forward compatible**: adding a new protocol = new `Mode` enum + preset; adding a new model with existing protocol = one line in `FromModelType`
- **No EOS suppression needed**: Harmony's `<|end|>` (not EOS) vs `<|call|>` (EOS) design naturally supports parallel calls

## Depends On

- FL PR #838 (tool/reasoning metadata migration) — for `bot_token_id`/`eot_token_id` resolution from genai_config.json
- GenAI 0.15.1+ with bot/eot APIs

## Testing

- 18/18 accumulator unit tests pass (11 existing + 7 new Harmony tests)
- E2E validation with `gpt-oss-20b-generic-cpu` model pending


sdk_v2/cpp/src/inferencing/generative/toolcalling/tool_call_config.h
```diff
@@ -0,0 +1,68 @@
...
+    cfg.message_token = "<|message|>";
+    cfg.header_regex = R"(to=functions\.(.+))";
+    cfg.channel_token = "<|channel|>";
+    cfg.end_tokens = {"<|end|>", "<|call|>"};
+    return cfg;
+  }
+
+  /// Infer config from model type string (from genai_config.json "model.type").
+  static ToolCallConfig FromModelType(const std::string& model_type) {
+    if (model_type == "gptoss") return Harmony();
```
justinchuby, line 63: Is there a way for this to be declared or detected so we don't need to hard code?

sdk_v2/cpp/src/inferencing/generative/toolcalling/tool_call_config.h
```diff
@@ -0,0 +1,68 @@
...
+    cfg.mode = Mode::kSimple;
+    return cfg;
+  }
+
+  /// Config for GPT-OSS (Harmony) models.
+  static ToolCallConfig Harmony() {
+    ToolCallConfig cfg;
+    cfg.mode = Mode::kHeaderInspection;
+    cfg.message_token = "<|message|>";
+    cfg.header_regex = R"(to=functions\.(.+))";
```
justinchuby, line 55: Are functions always starting with functions.? I was testing the muse glimmer model and it was reported that the tools were not provided with a name space. I wonder if the normalization should be done during tool declaration or somewhere else

sdk_v2/cpp/src/inferencing/generative/toolcalling/tool_call_config.h
sayanshaw24: The model.type field in genai_config.json already identifies the model family (e.g., "gptoss", "qwen2", "phi3"), and this is what FromModelType reads, so it's config-driven rather than truly hardcoded. The factory is just a mapping from that config value to the right protocol preset.

Longer term, we could make this fully declarative by adding a tool_call_format field to genai_config.json (e.g., "tool_call_format": "header_inspection") so the model package itself declares its protocol. For now, inferring from model.type avoids requiring config changes to existing model packages.

sdk_v2/cpp/src/inferencing/generative/toolcalling/tool_call_config.h
sayanshaw24: The to=functions. prefix is the convention used by the GPT-OSS/Harmony chat template - it's how the model was trained to emit tool calls. The regex to=functions\.(.+) captures the function name after that prefix.

For models like Muse Glimmer that don't use a namespace, we'd add a different ToolCallConfig preset with a regex that matches their header format (e.g., to=(.+) without the functions. prefix). That's exactly what the data-driven design enables: different presets for different protocols, no accumulator code changes needed.

The normalization question (should functions be stripped at declaration time vs parse time) - we handle it at parse time in the regex, which keeps tool definitions clean and model-agnostic.

sdk_v2/cpp/src/inferencing/generative/toolcalling/tool_call_config.h
justinchuby: SG. Eventually I would like to push us off of conditioning on model family for anything.

PR #980, opened 2026-08-11
Fix NvTensorRTRTX EP loading on Windows out-of-box

baijumeswani:
`*-trtrtx-gpu` models depend on `onnxruntime-genai-cuda.dll` and fail to load with error 126, unless a CUDA model had already been loaded in the same process or the cuda-ep bundle dir was manually added to `PATH`.

This PR preloads the `onnxruntime-genai-cuda.dll` so that other eps can consume it when loading their models.

In addition, cuda-ep is auto injected into download ep requests when user requested for nvtensorrtrtx ep.

### DIFF
--- a/sdk_v2/cpp/src/ep_detection/ep_detector.cc
+++ b/sdk_v2/cpp/src/ep_detection/ep_detector.cc
@@ -134,6 +134,27 @@ EpDownloadResult EpDetector::DownloadAndRegisterEps(const std::vector<std::strin
   EpDownloadResult result;
   result.success = true;
 
+  // Expand the requested set for EPs that depend on another EP's runtime. NvTensorRTRTX (a WinML EP)
+  // reuses the GenAI CUDA bridge that ships in the CUDA EP bundle, so requesting it by name must also
+  // register the CUDA EP. Only applies when a CUDA bootstrapper exists (i.e. an NVIDIA GPU is present).
+  std::vector<std::string> expanded_names;
+  if (names != nullptr) {
+    constexpr const char* kTrtRtxEp = "NvTensorRTRTXExecutionProvider";
+    constexpr const char* kCudaEp = "CUDAExecutionProvider";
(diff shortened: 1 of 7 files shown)


sdk_v2/cpp/src/ep_detection/ep_detector.cc
```diff
@@ -134,6 +134,32 @@ EpDownloadResult EpDetector::DownloadAndRegisterEps(const std::vector<std::strin
...
+    constexpr const char* kCudaEp = "CUDAExecutionProvider";
+    const bool wants_trtrtx = std::find(names->begin(), names->end(), kTrtRtxEp) != names->end();
+    const bool cuda_requested = std::find(names->begin(), names->end(), kCudaEp) != names->end();
+    const bool has_trtrtx_bootstrapper =
+        std::any_of(bootstrappers_.begin(), bootstrappers_.end(),
+                    [&](const auto& bs) { return bs->Name() == kTrtRtxEp; });
+    const bool has_cuda_bootstrapper =
+        std::any_of(bootstrappers_.begin(), bootstrappers_.end(),
+                    [&](const auto& bs) { return bs->Name() == kCudaEp; });
+    if (wants_trtrtx && has_trtrtx_bootstrapper && has_cuda_bootstrapper && !cuda_requested) {
```
skottmckay: Is it an error if the user wants TRT-RTX but the CUDA bootstrapper isn't available?

baijumeswani: This should be an impossible scenario because cuda ep is enabled on all devices where we detect nvidia gpu whereas nvtensorrtrtx ep is only available on a subset of those devices.

baijumeswani: But I can add a defensive guard against that just in case.

skottmckay (approved):

PR #981, opened 2026-08-11
Add model package support for sdkv2

xiaoyu-work:
## Summary

Add support for **ORT model packages** in the C++ SDK — catalog, download, cache discovery, and loading — alongside the existing flat model directories.

A model package is a directory that bundles several compiled variants of the same model behind a top-level `manifest.json`. ORT GenAI loads the directory directly and picks the variant matching the local hardware.

Stage 1 scope: one package per execution provider, single component, multiple variants. Design notes: `sdk_v2/cpp/docs/ModelPackageIntegration.md`.

## Changes

- **Layout detection** — new `ClassifyModelLayout()` in `src/util/model_layout.{h,cc}`. A root `genai_config.json` means flat; `manifest.json` without it means package.
- **Loading** — `model_load_manager` parses `manifest.json` and merges the per-variant `genai_config.json` files conservatively. `GenAIModelInstance` uses `OgaConfig::CreateFromPackageEp()` so GenAI does the variant selection.
- **Catalog** — parse `variantMetadata.modelPackage`, add `ModelInfo::IsModelPackage()`, and hide entries whose package variants are all unsupported on the current machine (fails open when compatibility is unknown).
- **EP detection** — new `EpDetector::GetModelCompatibilityForEpDevices()` to evaluate a variant's compatibility string against registered EP devices.
- **Download / cache** — keep the `inference_model.json` marker at the package root, and treat a package as a leaf when scanning the cache so variants don't surface as separate models.
- **`Model::Load`** — for packages only, a default load adopts the catalog EP so the right variant is selected.


xiaoyu-work:
@copilot resolve the merge conflicts in this pull request

PR #982, opened 2026-08-12
throw when audio item is passed to chat session

prathikr:
fixes https://github.com/microsoft/foundry-local-v2-bugbash/issues/6

### DIFF
--- a/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
@@ -223,8 +223,6 @@ struct TextSegment {
 //
 // - Text outside the markers is emitted as DEFAULT (visible) segments.
 // - Text inside the markers is emitted as REASONING segments. The markers themselves are stripped.
-// - Truncated reasoning (no closing marker) is treated as a REASONING segment running to end of input.
-// - Empty segments are skipped.
 // - When start_marker is empty, the entire input is returned as a single DEFAULT segment.
 //
 // Leading whitespace/newlines on visible segments that immediately follow a closed reasoning block are trimmed —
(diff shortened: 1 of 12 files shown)


sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_generator.h
```diff
@@ -123,6 +123,8 @@ class OnnxChatGenerator : public ChatGenerator {
                                                        bool use_full_context,
                                                        const std::vector<const ImageItem*>& images);
 
+  static void ValidateNoAudioInput(const std::vector<MessageItem>& messages);
```
baijumeswani: Does this need to be a static void? Could it be in the anonymous namespace in the .cc file?

baijumeswani (dismissed):

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_generator.cc
```diff
@@ -141,6 +141,7 @@ int OnnxChatGenerator::AppendMessages(const std::vector<MessageItem>& new_messag
   if (new_messages.empty()) {
     FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL, "new_messages must not be empty");
   }
+  ValidateNoAudioInput(new_messages);
```
natke: should we also validate no image input?

natke: already check it

prathikr: we already check for vision input, only audio gets through

natke (dismissed):

baijumeswani (dismissed):

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -413,6 +413,28 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
...
+    for (const auto& part : message.content) {
+      if (!part.view) {
+        continue;
+      }
+
+      // Check whether any advertised model input has the same item type as this message part.
+      const bool supported = std::any_of(io_info.inputs, io_info.inputs + io_info.num_inputs,
+                                         [&part](const Item* input) {
+                                           return input->type == part.view->type;
+                                         });
```
skottmckay: Could we add a generic input validation function to Session that checks the Request.items so all session types get that by default? Do that in Session::ProcessRequest before calling ProcessRequestImpl.

baijumeswani (approved):

PR #989, opened 2026-08-12
Validate temperature search option range

baijumeswani:
Reject temperature values outside of [0.0, 2.0]


bmehta001 (approved):

sdk_v2/cpp/src/inferencing/generative/chat/search_options.cc
```diff
@@ -53,7 +53,12 @@ int ApplySearchOptions(const SearchOptions& options,
 
   // Temperature
   if (options.temperature.has_value()) {
-    gen_params.SetSearchOption("temperature", static_cast<double>(*options.temperature));
+    const float temperature = *options.temperature;
```
bmehta001, line 56: Does the check here cover setting the temperature later, like through `SetOptions(options);`?

bmehta001 (changes requested):
Maybe `SearchOptions::FromParameters` needs a similar check?

bmehta001 (approved):

PR #991, opened 2026-08-12
Fixes from RC testing

skottmckay:
Implement CancelAll so shutdown doesn't deadlock.

### DIFF
--- a/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
@@ -377,6 +377,26 @@ void ChatSession::ProcessGeneratedOutput(std::string text, const ToolCallContext
                           total_tokens, prompt_tokens, completion_tokens));
 }
 
+void ChatSession::ClampMaxOutputToCatalogCeiling(SearchOptions& options, bool vision_turn) const {
+  const auto* max_output_ceiling =
+      CatalogModel().Info().GetPropertyInt(FOUNDRY_LOCAL_MODEL_PROP_MAX_OUTPUT_TOKENS_INT);
+  if (!max_output_ceiling || *max_output_ceiling <= 0) {
+    return;
+  }
+
+  // The catalog value is a ceiling, not a recommended length. A supplied value is capped to it; an omitted
+  // value keeps the conservative modality default (2048 text / 3072 vision, applied later in
+  // ApplySearchOptions) unless that default would itself exceed the ceiling, in which case we pin down.
+  const int ceiling = static_cast<int>(*max_output_ceiling);
+  const int modality_default = vision_turn ? 3072 : 2048;
+
+  if (options.max_output_tokens.has_value()) {
+    options.max_output_tokens = std::min(*options.max_output_tokens, ceiling);
+  } else if (ceiling < modality_default) {
+    options.max_output_tokens = ceiling;
+  }
+}
+
 void ChatSession::ProcessRequestImpl(const Request& request, Response& response) {
   // OpenAI chat completions JSON pass-through: a TEXT item tagged OPENAI_JSON. Routes to a separate handler that
   // never uses the cached generator or history (the JSON payload is self-contained).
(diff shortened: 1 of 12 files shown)


sdk_v2/cpp/src/service/audio_transcriptions_handler.cc
```diff
@@ -191,9 +191,11 @@ std::shared_ptr<HttpRequestHandler::OutgoingResponse> AudioTranscriptionsHandler
   std::thread streaming_thread([bg_session = std::move(session), body_ptr, &logger,
                                 req = std::move(session_request), &tracker,
                                 &session_manager = ctx_.session_manager]() mutable {
-    SessionRegistration reg(session_manager, bg_session);
-
     try {
+      // Register inside the try so a shutdown rejection (Register throws) is reported as a stream error
+      // instead of escaping this raw std::thread and calling std::terminate.
+      SessionRegistration reg(session_manager, bg_session);
```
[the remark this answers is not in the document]
skottmckay: fixed in web_service.cc to run connection_handler->stop() before JoinAll() (your option b), so no handler can Track() after the snapshot. Safe from deadlock since SseStreamBody::Push() never blocks and CancelAll() (run first) unwinds in-flight generation; the try-placement here stays for the separate case where Register() throws inside the thread.

sdk_v2/cpp/src/service/responses_handler.cc
```diff
@@ -477,6 +475,10 @@ std::shared_ptr<HttpRequestHandler::OutgoingResponse> ResponsesHandler::HandleSt
     };
 
     try {
+      // Register inside the try so a shutdown rejection (Register throws) is reported as a stream failure
+      // instead of escaping this raw std::thread and calling std::terminate.
+      SessionRegistration reg(session_manager, *session);
```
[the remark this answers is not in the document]
skottmckay: fixed in web_service.cc

sdk_v2/cpp/src/service/chat_completions_handler.cc
```diff
@@ -201,9 +201,11 @@ std::shared_ptr<HttpRequestHandler::OutgoingResponse> ChatCompletionsHandler::Ha
                                 req = std::move(session_request),
                                 include_usage, &tracker,
                                 &session_manager = ctx_.session_manager]() mutable {
-    SessionRegistration reg(session_manager, bg_session);
-
     try {
+      // Register inside the try so a shutdown rejection (Register throws) is reported as a stream error
+      // instead of escaping this raw std::thread and calling std::terminate.
+      SessionRegistration reg(session_manager, bg_session);
```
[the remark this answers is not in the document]
skottmckay: fixed in web_service.cc

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -376,6 +377,26 @@ void ChatSession::ProcessGeneratedOutput(std::string text, const ToolCallContext
                           total_tokens, prompt_tokens, completion_tokens));
 }
 
+void ChatSession::ClampMaxOutputToCatalogCeiling(SearchOptions& options, bool vision_turn) const {
+  const auto* max_output_ceiling =
+      CatalogModel().Info().GetPropertyInt(FOUNDRY_LOCAL_MODEL_PROP_MAX_OUTPUT_TOKENS_INT);
```
baijumeswani: Do we have any models with the max output tokens defined?

baijumeswani: Catalog v2 migration may impact this property. Should we leave it out from here? Treat the genai_config.json's max_length as authorities ceiling with the defaults in foundry-local as they currently are?

baijumeswani (dismissed):

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
skottmckay: Based on model info in the cache the vast majority define 2048. I think that conflicts with allowing in the 3072 though so this clamp would prevent that.

6 embeddings models define a max of 1. 

I'll remove and add a note that the genai_config.json is the source of truth so AI doesn't suggest this update again in the future.

baijumeswani: My guess is that some users may want to see the context length of the model from the catalog before they download the model. Which may be the reason we added it as a tag. I am not sure if this context length is ported to catalog v2 though. Regardless, i think the default max_tokens should be defined in foundry local with the cap coming from genai_config.json

baijumeswani (approved):

PR #996, opened 2026-08-13
GetCachedModels to return valid and accurate entries

baijumeswani:
This PR makes cached-model discovery resilient to catalog outages and returns every cached model variant instead of one selected model per alias.

- Falls back to `foundry.modelinfo.json` when all live catalog URLs fail.
- Treats valid disk models missing from metadata as BYOM models.
- Returns cached leaf variants in alias order and per-alias best-first order.

### Issues addressed

1. **Offline catalog returned no cached models**
   - Before: if every live catalog request failed, catalog population returned an empty list even when models and `foundry.modelinfo.json` were present locally.
   - Now: when all live catalog URLs fail, the local snapshot supplies model metadata.

2. **Locally valid models missing from metadata were omitted**
   - Before: a downloaded model could be invisible when neither the live response nor snapshot contained its ID.
   - Now: valid models found by the disk scan are included with BYOM metadata.

3. **Only one cached variant per alias was returned**
   - Before: `GetCachedModels()` returned the alias container, which exposed only its selected CPU/GPU/NPU variant.
   - Now: it returns every cached leaf variant under the alias and skips only uncached variants.

4. **Cache-only mode did not reconcile snapshot metadata with disk state**
   - Before: snapshot entries were created without scanned local paths, and disk-only models were not represented.
   - Now: cache-only mode still performs no live requests, but attaches scanned paths and adds unmatched models as BYOM entries.

5. **Multiple catalog URLs could produce duplicate model IDs**
   - Before: each URL independently resolved local IDs and could synthesize duplicate entries.
   - Now: successful live results are aggregated, deduplicated by model ID, and merged with local models once.

6. **Cached-model results could hide the canonical variant objects**
   - Before: callers received alias containers rather than the leaf returned by `GetModelVariant(model_id)`.
   - Now: both APIs return the same catalog-owned leaf pointer, keeping download, load, removal, and path state consistent.

### Code flow

```text
GetCachedModels()
       |
       v
Scan local model cache
       |
       +--> model IDs + local paths
       |
       v
Is catalog cache-only?
       |
   +---+---+
   |       |
  Yes      No
   |       |
   |       v
   |   Query configured live catalog URL(s)
   |       |
   |   +---+------------------+
   |   |                      |
   |   | Any URL succeeded    | All URLs failed
   |   v                      v
   | Aggregate + deduplicate  Load foundry.modelinfo.json
   | live metadata            (missing/invalid = empty)
   |   |                      |
   +---+----------------------+
       |
       v
Merge selected metadata with disk scan
       |
       +--> attach local paths by model ID
       +--> synthesize BYOM metadata for unmatched disk models
       |
       v
Save snapshot only when metadata came from live catalog
       |
       v
Group models by alias
       |
       v
Flatten and return every cached leaf variant
```

### DIFF
--- a/sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
+++ b/sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
@@ -767,7 +767,7 @@ class ICatalog {
   virtual std::string_view GetName() const = 0;
   virtual ModelList GetModels() const = 0;
 
-  /// Get every leaf model variant currently present in the local cache.
+  /// Get every individual model variant currently present in the local cache.
   virtual ModelList GetCachedModels() const = 0;
 
   virtual ModelList GetLoadedModels() const = 0;
--- a/sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
+++ b/sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
@@ -766,7 +766,10 @@ class ICatalog {
   virtual ~ICatalog() = default;
   virtual std::string_view GetName() const = 0;
   virtual ModelList GetModels() const = 0;
+
+  /// Get every leaf model variant currently present in the local cache.
   virtual ModelList GetCachedModels() const = 0;
+
   virtual ModelList GetLoadedModels() const = 0;
   virtual std::unique_ptr<IModel> GetModel(const std::string& alias) const = 0;
   virtual std::unique_ptr<IModel> GetModelVariant(const std::string& model_id) const = 0;
--- a/sdk_v2/cpp/src/catalog/azure_model_catalog.cc
+++ b/sdk_v2/cpp/src/catalog/azure_model_catalog.cc
@@ -12,9 +12,43 @@
 #include <fmt/format.h>
 
 #include <algorithm>
+#include <iterator>
+#include <unordered_set>
 #include <utility>
 
 namespace fl {
+
+namespace {
+
+ModelInfo MakeByomModelInfo(const std::string& model_id) {
+  auto [name, version] = Utils::SplitModelNameAndVersion(model_id);
+
+  ModelInfo info;
+  info.model_id = model_id;
+  info.name = name;
+  info.alias = name;
+  info.uri = "local://" + name;
+  info.version = version;
+  info.string_properties[FOUNDRY_LOCAL_MODEL_PROP_MODEL_PROVIDER_STR] = "Local";
+  info.string_properties[FOUNDRY_LOCAL_MODEL_PROP_MODEL_TYPE_STR] = "ONNX";
+  return info;
+}
+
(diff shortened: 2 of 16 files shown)


sdk_v2/cpp/include/foundry_local/foundry_local_cpp.h
```diff
@@ -766,7 +766,10 @@ class ICatalog {
   virtual ~ICatalog() = default;
   virtual std::string_view GetName() const = 0;
   virtual ModelList GetModels() const = 0;
+
+  /// Get every leaf model variant currently present in the local cache.
```
skottmckay: 'leaf' isn't a concept a C++ API user knows anything about.

sdk_v2/cpp/src/catalog/azure_model_catalog.cc
```diff
@@ -44,92 +78,95 @@ AzureModelCatalog::AzureModelCatalog(std::vector<std::pair<std::string, std::opt
...
-    try {
-      fetch_from(url, filter);
-    } catch (const std::exception& ex) {
-      // One failing URL shouldn't block others — skip and continue.
-      logger_.Log(LogLevel::Error,
-                  fmt::format("failed to fetch catalog from {}: {}", url, ex.what()));
-    }
-  }
+  auto catalog_result = GetLiveCatalogOrLocalSnapshot(cached_model_ids);
+  auto models = AddLocalModels(catalog_result.model_infos, local_models);
```
skottmckay, line 165: If catalog_result is from a snapshot do we need to call AddLocalModels?

I guess a new local model could have been added post-snapshot. Most likely it's just a bunch of de-duping to do from the call.

sdk_v2/cpp/src/catalog/azure_model_catalog.cc
```diff
@@ -44,92 +78,95 @@ AzureModelCatalog::AzureModelCatalog(std::vector<std::pair<std::string, std::opt
...
 
   std::vector<Model> models;
-  std::vector<ModelInfo> fetched_infos;
-  const std::string& cache_dir = cache_dir_;
+  models.reserve(model_infos.size());
+  for (const auto& info : model_infos) {
+    auto local_model = local_models.find(info.model_id);
+    auto local_path = local_model != local_models.end() ? local_model->second : std::string{};
+    models.push_back(model_factory_(ModelInfo(info), std::move(local_path)));
+  }
```
skottmckay: Can we do this at the same point we add the local model to model_infos?

sdk_v2/cpp/src/catalog/azure_model_catalog.cc
baijumeswani: yes. it addresses local models added post-snapshot and does the dedup by model id.

skottmckay (approved):

PR #1002, opened 2026-08-15
Add plan for supporting multiple catalogs better.

skottmckay:


sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -1040,6 +1040,10 @@ struct flModelApi {
   /// Use FL_MODEL_PROP_* constants for well-known keys, or any arbitrary string key.
   int64_t FL_API_T(Info_GetIntProperty, _In_ const flModelInfo* info, _In_ const char* key, int64_t default_value);
 
+  /// Which catalog source served this model: 0 = public (Azure), 1 = private, 2 = local.
+  /// Read-only. Defaults to 0 (public) for models without an explicit source.
+  int FL_API_T(Info_GetCatalogSource, _In_ const flModelInfo* info);
```
skottmckay: Catalog source should probably be an enum. `private` can be added in the future.

PR #1009, opened 2026-08-18
Add reasoning controls and preserve reasoning across chat turns

jennyf19:
# Add reasoning controls and preserve reasoning across chat turns

## Summary

Adds typed `chat_template_kwargs` to Chat Completions and preserves prior assistant reasoning for qualified model templates, including Qwen 3.8.

This combines two prompt-identity requirements:

1. Template controls such as `enable_thinking`, `preserve_thinking`, and `reasoning_effort` must be applied consistently to every render and retained-state decision.
2. Templates that explicitly support historical `reasoning_content` must receive the reasoning already stored in Foundry Local's typed conversation transcript.

### Chat template kwargs

- Parses typed `chat_template_kwargs` values without flattening booleans, numbers, arrays, or nested objects.
- Supports session defaults, request overrides, JSON payload precedence, and explicit `{}` clearing.
- Applies or clears kwargs under the shared tokenizer render lock.
- Threads effective kwargs through Generator, Engine, media, and continued-turn prompt rendering.
- Treats effective kwargs as retained-generator identity and rebuilds when they change.
- Supports stable GenAI packages and explicit capability selection for local GenAI builds.

### Preserved reasoning

- Retains incoming Chat Completions `reasoning_content` as reasoning-typed assistant transcript content.
- Preserves reasoning across stored and stateless Responses replay.
- Adds a behavioral template capability probe instead of branching on a model name alone.
- Qualifies preservation only when the loaded template:
  - renders distinct historical reasoning by default;
  - keeps visible assistant content distinct; and
  - omits historical reasoning when `preserve_thinking=false`.
- Emits assistant `reasoning_content` into canonical message JSON only for qualified templates.
- Lets the model template own `<think>` markers and `preserve_thinking` behavior.
- Uses the same capability, transcript, and effective kwargs for Generator and Engine prompt paths.
- Keeps the existing conservative omission behavior for unqualified templates.

## Why

Qwen 3.8 templates read prior assistant reasoning from `message.reasoning_content`. Foundry Local already classifies and stores generated reasoning separately, but previously omitted it when rebuilding canonical message JSON. Stateless Chat Completions also parsed incoming `reasoning_content` without retaining the bytes.

As a result, reconstructed prompts could contain empty historical thinking blocks and diverge from the conversation the model actually generated. This change makes the canonical prompt represent the typed transcript for templates that explicitly support preserved reasoning, while exact token-prefix comparison remains the final authority for retained-state reuse.

## Compatibility

- No reasoning is inserted into visible `content`.
- Unqualified templates do not receive `reasoning_content`.
- Existing tool-call, raw-envelope, positional-result, rollback, cancellation, and exact-prefix behavior remains in place.
- Qwen 3.8 support uses the existing GenAI `qwen3_5_text` model type and is qualified by rendered template behavior.

## Validation

- Added focused tests for Chat Completions reasoning ingestion and ordering.
- Added policy-driven message-projection tests.
- Added positive and fail-closed preserved-reasoning capability-probe tests.
- Updated warm/cold Responses replay tests for typed reasoning retention.
- Updated stateless Responses reasoning replay tests.
- Standalone 18-case chat-template kwargs CMake capability matrix passes.
- `git diff --check` passes.
- Editor diagnostics report no errors in changed source and test files.

### DIFF
--- a/sdk_v2/cpp/src/contracts/chat_completions.cc
+++ b/sdk_v2/cpp/src/contracts/chat_completions.cc
@@ -108,6 +108,12 @@ void from_json(const nlohmann::json& j, ChatCompletionRequest& r) {
 
   // metadata — map<string, string>
   opt(j, "metadata", r.metadata);
+
+  // chat_template_kwargs — typed JSON object
+  if (j.contains("chat_template_kwargs") && !j["chat_template_kwargs"].is_null()) {
+    auto kwargs = j["chat_template_kwargs"].get<std::map<std::string, nlohmann::json>>();
+    r.chat_template_kwargs = nlohmann::json(std::move(kwargs));
+  }
 }
 
 // ========================================================================
--- a/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
@@ -238,6 +238,7 @@ ToolCallContext ChatSession::BuildToolCallContext(const Request& request) const 
 
   tool_ctx.tool_call_start = get_param(FOUNDRY_LOCAL_MODEL_PROP_TOOL_CALL_START_STR);
   tool_ctx.tool_call_end = get_param(FOUNDRY_LOCAL_MODEL_PROP_TOOL_CALL_END_STR);
+  tool_ctx.template_kwargs_json = get_param("chat_template_kwargs");
 
   // Fall back to model info properties if not specified in the request
   const auto& info = CatalogModel().Info();
@@ -525,7 +526,8 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
         cached_tool_ctx_.supports_reasoning != turn_tool_ctx.supports_reasoning ||
         cached_tool_ctx_.reasoning_start != turn_tool_ctx.reasoning_start ||
         cached_tool_ctx_.reasoning_end != turn_tool_ctx.reasoning_end ||
-        cached_tool_ctx_.tools_json != turn_tool_ctx.tools_json;
+        cached_tool_ctx_.tools_json != turn_tool_ctx.tools_json ||
+        cached_tool_ctx_.template_kwargs_json != turn_tool_ctx.template_kwargs_json;
 
     // Classic Generator guidance and search settings are fixed at creation. Dynamic Engine options are supplied on
     // each BeginTurn, while the static Engine always rebuilds below because it cannot safely retain shared state.
--- a/sdk_v2/cpp/src/inferencing/generative/preprocessor.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/preprocessor.cc
@@ -73,9 +73,16 @@ std::string Preprocessor::ApplyChatTemplateWithOptions(const char* messages_json
                                                        const char* template_kwargs_json,
                                                        bool add_generation_prompt) {
   std::lock_guard<std::mutex> lock(mutex_);
+#if FOUNDRY_LOCAL_OGA_HAS_CHAT_TEMPLATE_KWARGS
   KeyValuePairs options;
   options.Add("chat_template_kwargs", template_kwargs_json ? template_kwargs_json : "{}");
   tokenizer_->UpdateOptions(options.Keys().data(), options.Values().data(), options.size());
+#else
+  if (template_kwargs_json && *template_kwargs_json) {
+    FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_USAGE,
+             "chat_template_kwargs requires a newer ONNX Runtime GenAI package");
+  }
+#endif
   OgaString result = tokenizer_->ApplyChatTemplate(
       /*template_str=*/nullptr, messages_json, tools_json, add_generation_prompt);
   return std::string(static_cast<const char*>(result));
--- a/sdk_v2/cpp/src/inferencing/generative/preprocessor.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/preprocessor.cc
@@ -6,6 +6,8 @@
 
 #include <ort_genai.h>
 
+#include <nlohmann/json.hpp>
+
 namespace fl {
 
 std::unique_ptr<Preprocessor> Preprocessor::Create(OgaModel& model, bool create_multimodal_processor) {
@@ -79,8 +81,16 @@ std::string Preprocessor::ApplyChatTemplateWithOptions(const char* messages_json
   tokenizer_->UpdateOptions(options.Keys().data(), options.Values().data(), options.size());
 #else
   if (template_kwargs_json && *template_kwargs_json) {
-    FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_USAGE,
-             "chat_template_kwargs requires a newer ONNX Runtime GenAI package");
+    auto template_kwargs = nlohmann::json::parse(
+        template_kwargs_json, nullptr, /*allow_exceptions=*/false);
+    if (!template_kwargs.is_object()) {
+      FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT,
+               "chat_template_kwargs must be a valid JSON object");
+    }
+    if (!template_kwargs.empty()) {
+      FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_USAGE,
+               "chat_template_kwargs requires a newer ONNX Runtime GenAI package");
+    }
   }
 #endif
   OgaString result = tokenizer_->ApplyChatTemplate(
--- a/sdk_v2/cpp/src/inferencing/generative/preprocessor.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/preprocessor.cc
@@ -65,9 +65,19 @@ std::unique_ptr<OgaSequences> Preprocessor::Encode(const char* text) {
 std::string Preprocessor::ApplyChatTemplate(const char* messages_json,
                                             const char* tools_json,
                                             bool add_generation_prompt) {
+  return ApplyChatTemplateWithOptions(messages_json, tools_json, nullptr, add_generation_prompt);
+}
+
+std::string Preprocessor::ApplyChatTemplateWithOptions(const char* messages_json,
+                                                       const char* tools_json,
+                                                       const char* template_kwargs_json,
+                                                       bool add_generation_prompt) {
   std::lock_guard<std::mutex> lock(mutex_);
-  OgaString result = tokenizer_->ApplyChatTemplate(/*template_str=*/nullptr, messages_json, tools_json,
-                                                   add_generation_prompt);
+  KeyValuePairs options;
+  options.Add("chat_template_kwargs", template_kwargs_json ? template_kwargs_json : "{}");
+  tokenizer_->UpdateOptions(options.Keys().data(), options.Values().data(), options.size());
+  OgaString result = tokenizer_->ApplyChatTemplate(
+      /*template_str=*/nullptr, messages_json, tools_json, add_generation_prompt);
   return std::string(static_cast<const char*>(result));
 }
 
--- a/sdk_v2/deps_versions.json
+++ b/sdk_v2/deps_versions.json
@@ -1,6 +1,6 @@
 {
   "_comment": "Single source of truth for native dependency versions in sdk_v2. Read by sdk_v2/cpp/cmake/Find*.cmake and sdk_v2/python/_build_backend/__init__.py. The .pipelines/foundry-local-packaging.yml literals must match; the 'Validate pinned versions' step fails the build on drift.",
   "onnxruntime":               { "version": "1.28.0" },
-  "onnxruntime-genai":         { "version": "0.15.2" },
+  "onnxruntime-genai":         { "version": "0.16.0-dev1001407373" },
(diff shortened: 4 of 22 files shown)


sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -441,19 +443,21 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
...
-    // Rebuild when: guidance requirements changed OR the previous turn had user-specified guidance
-    // (the finite grammar may have completed, causing IsDone() to return true on the next turn,
-    // and switching schemas requires a fresh grammar).
-    if (prev_needs_guidance != curr_needs_guidance || prev_has_user_guidance) {
-      // Guidance requirements changed — invalidate. The branch below will rebuild from full history.
+    // Rebuild when guidance requirements or template kwargs change, or when
+    // the previous turn had finite user guidance that may already be complete.
+    if (prev_needs_guidance != curr_needs_guidance || prev_has_user_guidance || template_kwargs_changed) {
+      // Invalidate so the branch below rebuilds from full history.
       cached_generator_.reset();
```
baijumeswani, line 543: **Cache rewind state becomes inconsistent after this rebuild.** When `template_kwargs_changed` resets the generator, the full-history generator created below replaces the old token sequence, but `pre_turn_token_count` remains `0`. `CommitTurn` records that value, and cancellation rollback and partial `UndoTurns` later use it as a valid rewind boundary. Cancelling or undoing this rebuilt turn can therefore rewind the generator to an empty sequence while `history_` still retains earlier turns; the next continuous turn appends only its new messages and loses that retained context. Please either invalidate the cache for rollback when this turn was rebuilt or track the correct rewind boundary, and add a regression covering kwargs A -> kwargs B -> undo/cancel -> replay versus a fresh full-history session.

baijumeswani (commented):
I found one cache/rollback correctness issue that should be addressed before approval; details are inline.

sdk_v2/deps_versions.json
```diff
@@ -1,6 +1,6 @@
 {
   "_comment": "Single source of truth for native dependency versions in sdk_v2. Read by sdk_v2/cpp/cmake/Find*.cmake and sdk_v2/python/_build_backend/__init__.py. The .pipelines/foundry-local-packaging.yml literals must match; the 'Validate pinned versions' step fails the build on drift.",
   "onnxruntime":               { "version": "1.28.0" },
-  "onnxruntime-genai":         { "version": "0.15.2" },
+  "onnxruntime-genai":         { "version": "0.16.0-dev1001407373" },
```
baijumeswani: Can we also keep this unchanged? Otherwise, it may become dificult for users to build foundry-local if the version of genai they need is a nightly one only available in ort-nightly. We can guard the changes behind a macro as is done in https://github.com/microsoft/foundry-local/pull/1071

bmehta001: Addressed in 44e9c47b after merging origin/main. The PR now leaves sdk_v2/deps_versions.json and the packaging pins unchanged from main: stable GenAI 0.16.0 and ORT 1.30.0 for ordinary and release builds. No nightly feed is required. The FOUNDRY_LOCAL_OGA_HAS_CHAT_TEMPLATE_KWARGS guard remains, with explicit rejection of non-empty kwargs when unsupported rather than silently ignoring them.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
bmehta001: Addressed in 44e9c47b using main's transcript-based cache handling. Changed or removed kwargs invalidate the generator before append, leaving the optional pre-turn token boundary unset. Undo therefore rebuilds from committed history instead of rewinding to zero; cancellation discards the uncommitted generator through the scope guard. The regressions now inspect real rewind boundaries (rather than the obsolete prompt-token-count shortcut) and compare changed/removed kwargs after undo or cancellation with fresh full-history replay.

sdk_v2/cpp/src/inferencing/generative/openresponses/response_converter.cc
```diff
@@ -341,8 +365,7 @@ static void AddJsonItemsToRequest(Request& request, const nlohmann::json& items)
     }
 
     if (type == "reasoning") {
-      // The text is private, but the assistant turn that produced it happened. Keep the boundary only.
-      request.AddOwnedItem(MakeAssistantTurnBoundary());
+      request.AddOwnedItem(MakeReasoningMessage(StoredReasoningText(entry)));
```
baijumeswani, line 369: This preservation behavior looks correct, but three existing tests now fail because they still expect the reasoning entry to be empty: `TypedReasoningInputItemBecomesAnAssistantBoundary`, `TypedConsecutiveReasoningItemsCollapseToOneBoundary`, and `HopOutputWithSeveralReasoningItemsStillEmitsOneBoundary`. Please update those tests to check the preserved `ReasoningText()` value, while still checking that reasoning does not appear in visible content. The focused test run should be green before approval.

sdk_v2/cpp/src/contracts/chat_completions.cc
```diff
@@ -367,6 +367,12 @@ void from_json(const nlohmann::json& j, ChatCompletionRequest& r) {
 
   // metadata — map<string, string>
   opt(j, "metadata", r.metadata);
+
+  // chat_template_kwargs — typed JSON object
+  if (j.contains("chat_template_kwargs") && !j["chat_template_kwargs"].is_null()) {
+    auto kwargs = j["chat_template_kwargs"].get<std::map<std::string, nlohmann::json>>();
```
baijumeswani, line 374: This adds a low-level template option, but standard OpenAI clients normally send Chat Completions `reasoning_effort` or Responses `reasoning.effort`. The first is currently ignored, and the second is parsed but does not change inference. That makes a request look accepted even though it has no effect. Please map the supported effort values to the model's template options, or reject unsupported values with a clear error. Add tests for both endpoints so clients get the same behavior.

sdk_v2/cpp/cmake/chat-template-kwargs.cmake
```diff
@@ -0,0 +1,39 @@
...
+    # UpdateOptions exists in older headers too; the bundled tokenizer determines support, not a C++ symbol.
+    # A cached package version cannot describe a caller-provided local build.
+    if(ORT_GENAI_HOME)
+        message(STATUS
+            "Local ORT GenAI chat template kwargs capability is unknown. "
+            "Set FOUNDRY_LOCAL_CHAT_TEMPLATE_KWARGS=ON if the local tokenizer supports it.")
+    elseif(ORT_GENAI_VERSION MATCHES "^0\\.16\\.0-dev([0-9]+)$")
+        if(CMAKE_MATCH_1 GREATER_EQUAL 1001407373)
+            set(FOUNDRY_LOCAL_OGA_HAS_CHAT_TEMPLATE_KWARGS ON)
+        endif()
```
baijumeswani, line 21: This version check says the stable 0.16 runtime supports the feature, but the packaged 0.16 runtime used in qualification could not render the Qwen3.8 template expression `enable_thinking is true`. A newer GenAI build was needed for the end-to-end continuation run. Please pin a runtime that can render the shipped template and add an installed-package Qwen smoke test. If the older runtime must remain supported, detect this behavior and return a clear compatibility error instead of reporting the capability as available.

sdk_v2/cpp/src/inferencing/generative/genai_model_instance.cc
```diff
@@ -67,24 +75,44 @@ ModelCapabilities ResolveModelCapabilities(std::string_view model_type, Preproce
...
+
+#if FOUNDRY_LOCAL_OGA_HAS_CHAT_TEMPLATE_KWARGS
+  try {
+    const auto reasoning_projection = preprocessor.ApplyChatTemplateWithOptions(
+      kReasoningProbeMessages, /*tools_json=*/nullptr, /*template_kwargs_json=*/nullptr,
+      /*add_generation_prompt=*/false);
+    const auto no_preserve_reasoning_projection = preprocessor.ApplyChatTemplateWithOptions(
+      kReasoningProbeMessages, /*tools_json=*/nullptr, kDisableReasoningPreservation,
+      /*add_generation_prompt=*/false);
+    capabilities.preserves_reasoning_history = model_capabilities_internal::ResolveRenderedProbes(
```
baijumeswani: Reasoning-history support is useful beyond this one Qwen model type, but this result is reached only when `model_type` is exactly `qwen3_5_text`. A compatible future model or template will silently lose reasoning history. Please keep the Qwen XML tool behavior model-specific, but run the reasoning-history check based on the loaded template/package capability rather than the model-type string.

sdk_v2/cpp/src/inferencing/generative/genai_model_instance.cc
```diff
@@ -98,7 +126,16 @@ ModelCapabilities model_capabilities_internal::ResolveRenderedProbes(
       tool_result_projection.find("probe-call-a") == std::string_view::npos &&
       tool_result_projection.find("probe-call-b") == std::string_view::npos;
 
-  return {native_qwen_xml_tool_calls, positional_tool_results};
+  const bool preserves_reasoning_history =
```
baijumeswani: This single check mixes three different questions: whether typed reasoning history is supported, whether reasoning should be preserved by default, and whether the template understands Qwen's `preserve_thinking` option. It can reject a compatible template with a different policy, or accept one that only prints the probe text as normal text. Please separate capability from policy: first verify that typed reasoning is handled as reasoning, then test the Qwen-specific preserve option separately.

sdk_v2/cpp/src/inferencing/generative/preprocessor.cc
```diff
@@ -65,9 +67,34 @@ std::unique_ptr<OgaSequences> Preprocessor::Encode(const char* text) {
...
+std::string Preprocessor::ApplyChatTemplateWithOptions(const char* messages_json,
+                                                       const char* tools_json,
+                                                       const char* template_kwargs_json,
+                                                       bool add_generation_prompt) {
   std::lock_guard<std::mutex> lock(mutex_);
-  OgaString result = tokenizer_->ApplyChatTemplate(/*template_str=*/nullptr, messages_json, tools_json,
-                                                   add_generation_prompt);
+#if FOUNDRY_LOCAL_OGA_HAS_CHAT_TEMPLATE_KWARGS
+  KeyValuePairs options;
+  options.Add("chat_template_kwargs", template_kwargs_json ? template_kwargs_json : "{}");
```
baijumeswani: In this supported-build path, native C/C++ callers can pass any raw string directly to GenAI. The fallback path validates that the value is a JSON object, and the HTTP path also requires an object, so behavior depends on how Foundry was built and which API the caller uses. Please parse and require a JSON object before this `#if`, then pass the normalized JSON to GenAI. This gives every API the same stable invalid-argument error for malformed input.

sdk_v2/cpp/test/internal_api/genai_model_instance_test.cc
```diff
@@ -140,6 +172,7 @@ TEST(ModelCapabilitiesTest, QualifiedQwenPackageResolvesBothProductionCapabiliti
   EXPECT_EQ(result.model->ModelType(), "qwen3_5_text");
   EXPECT_TRUE(result.model->HasNativeQwenXmlToolCalls());
   EXPECT_TRUE(result.model->HasPositionalToolResults());
+  EXPECT_TRUE(result.model->PreservesReasoningHistory());
```
baijumeswani: This is the only test that proves the real package and template produce the capability used by production, but it is skipped unless a developer sets an environment variable. The smaller tests provide prepared strings, so they cannot catch a package/runtime mismatch. Please make a qualified package fixture part of a required release or integration job and test a real multi-turn reasoning continuation there.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -1170,6 +1173,7 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
       cached_tool_ctx_.supports_reasoning != turn_tool_ctx.supports_reasoning ||
       cached_tool_ctx_.reasoning_start != turn_tool_ctx.reasoning_start ||
       cached_tool_ctx_.reasoning_end != turn_tool_ctx.reasoning_end ||
+      cached_tool_ctx_.template_kwargs_json != turn_tool_ctx.template_kwargs_json ||
```
baijumeswani, line 1187: This compares the raw JSON text. Two equivalent objects with different whitespace or key order will be treated as different and force a full prompt rebuild. Please validate and serialize the object into one consistent form when the option is accepted, then compare that normalized value here. This is not a correctness blocker, but it avoids unnecessary long-context prefill work.

baijumeswani (commented):
I reviewed the current head and also ran the focused tests and the real Qwen3.8-27B continuation test. The main continuation behavior works: warm and cold runs match, reasoning stays separate from visible text, and the thinking controls work when passed through template options. However, I do not think this is ready to approve yet. Three committed tests currently fail, the standard OpenAI reasoning controls do not affect generation, and the released runtime path still needs a package-level test with this model. The inline comments describe the concrete fixes. The model-type/probe and validation issues should also be addressed so this does not work only for one package by accident. Canonicalizing equivalent JSON options is lower priority but would avoid unnecessary prompt rebuilds.

baijumeswani (approved):

PR #1011, opened 2026-08-18
Synchronize tokenizer and multimodal preprocessing across shared sessions

baijumeswani:
### Summary

Fixes an intermittent Windows `invalid string_view position` failure when concurrent vision sessions share the same loaded model.

### Root cause

ORT Extensions tokenizer encoding is not reentrant. Foundry Local synchronized its model tokenizer previously, but ORT GenAI's shared multimodal processor owns a separate tokenizer that remained unguarded. Concurrent preprocessing could therefore corrupt mutable tokenizer state.

See discussion https://github.com/microsoft/foundry-local/pull/1008#issuecomment-5323549654 for details.


sdk_v2/cpp/src/inferencing/generative/preprocessor.cc
```diff
@@ -0,0 +1,118 @@
...
+#include <ort_genai.h>
+
+namespace fl {
+
+std::unique_ptr<Preprocessor> Preprocessor::Create(OgaModel& model, bool create_multimodal_processor) {
+  std::unique_ptr<OgaTokenizer> tokenizer;
+  try {
+    tokenizer = OgaTokenizer::Create(model);
+  } catch (const std::runtime_error& e) {
+    throw std::runtime_error(std::string("failed to create tokenizer: ") + e.what());
```
skottmckay: nit: FL_LOG_AND_THROW or FL_THROW

skottmckay (dismissed):

skottmckay (approved):

PR #1046, opened 2026-08-26
Add token-aware reasoning to Chat Completions and Responses

baijumeswani:
## Summary

Adds token-aware reasoning classification and exposes reasoning separately from visible output across Chat Completions and Responses.

Reasoning boundaries are detected from ORT GenAI BOR/EOR token metadata before detokenization, which handles special markers that decode to empty strings. Decoded marker matching remains as a compatibility fallback.

## Changes

- Classifies generated output into ordered reasoning and visible-text segments using token IDs.
- Supports empty-decoded special markers, multi-token markers, decoded-text fallback, truncated reasoning, and incomplete marker metadata.
- Excludes reasoning boundary tokens and EOS/control tokens from reasoning-token usage.
- Prevents tool-call-shaped reasoning from being interpreted as a real tool call.
- Preserves generation order across reasoning, visible text, and structured tool calls.
- Returns non-streaming Chat Completions reasoning through `message.reasoning_content`.
- Streams Chat Completions reasoning through `delta.reasoning_content` while keeping visible output in `delta.content`.
- Emits typed reasoning output items and events through Responses.
- Reports reasoning tokens through Chat Completions and Responses usage details.
- Emits the complete Responses function-call streaming lifecycle, including argument delta and done events.
- Preserves malformed marker-wrapped tool output as visible text instead of dropping it.
- Preserves an assistant history boundary after reasoning-only turns.

## API behavior

Reasoning-capable Chat Completions responses may contain both `message.reasoning_content` and visible `message.content`.

Streaming chunks contain either `delta.reasoning_content` or `delta.content`, never both. Models that do not produce reasoning omit `reasoning_content`.

The field is compatible with GitHub Copilot CLI and DeepSeek-style Chat Completions clients. GitHub Copilot CLI consumes it as separate reasoning events without mixing it into visible assistant output.

## Validation

- Canonical Linux Release build succeeds.
- Focused reasoning, Chat Completions, Responses, tool-call, history, and serialization tests pass: 175/175.
- Live Qwen3 HTTP tests pass for streaming and non-streaming reasoning output.
- GitHub Copilot CLI BYOK parses `reasoning_content` into separate reasoning timeline events while visible output remains in `content`.
- Full unit validation passes apart from existing model-cache fixtures that require `FOUNDRY_TEST_DATA_DIR`.
- Independent correctness, API, quality, and test reviews completed with no remaining material findings.

### DIFF
--- a/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
@@ -719,8 +719,7 @@ void ChatSession::ProcessChatCompletionsJson(const std::string& request_json, co
   int next_tool_call_index = 0;
   std::vector<GeneratedOutputEvent> generated_events;
 
-  // Chat Completions suppresses REASONING segments, but final response construction still consumes this same typed
-  // sequence so it does not need to re-split decoded text.
+  // Use the same typed segments for streaming and final response construction.
   auto splitter = CreateReasoningSplitter(tool_ctx, Model());
 
   auto emit_visible_text = [&](std::string visible) {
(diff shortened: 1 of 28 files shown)


jiafatom (commented):
The internal token-aware split and Responses API reasoning events look good. For equivalent separation in OpenAI-compatible Chat Completions, could we add an optional `reasoning_content` member to `ChatCompletionDelta`, serialize it when present, and emit reasoning segments through that field instead of dropping them?

The expected reasoning chunk would be:

```json
{
  "choices": [{
    "delta": {
      "reasoning_content": "Analyzing project files..."
    }
  }]
}
```

followed by visible output through:

```json
{
  "choices": [{
    "delta": {
      "content": "I found three issues."
    }
  }]
}
```

Tests should cover reasoning-only and interleaved reasoning/content chunks so ordinary models continue emitting only `delta.content`.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -796,43 +743,39 @@ void ChatSession::ProcessChatCompletionsJson(const std::string& request_json, co
       // REASONING segments: intentionally dropped from the Chat Completions stream. Never feed reasoning text to
```
jiafatom: Could we stream these segments through `delta.reasoning_content` instead of dropping them? The splitter already provides the necessary classification, and emitting a separate field would let Copilot/VS Code and other OpenAI-compatible clients hide, collapse, or style reasoning independently while preserving visible output in `delta.content`.

jiafatom (dismissed):
Please fix the copilot comment and then we are good to go.

jiafatom (approved):

PR #1060, opened 2026-09-01
Fix malformed tool-call recovery for local models

apsonawane:
## Summary

Improves tool-call parsing and streaming recovery for local models that emit malformed or legacy tool-call payloads.

The PR was rebased onto the latest `main` and refactored to remove functionality already covered by #838, #1043, #1046, and #1083.

## Changes

- Recover narrowly malformed tool-call JSON shapes.
- Normalize legacy tool names against advertised tools.
- Preserve malformed tool-call blocks as visible text when recovery fails.
- Recover complete calls with missing or incorrect closing markers.
- Avoid treating tool-call markers inside JSON strings as nested calls.
- Preserve all-or-nothing validation for arrays containing invalid calls.
- Pass advertised tool metadata through streaming parser paths.

## Validation

- C++ build completed successfully.
- 51 focused parser and streaming accumulator tests passed.
- Ran 10 LiveCodeBench samples locally using the ONNX model:
  - All 10 reached evaluation.
  - 3 of 10 resolved successfully.
  - No infrastructure, authentication, or missing-result errors.


sdk_v2/cpp/src/items/message_item.h
```diff
@@ -62,6 +63,7 @@ struct MessagePart {
 struct MessageItem : Item {
   flMessageRole role;
   std::vector<MessagePart> content;
+  std::vector<ToolCallItem> tool_calls;
```
skottmckay, line 66: Does this need to be part of the MessageItem vs. multiple Items being provided in the Request (e.g. MessageItem and potentially multiple ToolResultItem instances) to keep them decoupled?

skottmckay: Or multiple ToolCallItem instances in the Response items.

sdk_v2/cpp/src/inferencing/generative/toolcalling/tool_call_utils.cc
```diff
@@ -121,7 +344,20 @@ std::vector<ParsedToolCall> ParseToolCalls(const std::string& text,
     }
```
kunal-vaishnavi, line 362: From Copilot:

Could we locate the closing marker only when it appears outside a JSON string? As written, `find()` can match marker text inside a valid argument—for example, `"cmd":"grep '</tool_call>' file"`—which truncates the payload, makes parsing fail, and skips the valid call. Please scan with JSON string/escape state and add a regression test covering an end marker embedded in an argument string.

kunal-vaishnavi (dismissed):

sdk_v2/cpp/src/inferencing/generative/toolcalling/tool_call_stream_accumulator.h
```diff
@@ -114,7 +117,7 @@ class ToolCallStreamAccumulator {
           tool_call_buffer_ += buffer_.substr(0, found + marker.size());
           buffer_.erase(0, found + marker.size());
 
-          auto parsed = ParseToolCalls(tool_call_buffer_, start_marker_, end_marker_);
+          auto parsed = ParseToolCalls(tool_call_buffer_, start_marker_, end_marker_, tools_json_);
```
baijumeswani: The closing marker just above this call is found even when it is inside a JSON string. For example, a shell command that searches for `</tool_call>` would be cut in the middle and the valid call would be lost. Please find closing markers only when they are outside JSON strings, and use the same helper here and in `ParseToolCalls`.

sdk_v2/cpp/src/inferencing/generative/toolcalling/tool_call_utils.cc
```diff
@@ -29,65 +28,286 @@ std::string RandomAlphanumeric(int length) {
...
-    tc.arguments = args_it->is_string() ? args_it->get<std::string>() : args_it->dump();
-  } else if (auto params_it = call.find("parameters"); params_it != call.end()) {
-    tc.arguments = params_it->is_string() ? params_it->get<std::string>() : params_it->dump();
+  name = trim(std::move(name));
+  if (name.starts_with("function=")) {
+    name = trim(name.substr(sizeof("function=") - 1));
   }
 
-  return tc;
+  if (name == "exec_command" &&
```
baijumeswani: This makes Foundry Local rename `exec_command` to `shell`, even when the advertised tool list is empty. That can create a call to a tool the client never offered. Please remove the empty-list fallback and move these name aliases into model metadata or caller configuration instead of hard-coding them here.

sdk_v2/cpp/src/inferencing/generative/toolcalling/tool_call_utils.cc
```diff
@@ -29,65 +28,286 @@ std::string RandomAlphanumeric(int length) {
...
+        } else {
+          tc.arguments = arguments.is_string() ? arguments.get<std::string>() : arguments.dump();
+        }
+        return tc.name.empty() ? std::nullopt
+                               : std::optional<ParsedToolCall>(std::move(tc));
+      } else {
+        return std::nullopt;
+      }
+
+      tc.name = NormalizeToolName(std::move(tc.name), advertised_tools);
```
baijumeswani: After repairing a malformed call, we should check that the final name is one of the tools the client actually offered. Otherwise malformed text can become a real call to an unknown tool. Please reject the recovery when the normalized name is not in the advertised tool list.

sdk_v2/cpp/src/inferencing/generative/toolcalling/tool_call_utils.cc
```diff
@@ -121,7 +344,20 @@ std::vector<ParsedToolCall> ParseToolCalls(const std::string& text,
...
 
     std::string content = text.substr(content_start, end_pos - content_start);
-    auto calls = DeserializeToolCalls(content);
+    auto calls = DeserializeToolCalls(content, advertised_tools);
+
+    // If a malformed call was left open before another call, parse the
+    // innermost complete block rather than combining both payloads. Only try
+    // this recovery after the outer payload fails because marker text may be
+    // valid JSON string content.
+    if (calls.empty()) {
```
baijumeswani: When this finds a valid inner call, the malformed text before it disappears. The caller sees only the recovered call and cannot see what was dropped. Please return the leftover text as text, or make the streaming parser keep it, so recovery never silently removes model output.

sdk_v2/cpp/src/inferencing/generative/toolcalling/tool_call_utils.cc
```diff
@@ -29,65 +28,286 @@ std::string RandomAlphanumeric(int length) {
...
+      return results;
+    }
+    const size_t content_end = json_text.find_last_not_of(" \t\r\n");
+    const std::string normalized_text =
+        json_text.substr(content_start, content_end - content_start + 1);
+
+    auto json = nlohmann::json::parse(normalized_text, nullptr, false);
+    bool repaired_missing_name_prefix = false;
+
+    // Some models finish a complete object one outer brace early. Repair
```
baijumeswani, line 87: This file now has several different guesses for repairing broken JSON. A successful JSON parse does not prove that a guess kept the model's intended meaning. Please limit each repair to a captured output shape from a supported model, require an advertised tool match, and add a test showing the recovered call can be executed and continued safely.

baijumeswani (commented):
The smaller scope is much better, but I do not think the recovery code is ready yet. A few cases can still turn ordinary text into a tool call or lose text. I added the details inline. Please also rebase onto current main now that #1083 has merged, rerun the focused tests, and include a before/after model test that continues after the recovered call.

baijumeswani (changes requested):

baijumeswani (dismissed):

sayanshaw24 (approved):

PR #1062, opened 2026-09-02
Bring Your Own Model Language Bindings for C#, JavaScript, Python, and Rust SDKs

selenayang888:
Extends the Bring Your Own Model (BYOM) API introduced in [BYOM](https://github.com/microsoft/foundry-local/pull/928) to the C#, JavaScript/TypeScript, Python, and Rust SDK v2 bindings.

Applications can select the local catalog, register an existing local ONNX model directory, use the returned model through the standard inference APIs, and unregister it without deleting or taking ownership of the model assets.

### DIFF
--- a/sdk_v2/cs/src/Catalog.cs
+++ b/sdk_v2/cs/src/Catalog.cs
@@ -118,11 +118,12 @@ internal sealed class Catalog : ICatalog
     {
         Detail.Throw.IfNull(metadata);
 
+        using var nativeMetadata = new NativeModelInfo();
+        PopulateNativeMetadata(nativeMetadata, metadata);
+
         return await Utils.CallWithExceptionHandlingAsync(
             () =>
             {
-                using var nativeMetadata = new NativeModelInfo();
-                PopulateNativeMetadata(nativeMetadata, metadata);
                 NativeModel model;
                 lock (_nativeLifetimeLock)
                 {
--- a/sdk_v2/cs/src/Catalog.cs
+++ b/sdk_v2/cs/src/Catalog.cs
@@ -12,18 +12,25 @@ using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 
 using NativeCatalog = Microsoft.AI.Foundry.Local.Detail.Native.Catalog;
+using NativeModel = Microsoft.AI.Foundry.Local.Detail.Native.Model;
+using NativeModelInfo = Microsoft.AI.Foundry.Local.Detail.Native.MutableModelInfo;
 
 internal sealed class Catalog : ICatalog
 {
     private readonly NativeCatalog _nativeCatalog;
     private readonly ILogger _logger;
+    private readonly object _nativeLifetimeLock;
+    private readonly Func<bool> _isManagerDisposed;
 
     public string Name { get; }
 
-    internal Catalog(NativeCatalog nativeCatalog, ILogger logger)
+    internal Catalog(NativeCatalog nativeCatalog, ILogger logger, object nativeLifetimeLock,
+                     Func<bool> isManagerDisposed)
     {
         _nativeCatalog = nativeCatalog;
         _logger = logger;
+        _nativeLifetimeLock = nativeLifetimeLock;
+        _isManagerDisposed = isManagerDisposed;
         Name = _nativeCatalog.GetName();
     }
 
@@ -104,5 +111,112 @@ internal sealed class Catalog : ICatalog
             },
             $"Error getting latest version for model with name '{model.Info.Name}'.",
             _logger, ct).ConfigureAwait(false);
+    }
+
+    public async Task<IModel> RegisterModelAsync(string modelPath, string modelId, ModelInfo metadata,
+                                                 CancellationToken? ct = null)
+    {
+        Detail.Throw.IfNull(metadata);
+
+        return await Utils.CallWithExceptionHandlingAsync(
+            () =>
+            {
+                using var nativeMetadata = new NativeModelInfo();
+                PopulateNativeMetadata(nativeMetadata, metadata);
+                NativeModel model;
+                lock (_nativeLifetimeLock)
+                {
+                    ThrowIfManagerDisposed();
+                    model = _nativeCatalog.RegisterModel(modelPath, modelId, nativeMetadata);
+                }
+                return (IModel)new Model(model, _logger);
+            },
+            $"Error registering model '{modelId}'.", _logger, ct).ConfigureAwait(false);
+    }
+
+    public async Task UnregisterModelAsync(string aliasOrModelId, CancellationToken? ct = null)
+    {
+        await Utils.CallWithExceptionHandlingAsync(
+            () =>
+            {
+                lock (_nativeLifetimeLock)
+                {
+                    ThrowIfManagerDisposed();
+                    _nativeCatalog.UnregisterModel(aliasOrModelId);
+                }
+            },
+            $"Error unregistering model '{aliasOrModelId}'.", _logger, ct).ConfigureAwait(false);
+    }
+
+    private void ThrowIfManagerDisposed()
+    {
+        Detail.Throw.IfDisposed(_isManagerDisposed(), this);
+    }
+
+    private static void PopulateNativeMetadata(NativeModelInfo nativeMetadata, ModelInfo metadata)
+    {
+        SetStringIfNotEmpty(nativeMetadata, ModelInfoPropertyKeys.DisplayName, metadata.DisplayName);
+        SetStringIfNotEmpty(nativeMetadata, ModelInfoPropertyKeys.ModelType, metadata.ModelType);
+        SetStringIfNotEmpty(nativeMetadata, ModelInfoPropertyKeys.Publisher, metadata.Publisher);
+        SetStringIfNotEmpty(nativeMetadata, ModelInfoPropertyKeys.License, metadata.License);
+        SetStringIfNotEmpty(nativeMetadata, ModelInfoPropertyKeys.LicenseDescription, metadata.LicenseDescription);
+        SetStringIfNotEmpty(nativeMetadata, ModelInfoPropertyKeys.Task, metadata.Task);
+        SetStringIfNotEmpty(nativeMetadata, ModelInfoPropertyKeys.ModelProvider, metadata.ProviderType);
+        SetStringIfNotEmpty(nativeMetadata, ModelInfoPropertyKeys.MinimumFoundryLocalVersion, metadata.MinFLVersion);
+        SetStringIfNotEmpty(nativeMetadata, ModelInfoPropertyKeys.InputModalities, metadata.InputModalities);
+        SetStringIfNotEmpty(nativeMetadata, ModelInfoPropertyKeys.OutputModalities, metadata.OutputModalities);
+        SetStringIfNotEmpty(nativeMetadata, ModelInfoPropertyKeys.Capabilities, metadata.Capabilities);
+
+        if (metadata.Runtime?.DeviceType is DeviceType.CPU or DeviceType.GPU or DeviceType.NPU)
+        {
+            nativeMetadata.SetStringProperty(ModelInfoPropertyKeys.DeviceType, metadata.Runtime.DeviceType.ToString());
+        }
+
+        SetStringIfNotEmpty(nativeMetadata, ModelInfoPropertyKeys.ExecutionProvider,
+                            metadata.Runtime?.ExecutionProvider);
+
+        if (metadata.FileSizeMb.HasValue)
+        {
+            nativeMetadata.SetIntProperty(ModelInfoPropertyKeys.FileSizeMb, metadata.FileSizeMb.Value);
+        }
+
+        if (metadata.SupportsToolCalling.HasValue)
+        {
+            nativeMetadata.SetIntProperty(ModelInfoPropertyKeys.SupportsToolCalling,
+                                          metadata.SupportsToolCalling.Value ? 1 : 0);
+        }
+
+        if (metadata.MaxOutputTokens.HasValue)
+        {
+            nativeMetadata.SetIntProperty(ModelInfoPropertyKeys.MaxOutputTokens, metadata.MaxOutputTokens.Value);
+        }
+
+        if (metadata.CreatedAtUnix != 0)
+        {
+            nativeMetadata.SetIntProperty(ModelInfoPropertyKeys.CreatedAtUnix, metadata.CreatedAtUnix);
+        }
+
+        if (metadata.ContextLength.HasValue)
+        {
+            nativeMetadata.SetIntProperty(ModelInfoPropertyKeys.ContextLength, metadata.ContextLength.Value);
+        }
+
+        foreach (var property in metadata.StringProperties)
+        {
+            nativeMetadata.SetStringProperty(property.Key, property.Value);
+        }
+
+        foreach (var property in metadata.IntProperties)
+        {
+            nativeMetadata.SetIntProperty(property.Key, property.Value);
+        }
+    }
+
+    private static void SetStringIfNotEmpty(NativeModelInfo metadata, string key, string? value)
+    {
+        if (!string.IsNullOrEmpty(value))
+        {
+            metadata.SetStringProperty(key, value!);
+        }
     }
 }
--- a/sdk_v2/cs/src/FoundryModelInfo.cs
+++ b/sdk_v2/cs/src/FoundryModelInfo.cs
@@ -7,6 +7,7 @@
 namespace Microsoft.AI.Foundry.Local;
 
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Text.Json.Serialization;
 
 using NativeModelType = Microsoft.AI.Foundry.Local.Detail.Native.Model;
@@ -60,6 +61,21 @@ public record ModelSettings
 
 public record ModelInfo
 {
+    /// <summary>
+    /// Creates mutable metadata for local model registration. Model identity is supplied separately to
+    /// <see cref="ICatalog.RegisterModelAsync"/> and is populated by the catalog on the returned model.
+    /// </summary>
+    [SetsRequiredMembers]
+    public ModelInfo()
+    {
+        Id = string.Empty;
+        Name = string.Empty;
+        Alias = string.Empty;
+        ProviderType = string.Empty;
+        Uri = string.Empty;
+        ModelType = string.Empty;
+    }
+
     [JsonPropertyName("id")]
     public required string Id { get; init; }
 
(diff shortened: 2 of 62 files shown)


baijumeswani (commented):
Thanks for adding BYOM support across the language SDKs. The basic path looks good, and the PR checks are green. I also built the native SDK and ran the focused tests locally: JS 34/34 passed, Python 17 passed with 4 environment skips, and the Rust BYOM round trip passed. I found a few cases that are not covered by those tests. The most important one is a native crash when a JS catalog is used after the manager is disposed. Python also loses custom metadata after the same model is fetched from the catalog again. I do not think this is ready for approval until the lifetime and metadata round-trip issues are fixed. I left simple inline comments with reproduction details and suggested tests.

sdk_v2/js/native/src/catalog.cc
```diff
@@ -242,4 +247,93 @@ Napi::Value Catalog::GetLatestVersion(const Napi::CallbackInfo& info) {
...
+  std::string model_path;
+  std::string model_id;
+  std::shared_ptr<foundry_local::ModelInfo> metadata;
+  if (!ReadRegistrationArgs(info, model_path, model_id, metadata)) return env.Undefined();
+
+  auto* catalog = impl_;
+  auto manager_pin = std::make_shared<Napi::ObjectReference>(CloneManager(manager_));
+  return PromiseWorker<std::unique_ptr<foundry_local::IModel>>::Run(
+      env,
+      [catalog, model_path, model_id, metadata]() {
```
baijumeswani, line 292: This worker keeps the JavaScript catalog object alive, but it does not keep the native manager alive after `manager.dispose()`. The worker can then call a freed catalog pointer. I also reproduced a simpler case: read `catalog.name` after `manager.dispose()`, and Node crashed with exit code 139. Please share the native lifetime with these catalog objects, or make every catalog call fail safely after manager disposal. Please add tests for using a catalog after disposal and disposing while register/unregister is still running.

sdk_v2/python/src/foundry_local_sdk/catalog.py
```diff
@@ -55,12 +82,77 @@ def list_models(self) -> list[IModel]:
...
+                metadata._native_ptr,
+                out,
+            )
+        )
+        if out[0] == ffi.NULL:
+            raise FoundryLocalException("RegisterModel returned no model.")
+        return _ModelImpl(
+            out[0],
+            parent=self,
+            string_properties=metadata._string_properties,
```
baijumeswani, line 137: These copied dictionaries only stay on the model returned by this call. When I fetched the same model again with `get_model_variant`, the custom string changed from `persist-me` to `None` and the custom integer changed from `42` to `-1`. Please read these properties from the registered native metadata so the result is the same after another lookup or manager restart. A round-trip test would catch this.

sdk_v2/js/src/modelInfo.ts
```diff
@@ -0,0 +1,105 @@
...
+  DisplayName: "display_name",
+  ModelType: "type",
+  Publisher: "publisher",
+  License: "license",
+  LicenseDescription: "license_description",
+  Task: "task",
+  ModelProvider: "model_provider",
+  MinFoundryLocalVersion: "min_fl_version",
+  ParentUri: "parent_uri",
+  ToolCallStart: "tool_call_start",
```
baijumeswani, line 15: We can set tool and reasoning markers here, but the public model snapshot does not expose these values again. The same applies to the reasoning flags below. Please add these fields to the read-only model info path and test that registration followed by a catalog lookup returns the same values.

sdk_v2/js/native/src/catalog.cc
```diff
@@ -242,4 +247,93 @@ Napi::Value Catalog::GetLatestVersion(const Napi::CallbackInfo& info) {
...
+bool ReadRegistrationArgs(const Napi::CallbackInfo& info, std::string& model_path, std::string& model_id,
+                          std::shared_ptr<foundry_local::ModelInfo>& metadata) {
+  Napi::Env env = info.Env();
+  if (info.Length() != 3 || !info[0].IsString() || !info[1].IsString()) {
+    Napi::TypeError::New(env, "registerModel(modelPath: string, modelId: string, metadata: ModelInfo)")
+        .ThrowAsJavaScriptException();
+    return false;
+  }
+  NativeModelInfo* wrapper = ExtractModelInfo(env, info[2]);
+  if (wrapper == nullptr || (metadata = wrapper->Snapshot()) == nullptr) {
```
baijumeswani, line 269: `Snapshot()` allocates and calls the C++ wrapper, so it can throw. This call is outside `CallChecked`, which means a native exception may escape without the normal JavaScript error conversion. Please create the snapshot inside the common checked boundary before starting the worker.

sdk_v2/js/src/modelInfo.ts
selenayang888: Thanks. I added the marker and reasoning fields to the native snapshot and public `ModelInfo` type, then added a test that registers these values, fetches the model again through `getModelVariant()`, and verifies they round-trip correctly.

sdk_v2/python/src/foundry_local_sdk/catalog.py
selenayang888: Thanks for catching this. Fixed by removing the Python-side dictionaries attached only to the initially returned model. Arbitrary string and integer properties are now read from the registered native metadata.

Added round-trip tests confirming `persist-me` and `42` are preserved after both `get_model_variant()` and manager recreation.

sdk_v2/js/native/src/catalog.cc
selenayang888: `Snapshot()` now executes inside `CallCheckedVoid` before worker creation.

sdk_v2/js/native/src/catalog.cc
selenayang888: Fixed by sharing native manager lifetime with in-flight catalog workers and rejecting catalog calls safely after manager disposal.

Added tests covering:

Accessing `catalog.name` after disposal throws `FoundryLocalError` instead of crashing.
Disposing after register/unregister workers have started.
Registration and unregistration complete safely without dereferencing freed catalog pointers.

sdk_v2/js/native/src/manager.cc
```diff
@@ -250,19 +250,37 @@ Napi::Value Manager::GetCatalog(const Napi::CallbackInfo& info) {
...
     return Catalog::NewInstance(env, std::move(token));
   });
 }
 
 Napi::Value Manager::Dispose(const Napi::CallbackInfo& info) {
   Napi::Env env = info.Env();
-  // Idempotent — releasing an already-null unique_ptr is a no-op.
+  // Idempotent. In-flight catalog workers hold a temporary shared lease; retained catalogs hold only weak ownership
+  // and reject new calls after this reset.
+  disposed_->store(true);
```
baijumeswani, line 277: The new lease/disposed mechanism protects `Catalog`, but it is not propagated into `ModelCtorToken`. Models returned before this call still hold raw manager-owned `IModel*` values, and pinning the JavaScript manager object does not prevent this explicit `impl_.reset()`. A later `model.info`/`isCached`, or an in-flight `load`/`unload`/`download` worker, can therefore dereference freed native state. Please pass the shared native-manager lease and disposed flag into every model wrapper, reject model calls after disposal, and have asynchronous model workers retain that lease until completion. Add retained-model and dispose-during-model-operation tests alongside the catalog lifetime tests.

sdk_v2/cs/src/Catalog.cs
```diff
@@ -105,4 +112,112 @@ public async Task<IModel> GetLatestVersionAsync(IModel model, CancellationToken?
...
+        return await Utils.CallWithExceptionHandlingAsync(
+            () =>
+            {
+                NativeModel model;
+                lock (_nativeLifetimeLock)
+                {
+                    ThrowIfManagerDisposed();
+                    model = _nativeCatalog.RegisterModel(modelPath, modelId, nativeMetadata);
+                }
+                return (IModel)new Model(model, _logger);
```
baijumeswani: This returned `Model` does not receive `_nativeLifetimeLock` or `_isManagerDisposed`. Registration itself is protected, but `FoundryLocalManager.Dispose()` can subsequently race `Model.Info`, `LoadAsync`, `UnloadAsync`, or `IsCachedAsync`, all of which use manager-owned native handles without this lock. Please propagate a manager lifetime guard into every model returned by catalog operations and hold it across each native model call; add sequential post-dispose and concurrent dispose/operation tests.

sdk_v2/python/src/foundry_local_sdk/catalog.py
```diff
@@ -55,12 +82,72 @@ def list_models(self) -> list[IModel]:
...
+            raise FoundryLocalException("Models can only be registered in the local catalog.")
+        if not isinstance(metadata, ModelInfoBuilder):
+            raise TypeError("metadata must be a ModelInfoBuilder instance")
+
+        from foundry_local_sdk._native.api import api, ffi
+
+        model_path_bytes = os.fspath(model_path).encode("utf-8")
+        model_id_bytes = model_id.encode("utf-8")
+        out = ffi.new("flModel**")
+        api.check_status(
```
baijumeswani: `_ensure_manager_open()` is a check outside the native call. Another thread can run `manager.close()` after line 112 and release the manager before this dispatch uses the borrowed catalog pointer. The same check-then-call pattern exists in catalog queries and most `_ModelImpl` lifecycle/state methods. Please hold the manager's lock across the open check and complete native call everywhere a manager-owned catalog/model pointer is used (an `RLock` may simplify nested helpers), and add a deterministic close-race test.

sdk_v2/cs/src/FoundryModelInfo.cs
```diff
@@ -60,6 +61,21 @@ public record ModelSettings
 
 public record ModelInfo
 {
+    /// <summary>
+    /// Creates mutable metadata for local model registration. Model identity is supplied separately to
+    /// <see cref="ICatalog.RegisterModelAsync"/> and is populated by the catalog on the returned model.
+    /// </summary>
+    [SetsRequiredMembers]
```
baijumeswani: Using the existing immutable `ModelInfo` record as the mutable registration builder changes its public value contract: `[SetsRequiredMembers]` now permits `new ModelInfo()` with empty required identity fields, the new dictionary fields participate in synthesized equality by reference, and `with` clones share those mutable dictionaries. Please keep `ModelInfo` as the read-only value and introduce a separate `ModelInfoBuilder`/`MutableModelInfo`, as the other bindings do. This preserves required-member enforcement, equality/hash behavior, and clone isolation.

sdk_v2/js/native/src/model.cc
```diff
@@ -128,7 +133,15 @@ Napi::Object SnapshotModelInfo(Napi::Env env, const foundry_local::ModelInfo& in
   SetOptionalString(env, out, "modelProvider", info.ModelProvider());
   SetOptionalString(env, out, "minFLVersion", info.MinFlVersion());
   SetOptionalString(env, out, "parentUri", info.ParentUri());
+  SetOptionalString(env, out, "toolCallStart", info.GetStringProperty(FOUNDRY_LOCAL_MODEL_PROP_TOOL_CALL_START_STR));
```
baijumeswani, line 156: Adding the known marker fields fixes their round trip, but arbitrary properties accepted by `MutableModelInfo.setStringProperty`/`setIntProperty` are still omitted from the returned JavaScript snapshot. C# and Rust have the same write-only arbitrary-metadata gap, while Python now supports keyed native readback. Please define one cross-binding contract and, if arbitrary values are intended to be observable, expose keyed string/int getters (or complete detached maps) and test fresh lookup plus manager-recreation persistence.

sdk_v2/js/src/catalog.ts
```diff
@@ -94,22 +97,68 @@ export class Catalog {
...
+  }
+  return native;
+}
+
 function wrapAll(natives: readonly NativeModel[]): IModel[] {
   return natives.map((n) => wrapNativeModel(n));
 }
+
+function validateNonEmptyString(value: string, name: string): void {
+  if (typeof value !== "string" || value.trim() === "") {
```
baijumeswani, line 152: This validation accepts embedded NULs. Because the native API consumes NUL-terminated strings, values such as `model:1\0suffix` are silently truncated and can register, look up, or unregister a different ID than the caller supplied. Please reject `\0` before native dispatch for paths, IDs/aliases, metadata keys, and string metadata values. Rust already gets this behavior from `CString::new`; C# and Python need equivalent boundary validation as well.

baijumeswani (commented):
Fresh review against the current head, focused on the BYOM lifecycle and metadata issues found during qualification. The earlier JavaScript catalog-disposal and Python metadata-persistence concerns appear addressed. The inline comments below cover the remaining lifetime, API-semantics, metadata-parity, and FFI-boundary concerns.

sdk_v2/js/native/src/manager.cc
selenayang888: Fixed. Models now carry the manager lease/disposed state, and async workers retain both manager and model ownership. Added retained-model and concurrent-disposal tests.

sdk_v2/cs/src/Catalog.cs
selenayang888: Every model returned by catalog operations now receives the manager lifetime lock and disposed-state guard. The lock is held across native model calls, variant operations, session creation, and deprecated client paths. Native manager shutdown/release uses the same lock. Added sequential and concurrent disposal tests.

sdk_v2/python/src/foundry_local_sdk/catalog.py
selenayang888: Catalog/model native calls now hold the manager `RLock` across the open check and dispatch. Added a deterministic close-race test.

sdk_v2/cs/src/FoundryModelInfo.cs
selenayang888: Fixed.  `ModelInfo` remains the immutable read-only record with required identity members. The parameterless [SetsRequiredMembers] constructor, mutable dictionaries, and mutation methods were removed.

Registration now uses a separate public `ModelInfoBuilder`, and `RegisterModelAsync` accepts that builder. This preserves required-member enforcement, record equality/hash behavior, and with clone isolation.

sdk_v2/js/native/src/model.cc
selenayang888: Addressed with a consistent keyed getter contract. Models in JavaScript, C#, Python, and Rust now expose string and integer property getters backed by the native metadata accessors. Added fresh-lookup and manager-recreation coverage.

sdk_v2/js/src/catalog.ts
selenayang888: Fixed. JS, C#, and Python now reject embedded NULs in paths, identifiers, metadata keys/values, and getter keys before native dispatch.

baijumeswani (commented):
Thanks for addressing the earlier feedback. I found two remaining JavaScript lifetime gaps. Both can leave native work using a manager after dispose().

sdk_v2/js/native/src/session.cc
```diff
@@ -315,7 +315,7 @@ ChatSession::ChatSession(const Napi::CallbackInfo& info) : Napi::ObjectWrap<Chat
   }
   Napi::Object model_obj = info[0].As<Napi::Object>();
   Model* model = Napi::ObjectWrap<Model>::Unwrap(model_obj);
-  foundry_local::IModel* native = model != nullptr ? model->native_impl() : nullptr;
+  foundry_local::IModel* native = model != nullptr ? model->native_impl(env) : nullptr;
```
baijumeswani, line 427: This check protects session creation, but the session does not keep the native manager alive. It only keeps a JavaScript reference to the Manager object. Calling manager.dispose() resets the native shared pointer, so a retained session or an in-flight request can use native state owned by a destroyed manager. Please store a shared native-manager lease and the disposed flag in each session wrapper. Reject new session calls after disposal, and let each in-flight worker keep its own lease until it finishes. Please add tests for using a retained session after disposal and disposing during a request.

sdk_v2/js/native/src/manager.cc
```diff
@@ -222,7 +222,7 @@ Manager::Manager(const Napi::CallbackInfo& info) : Napi::ObjectWrap<Manager>(inf
       kvp.Set("DisableNonessentialTelemetry", "true");
       config.SetAdditionalOptions(kvp);
     }
-    impl_ = std::make_unique<foundry_local::Manager>(std::move(config));
+    impl_ = std::make_shared<foundry_local::Manager>(std::move(config));
```
baijumeswani, line 219: Please also use this shared ownership for EpDownloadWorker. The worker still receives impl_.get() and stores a raw Manager pointer. A progress callback can call manager.dispose(), which resets impl_ while the worker is still inside DownloadAndRegisterEps. Pass a shared_ptr to the worker instead (for example, copy impl_ before queuing and store that copy in the worker), then add a test that disposes the manager from a download progress callback.

sdk_v2/js/native/src/session.h
```diff
@@ -52,6 +52,8 @@ class ChatSession : public Napi::ObjectWrap<ChatSession> {
   bool ThrowIfDisposed(Napi::Env env);
 
   std::unique_ptr<foundry_local::ChatSession> impl_;
+  std::shared_ptr<foundry_local::Manager> manager_lifetime_;
```
baijumeswani, line 58: Please make sure the native session is destroyed before this manager lease. C++ destroys members in reverse declaration order, so the current order releases `manager_lifetime_` before `impl_` during garbage collection. If this is the last manager lease, the session destructor can then use an invalid model. Declare the manager lease before `impl_`, or add an explicit destructor that resets `impl_` first, for all three session types.

sdk_v2/js/native/src/session.cc
```diff
@@ -98,17 +98,20 @@ foundry_local::Request* UnwrapRequest(Napi::Env env, const Napi::Value& v) {
...
 
   using Result = std::shared_ptr<foundry_local::Response>;
   struct Pins {
+    std::shared_ptr<foundry_local::Manager> manager_lifetime;
     Napi::ObjectReference manager;
     Napi::ObjectReference request;
   };
-  auto pins = std::make_shared<Pins>(Pins{std::move(manager_ref), std::move(req_pin)});
+  auto pins = std::make_shared<Pins>(
+      Pins{std::move(manager_lifetime), std::move(manager_ref), std::move(req_pin)});
```
baijumeswani, line 196: These pins keep the manager and request alive, but the worker still uses a raw session pointer. `session.dispose()` or garbage collection can destroy `impl_` while the worker is running. Please give each admitted worker shared ownership of the native session, or block disposal until the operation finishes. Please cover both normal and streaming requests with dispose/GC tests.

baijumeswani (commented):
Thanks for addressing the earlier manager lifetime comments. The manager lease and EP download ownership changes are good. I found two remaining session lifetime issues that should be fixed before merge.

baijumeswani (approved):

PR #1064, opened 2026-09-02
Fix streaming audio language selection

austinh-msft:
The AudioSession API never binds the language to the underlying model preventing recognition of things other than English. This enables properly binding for multi-lingual model

### DIFF
--- a/sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
@@ -332,6 +332,12 @@ void AudioSession::ProcessStreamingAudio(const AudioItem& format_item, ItemQueue
   }
 
   auto generator = OgaGenerator::Create(oga_model, *gen_params);
+  if (IsNemotronSpeechModel()) {
+    auto language = effective_kvp.find("language");
+    if (language != effective_kvp.end()) {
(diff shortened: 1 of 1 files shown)


sdk_v2/cpp/src/inferencing/generative/audio/audio_session.cc
```diff
@@ -332,6 +332,12 @@ void AudioSession::ProcessStreamingAudio(const AudioItem& format_item, ItemQueue
   }
 
   auto generator = OgaGenerator::Create(oga_model, *gen_params);
+  if (IsNemotronSpeechModel()) {
```
baijumeswani, line 335: This fixes at the cpp level. Are there changes required at other language binding layers?

austinh-msft: All generic AudioSession projections ultimately call the shared C++ AudioSession::ProcessStreamingAudio path: 
* C# and Python/Rust call the C ABI Session_ProcessRequest
* JavaScript’s native addon wraps foundry_local::AudioSession directly
* Each projection supports passing "language" through RequestOptions.additionalOptions / AdditionalOptions

It does look like there are similar issues in the LiveAudioTranscriptionSession but that seems deprecated.

baijumeswani (approved):

PR #1071, opened 2026-09-03
Enable model-selected Engine chat batching

bmehta001:
## Summary

- select the existing Generator or dynamic Engine batching from each model's `genai_config.json`
- add one model-owned Engine dispatcher that keeps every Engine/Request operation on its owner thread while serving concurrent independent `ChatSession`s
- retain dynamic Engine requests across turns for KV-cache reuse, with safe rebuild from committed history after eviction or cancellation
- bound capacity waits without failing the shared Engine or active peer conversations
- preserve Generator behavior for models without Engine batching and reject unsupported/ambiguous Engine configurations
- capability-gate Engine sources and tests so the shipping ORT GenAI 0.15.2 dependency remains buildable while the implementation is ready for an Engine-capable replacement

### Missing assistant-turn boundary

ORT GenAI Engine does not append or retain a generated EOS token in the request's logical sequence. A later `BeginTurn` must therefore include the model-specific tokens that close the prior assistant turn before introducing the next user turn. The Generator path already retains EOS, so reusing its append fragment caused dynamic Engine sessions to lose conversational context even though the request and KV cache remained resident.

This change derives the continuation boundary from the model's own chat template using a collision-free synthetic assistant marker, removes everything through that marker before tokenization, and submits only the model-specific assistant boundary plus the new-turn fragment. This avoids hard-coding token IDs or template syntax and avoids injecting fresh-conversation BOS/default-system framing into retained state.

### Per-turn output limits

When callers explicitly provide `max_output_tokens`, every Engine turn forwards that value through `OgaTurnOptions`, validates it against the retained-plus-new input context, and waits for the corresponding terminal event. When callers omit `max_output_tokens`, Foundry does not invent or forward a default Engine turn limit; OGA enforces the Request/session context ceiling.

Engine usage and finish reasons come from terminal events. Generator-backed requests retain their established host-side default and sampling behavior.

### Capacity and cancellation

Completed dynamic requests remain resident for KV-cache reuse and are evicted least-recently-used when capacity is needed. If every resident request is active, the dispatcher continues servicing cancellation/close commands and waits up to a bounded deadline. Expiration cancels and closes only the newest waiting no-progress conversation rather than poisoning the model-owned Engine or active peers.

A late cancellation after completion is a no-op. Canceled turns are not committed to history: retained state is rewound when the selected inference path can restore its pre-turn position, otherwise it is discarded and rebuilt from committed history on the next request.

## Backend contract

| Model configuration | Backend | Multi-turn behavior |
|---|---|---|
| `engine.dynamic_batching` | Shared dynamic Engine | Retain request and paged KV state; rebuild after eviction or non-restorable cancellation |
| `engine.static_batching` | Unsupported configuration | Model load fails explicitly |
| No `engine` object | Existing Generator | Retain existing Generator continuation |
| Empty, misspelled, or ambiguous `engine` object | Configuration error | No implicit fallback |

Engine selection is limited to text chat models. Multimodal Engine declarations fail at model load rather than failing later during a request.

## Runtime compatibility

The repository and release pipeline remain pinned to `Microsoft.ML.OnnxRuntimeGenAI.Foundry` 0.15.2. CMake probes for the dynamic Engine API and excludes Engine-only sources/tests when it is unavailable, so current release builds remain Generator-only. The future implementation was also compiled against cached `0.16.0-dev1001407373`, where the capability probe succeeds.

An Engine-enabled CI lane must stage a real paged-KV model and set `FOUNDRY_LOCAL_DYNAMIC_ENGINE_TEST_MODEL_PATH`; the test policy documents this requirement so an enabled lane cannot silently rely on the conventional shared-buffer fixture.

## Validation

- Windows Debug build with shipping ORT GenAI 0.15.2
- Linux/WSL RelWithDebInfo build with shipping ORT GenAI 0.15.2
- Windows RelWithDebInfo build with ORT GenAI `0.16.0-dev1001407373`, compiling the Engine implementation and dynamic test suite
- 12 targeted Windows tests for Generator cancellation recovery, session recovery, and Engine option planning
- 9 future-runtime option tests passed; 6 paged-model Engine tests compiled and skipped locally because `FOUNDRY_LOCAL_DYNAMIC_ENGINE_TEST_MODEL_PATH` was not set
- changed C++ files pass the repository clang-format configuration

### DIFF
--- a/.pipelines/foundry-local-packaging.yml
+++ b/.pipelines/foundry-local-packaging.yml
@@ -53,12 +53,17 @@ parameters:
 
 variables:
 - group: FoundryLocal-ESRP-Signing
-# C++ SDK (sdk_v2/cpp) native dependency versions. Must match cmake defaults
-# in sdk_v2/deps_versions.json.
+# C++ SDK (sdk_v2/cpp) native dependency versions. Release builds use the
+# stable pins from sdk_v2/deps_versions.json; non-release CI validates the
+# selected GenAI nightly before it becomes a stable SDK default.
 - name: cppOrtVersion
   value: '1.28.0'
-- name: cppGenaiVersion
-  value: '0.15.3'
+- ${{ if eq(parameters.isRelease, true) }}:
+  - name: cppGenaiVersion
+    value: '0.15.3'
+- ${{ else }}:
+  - name: cppGenaiVersion
+    value: '0.16.0-dev1001400138'
 - name: cppWinmlVersion
   value: '2.1.70'
 - name: cppBuildConfig
--- a/.pipelines/foundry-local-packaging.yml
+++ b/.pipelines/foundry-local-packaging.yml
@@ -60,7 +60,7 @@ variables:
   value: '1.28.0'
 - ${{ if eq(parameters.isRelease, true) }}:
   - name: cppGenaiVersion
-    value: '0.15.3'
+    value: '0.15.2'
 - ${{ else }}:
   - name: cppGenaiVersion
     value: '0.16.0-dev1001400138'
--- a/.pipelines/foundry-local-packaging.yml
+++ b/.pipelines/foundry-local-packaging.yml
@@ -58,7 +58,7 @@ variables:
 - name: cppOrtVersion
   value: '1.28.0'
 - name: cppGenaiVersion
-  value: '0.15.2'
+  value: '0.15.3'
 - name: cppWinmlVersion
   value: '2.1.70'
 - name: cppBuildConfig
--- a/.pipelines/foundry-local-packaging.yml
+++ b/.pipelines/foundry-local-packaging.yml
@@ -53,17 +53,13 @@ parameters:
 
 variables:
 - group: FoundryLocal-ESRP-Signing
-# C++ SDK (sdk_v2/cpp) native dependency versions. Release builds use the
-# stable pins from sdk_v2/deps_versions.json; non-release CI validates the
-# selected GenAI nightly before it becomes a stable SDK default.
+# C++ SDK (sdk_v2/cpp) native dependency versions. Keep these synchronized
+# with sdk_v2/deps_versions.json so every language package is built and tested
+# against the same runtime ABI.
 - name: cppOrtVersion
   value: '1.28.0'
-- ${{ if eq(parameters.isRelease, true) }}:
-  - name: cppGenaiVersion
-    value: '0.15.2'
-- ${{ else }}:
-  - name: cppGenaiVersion
-    value: '0.16.0-dev1001407373'
+- name: cppGenaiVersion
+  value: '0.15.2'
 - name: cppWinmlVersion
   value: '2.1.70'
 - name: cppBuildConfig
--- a/.pipelines/foundry-local-packaging.yml
+++ b/.pipelines/foundry-local-packaging.yml
@@ -272,4 +272,3 @@ extends:
         ortVersion: $(cppOrtVersion)
         genaiVersion: $(cppGenaiVersion)
         winmlVersion: $(cppWinmlVersion)
-        bundleNativeDependencies: ${{ not(parameters.isRelease) }}
--- a/.pipelines/foundry-local-packaging.yml
+++ b/.pipelines/foundry-local-packaging.yml
@@ -63,7 +63,7 @@ variables:
     value: '0.15.2'
 - ${{ else }}:
   - name: cppGenaiVersion
-    value: '0.16.0-dev1001400138'
+    value: '0.16.0-dev1001407373'
 - name: cppWinmlVersion
   value: '2.1.70'
 - name: cppBuildConfig
--- a/.pipelines/foundry-local-packaging.yml
+++ b/.pipelines/foundry-local-packaging.yml
@@ -53,13 +53,17 @@ parameters:
 
 variables:
 - group: FoundryLocal-ESRP-Signing
-# C++ SDK (sdk_v2/cpp) native dependency versions. Keep these synchronized
-# with sdk_v2/deps_versions.json so every language package is built and tested
-# against the same runtime ABI.
+# C++ SDK (sdk_v2/cpp) native dependency versions. Release builds use the
+# stable pins from sdk_v2/deps_versions.json; non-release CI validates the
+# selected GenAI nightly before it becomes a stable SDK default.
 - name: cppOrtVersion
   value: '1.28.0'
-- name: cppGenaiVersion
-  value: '0.15.2'
+- ${{ if eq(parameters.isRelease, true) }}:
+  - name: cppGenaiVersion
+    value: '0.15.2'
+- ${{ else }}:
+  - name: cppGenaiVersion
+    value: '0.16.0-dev1001400138'
 - name: cppWinmlVersion
   value: '2.1.70'
 - name: cppBuildConfig
--- a/.pipelines/foundry-local-packaging.yml
+++ b/.pipelines/foundry-local-packaging.yml
@@ -272,3 +272,4 @@ extends:
         ortVersion: $(cppOrtVersion)
         genaiVersion: $(cppGenaiVersion)
         winmlVersion: $(cppWinmlVersion)
+        bundleNativeDependencies: ${{ not(parameters.isRelease) }}
--- a/sdk_v2/cpp/include/foundry_local/foundry_local_c.h
+++ b/sdk_v2/cpp/include/foundry_local/foundry_local_c.h
@@ -269,13 +269,13 @@ typedef enum flTensorDataType {
 #define FOUNDRY_LOCAL_MODEL_PROP_CREATION_TIME_STR "creation_time"              ///< ISO-8601 UTC timestamp
 
 /* flModelInfo Int properties. Comments provide details on the type and expected values. */
-#define FOUNDRY_LOCAL_MODEL_PROP_SUPPORTS_TOOL_CALLING_INT "supports_tool_calling"          ///< optional bool (not set or -1=unknown, 0=false, 1=true)
-#define FOUNDRY_LOCAL_MODEL_PROP_SUPPORTS_REASONING_INT "supports_reasoning"                ///< optional bool (not set or -1=unknown, 0=false, 1=true)
-#define FOUNDRY_LOCAL_MODEL_PROP_FILESIZE_MB_INT "filesize_mb"                              ///< optional int32_t
-#define FOUNDRY_LOCAL_MODEL_PROP_MAX_OUTPUT_TOKENS_INT "max_output_tokens"                  ///< optional int32_t
-#define FOUNDRY_LOCAL_MODEL_PROP_CREATED_AT_UNIX_INT "created_at_unix"                      ///< Unix timestamp. default=0
-#define FOUNDRY_LOCAL_MODEL_PROP_IS_TEST_MODEL_INT "is_test_model"                          ///< bool (0=false, 1=true)
-#define FOUNDRY_LOCAL_MODEL_PROP_CONTEXT_LENGTH_INT "context_length"                        ///< optional int64_t
+#define FOUNDRY_LOCAL_MODEL_PROP_SUPPORTS_TOOL_CALLING_INT "supports_tool_calling"  ///< optional bool (not set or -1=unknown, 0=false, 1=true)
+#define FOUNDRY_LOCAL_MODEL_PROP_SUPPORTS_REASONING_INT "supports_reasoning"        ///< optional bool (not set or -1=unknown, 0=false, 1=true)
+#define FOUNDRY_LOCAL_MODEL_PROP_FILESIZE_MB_INT "filesize_mb"                      ///< optional int32_t
+#define FOUNDRY_LOCAL_MODEL_PROP_MAX_OUTPUT_TOKENS_INT "max_output_tokens"          ///< optional int32_t
+#define FOUNDRY_LOCAL_MODEL_PROP_CREATED_AT_UNIX_INT "created_at_unix"              ///< Unix timestamp. default=0
+#define FOUNDRY_LOCAL_MODEL_PROP_IS_TEST_MODEL_INT "is_test_model"                  ///< bool (0=false, 1=true)
+#define FOUNDRY_LOCAL_MODEL_PROP_CONTEXT_LENGTH_INT "context_length"                ///< optional int64_t
 #define FOUNDRY_LOCAL_MODEL_PROP_SUPPORTS_HYBRID_REASONING_INT "supports_hybrid_reasoning"  ///< optional bool
 
 #define FOUNDRY_LOCAL_MODEL_PROP_INPUT_MODALITIES_STR "input_modalities"    ///< optional, comma-separated
@@ -298,10 +298,9 @@ typedef enum flTensorDataType {
 #define FOUNDRY_LOCAL_PARAM_FREQUENCY_PENALTY "frequency_penalty"
 /// Float presence penalty. Currently only the neutral value 0 is supported.
 #define FOUNDRY_LOCAL_PARAM_PRESENCE_PENALTY "presence_penalty"
-#define FOUNDRY_LOCAL_PARAM_SEED "seed"  ///< int. for reproducible outputs
-/// Bool. Beam-search early-stopping policy. Only false is currently supported.
-#define FOUNDRY_LOCAL_PARAM_EARLY_STOPPING "early_stopping"
-#define FOUNDRY_LOCAL_PARAM_DO_SAMPLE "do_sample"  ///< bool. whether to sample (false = greedy decoding)
+#define FOUNDRY_LOCAL_PARAM_SEED "seed"                            ///< int. for reproducible outputs
+#define FOUNDRY_LOCAL_PARAM_EARLY_STOPPING "early_stopping"        ///< bool. whether to stop on stop sequence or only at max tokens
+#define FOUNDRY_LOCAL_PARAM_DO_SAMPLE "do_sample"                  ///< bool. whether to sample (false = greedy decoding)
 
 /* Request options */
 #define FOUNDRY_LOCAL_PARAM_TOOL_CHOICE "tool_choice"  ///< string: See flToolChoice for the typed enum.
--- a/sdk_v2/cpp/include/foundry_local/foundry_local_c.h
+++ b/sdk_v2/cpp/include/foundry_local/foundry_local_c.h
@@ -299,7 +299,7 @@ typedef enum flTensorDataType {
 /// Float presence penalty. Currently only the neutral value 0 is supported.
 #define FOUNDRY_LOCAL_PARAM_PRESENCE_PENALTY "presence_penalty"
 #define FOUNDRY_LOCAL_PARAM_SEED "seed"  ///< int. for reproducible outputs
-/// Bool. Legacy beam-search policy; unsupported by Engine backends.
+/// Bool. Beam-search early-stopping policy; unsupported by Engine backends.
 #define FOUNDRY_LOCAL_PARAM_EARLY_STOPPING "early_stopping"
 #define FOUNDRY_LOCAL_PARAM_DO_SAMPLE "do_sample"  ///< bool. whether to sample (false = greedy decoding)
 
--- a/sdk_v2/cpp/include/foundry_local/foundry_local_c.h
+++ b/sdk_v2/cpp/include/foundry_local/foundry_local_c.h
@@ -269,13 +269,13 @@ typedef enum flTensorDataType {
 #define FOUNDRY_LOCAL_MODEL_PROP_CREATION_TIME_STR "creation_time"              ///< ISO-8601 UTC timestamp
 
 /* flModelInfo Int properties. Comments provide details on the type and expected values. */
-#define FOUNDRY_LOCAL_MODEL_PROP_SUPPORTS_TOOL_CALLING_INT "supports_tool_calling"  ///< optional bool (not set or -1=unknown, 0=false, 1=true)
-#define FOUNDRY_LOCAL_MODEL_PROP_SUPPORTS_REASONING_INT "supports_reasoning"        ///< optional bool (not set or -1=unknown, 0=false, 1=true)
-#define FOUNDRY_LOCAL_MODEL_PROP_FILESIZE_MB_INT "filesize_mb"                      ///< optional int32_t
-#define FOUNDRY_LOCAL_MODEL_PROP_MAX_OUTPUT_TOKENS_INT "max_output_tokens"          ///< optional int32_t
-#define FOUNDRY_LOCAL_MODEL_PROP_CREATED_AT_UNIX_INT "created_at_unix"              ///< Unix timestamp. default=0
-#define FOUNDRY_LOCAL_MODEL_PROP_IS_TEST_MODEL_INT "is_test_model"                  ///< bool (0=false, 1=true)
-#define FOUNDRY_LOCAL_MODEL_PROP_CONTEXT_LENGTH_INT "context_length"                ///< optional int64_t
+#define FOUNDRY_LOCAL_MODEL_PROP_SUPPORTS_TOOL_CALLING_INT "supports_tool_calling"          ///< optional bool (not set or -1=unknown, 0=false, 1=true)
+#define FOUNDRY_LOCAL_MODEL_PROP_SUPPORTS_REASONING_INT "supports_reasoning"                ///< optional bool (not set or -1=unknown, 0=false, 1=true)
+#define FOUNDRY_LOCAL_MODEL_PROP_FILESIZE_MB_INT "filesize_mb"                              ///< optional int32_t
+#define FOUNDRY_LOCAL_MODEL_PROP_MAX_OUTPUT_TOKENS_INT "max_output_tokens"                  ///< optional int32_t
+#define FOUNDRY_LOCAL_MODEL_PROP_CREATED_AT_UNIX_INT "created_at_unix"                      ///< Unix timestamp. default=0
+#define FOUNDRY_LOCAL_MODEL_PROP_IS_TEST_MODEL_INT "is_test_model"                          ///< bool (0=false, 1=true)
+#define FOUNDRY_LOCAL_MODEL_PROP_CONTEXT_LENGTH_INT "context_length"                        ///< optional int64_t
 #define FOUNDRY_LOCAL_MODEL_PROP_SUPPORTS_HYBRID_REASONING_INT "supports_hybrid_reasoning"  ///< optional bool
 
 #define FOUNDRY_LOCAL_MODEL_PROP_INPUT_MODALITIES_STR "input_modalities"    ///< optional, comma-separated
@@ -299,7 +299,7 @@ typedef enum flTensorDataType {
 /// Float presence penalty. Currently only the neutral value 0 is supported.
 #define FOUNDRY_LOCAL_PARAM_PRESENCE_PENALTY "presence_penalty"
 #define FOUNDRY_LOCAL_PARAM_SEED "seed"  ///< int. for reproducible outputs
-/// Bool. Beam-search early-stopping policy; unsupported by Engine backends.
+/// Bool. Beam-search early-stopping policy. Only false is currently supported.
 #define FOUNDRY_LOCAL_PARAM_EARLY_STOPPING "early_stopping"
 #define FOUNDRY_LOCAL_PARAM_DO_SAMPLE "do_sample"  ///< bool. whether to sample (false = greedy decoding)
 
--- a/sdk_v2/cpp/include/foundry_local/foundry_local_c.h
+++ b/sdk_v2/cpp/include/foundry_local/foundry_local_c.h
@@ -294,8 +294,10 @@ typedef enum flTensorDataType {
 #define FOUNDRY_LOCAL_PARAM_TOP_P "top_p"                          ///< float [0.0, 1.0]. nucleus sampling
 #define FOUNDRY_LOCAL_PARAM_TOP_K "top_k"                          ///< int. top-k sampling
 #define FOUNDRY_LOCAL_PARAM_MAX_OUTPUT_TOKENS "max_output_tokens"  ///< int. max tokens to generate
-#define FOUNDRY_LOCAL_PARAM_FREQUENCY_PENALTY "frequency_penalty"  ///< float [-2.0, 2.0]
-#define FOUNDRY_LOCAL_PARAM_PRESENCE_PENALTY "presence_penalty"    ///< float [-2.0, 2.0]
+/// Float frequency penalty. Currently only the neutral value 0 is supported.
+#define FOUNDRY_LOCAL_PARAM_FREQUENCY_PENALTY "frequency_penalty"
+/// Float presence penalty. Currently only the neutral value 0 is supported.
+#define FOUNDRY_LOCAL_PARAM_PRESENCE_PENALTY "presence_penalty"
 #define FOUNDRY_LOCAL_PARAM_SEED "seed"                            ///< int. for reproducible outputs
 #define FOUNDRY_LOCAL_PARAM_EARLY_STOPPING "early_stopping"        ///< bool. whether to stop on stop sequence or only at max tokens
 #define FOUNDRY_LOCAL_PARAM_DO_SAMPLE "do_sample"                  ///< bool. whether to sample (false = greedy decoding)
--- a/sdk_v2/cpp/include/foundry_local/foundry_local_c.h
+++ b/sdk_v2/cpp/include/foundry_local/foundry_local_c.h
@@ -298,9 +298,10 @@ typedef enum flTensorDataType {
 #define FOUNDRY_LOCAL_PARAM_FREQUENCY_PENALTY "frequency_penalty"
 /// Float presence penalty. Currently only the neutral value 0 is supported.
 #define FOUNDRY_LOCAL_PARAM_PRESENCE_PENALTY "presence_penalty"
-#define FOUNDRY_LOCAL_PARAM_SEED "seed"                            ///< int. for reproducible outputs
-#define FOUNDRY_LOCAL_PARAM_EARLY_STOPPING "early_stopping"        ///< bool. whether to stop on stop sequence or only at max tokens
-#define FOUNDRY_LOCAL_PARAM_DO_SAMPLE "do_sample"                  ///< bool. whether to sample (false = greedy decoding)
+#define FOUNDRY_LOCAL_PARAM_SEED "seed"  ///< int. for reproducible outputs
+/// Bool. Legacy beam-search policy; unsupported by Engine backends.
+#define FOUNDRY_LOCAL_PARAM_EARLY_STOPPING "early_stopping"
+#define FOUNDRY_LOCAL_PARAM_DO_SAMPLE "do_sample"  ///< bool. whether to sample (false = greedy decoding)
 
 /* Request options */
 #define FOUNDRY_LOCAL_PARAM_TOOL_CHOICE "tool_choice"  ///< string: See flToolChoice for the typed enum.
--- a/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
@@ -159,6 +159,17 @@ ToolCallContext ChatSession::BuildToolCallContext(const Request& request) const 
     }
   }
 
+  // Catalog metadata is immutable and may not contain markers for models whose
+  // tokenizer defines them dynamically. Read those markers from the loaded GenAI
+  // model without mutating the published ModelInfo.
+  const auto& tag_info = model_.GetTagInfo();
+  if (tool_ctx.tool_call_start.empty()) {
+    tool_ctx.tool_call_start = tag_info.bot_str;
+  }
+  if (tool_ctx.tool_call_end.empty()) {
+    tool_ctx.tool_call_end = tag_info.eot_str;
+  }
+
   // Check if the model supports chain-of-thought reasoning
   const auto* reasoning_val = info.GetPropertyInt(FOUNDRY_LOCAL_MODEL_PROP_SUPPORTS_REASONING_INT);
   if (reasoning_val && *reasoning_val == 1) {
@@ -181,6 +192,12 @@ ToolCallContext ChatSession::BuildToolCallContext(const Request& request) const 
     if (val) {
       tool_ctx.reasoning_end = *val;
     }
+  }
+  if (tool_ctx.reasoning_start.empty()) {
+    tool_ctx.reasoning_start = tag_info.bor_str;
+  }
+  if (tool_ctx.reasoning_end.empty()) {
+    tool_ctx.reasoning_end = tag_info.eor_str;
   }
 
   // Accumulate tool definitions from the session.
@@ -317,7 +334,8 @@ static std::vector<TextSegment> SplitReasoningContent(const std::string& text,
 void ChatSession::ProcessGeneratedOutput(std::string text, const ToolCallContext& tool_ctx,
                                          const SearchOptions& effective_options, bool canceled,
                                          Response& response, int prompt_tokens, int total_tokens,
-                                         std::vector<ParsedToolCall> pre_parsed_calls) {
+                                         std::vector<ParsedToolCall> pre_parsed_calls,
+                                         std::optional<flFinishReason> backend_finish_reason) {
   int completion_tokens = total_tokens - prompt_tokens;
 
   // Check if the generated text contains tool calls. If the caller has already parsed them (streaming path), reuse
@@ -379,6 +397,8 @@ void ChatSession::ProcessGeneratedOutput(std::string text, const ToolCallContext
     response.finish_reason = FOUNDRY_LOCAL_FINISH_NONE;
   } else if (has_tool_calls) {
     response.finish_reason = FOUNDRY_LOCAL_FINISH_TOOL_CALLS;
+  } else if (backend_finish_reason.has_value()) {
+    response.finish_reason = *backend_finish_reason;
   } else {
     int max_output = effective_options.max_output_tokens.value_or(0);
 
@@ -636,9 +656,11 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
   }
 
   int total_tokens = cached_generator_->TokenCount();
+  std::optional<flFinishReason> backend_finish_reason;
   if (const auto turn_usage = cached_generator_->GetTurnUsage()) {
     prompt_tokens = turn_usage->prompt_tokens;
     total_tokens = turn_usage->prompt_tokens + turn_usage->generated_tokens;
+    backend_finish_reason = turn_usage->finish_reason;
   }
   bool discard_generator = false;
 
@@ -651,7 +673,8 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
   }
 
   ProcessGeneratedOutput(std::move(text), cached_tool_ctx_, effective_options, request.canceled,
-                         response, prompt_tokens, total_tokens, std::move(streamed_tool_calls));
+                         response, prompt_tokens, total_tokens, std::move(streamed_tool_calls),
+                         backend_finish_reason);
 
   if (discard_generator) {
     cached_generator_.reset();
@@ -872,16 +895,19 @@ void ChatSession::ProcessChatCompletionsJson(const std::string& request_json, co
   }
 
   int total_tokens = generator->TokenCount();
+  std::optional<flFinishReason> backend_finish_reason;
   if (const auto turn_usage = generator->GetTurnUsage()) {
     prompt_tokens = turn_usage->prompt_tokens;
     total_tokens = turn_usage->prompt_tokens + turn_usage->generated_tokens;
+    backend_finish_reason = turn_usage->finish_reason;
   }
 
   // Process the generated output into response items (MessageItem, ToolCallItem, etc.)
   // This also updates finish_reason, and usage on the response. Streamed-parsed tool calls are reused so call_ids
   // stay stable across stream deltas and the final ChatCompletionResponse.
   ProcessGeneratedOutput(std::move(text), tool_ctx, options, original_request.canceled,
-                         response, prompt_tokens, total_tokens, std::move(streamed_tool_calls));
+                         response, prompt_tokens, total_tokens, std::move(streamed_tool_calls),
+                         backend_finish_reason);
 
   // Emit final streaming chunk with finish_reason
   if (is_streaming) {
--- a/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
@@ -509,10 +509,13 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
 
     bool prev_needs_guidance = cached_tool_ctx_.tool_output && !cached_tool_ctx_.text_output;
     bool curr_needs_guidance = turn_tool_ctx.tool_output && !turn_tool_ctx.text_output;
+    const bool guidance_changed =
+        cached_tool_ctx_.guidance_type != turn_tool_ctx.guidance_type ||
+        cached_tool_ctx_.guidance_data != turn_tool_ctx.guidance_data;
 
     const bool static_engine =
         Model().GetGenAIConfig().GetChatBackendKind() == ChatBackendKind::kStaticEngine;
-    if (prev_needs_guidance != curr_needs_guidance || static_engine) {
+    if (prev_needs_guidance != curr_needs_guidance || guidance_changed || static_engine) {
       // Guidance requirements changed — invalidate. The branch below will rebuild from full history.
       cached_generator_.reset();
       cached_tool_ctx_ = {};
--- a/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
@@ -5,9 +5,7 @@
 
 #include "contracts/chat_completions.h"
 #include "contracts/chat_completions_converter.h"
-#ifdef FOUNDRY_LOCAL_HAS_OGA_ENGINE
 #include "inferencing/generative/chat/onnx_engine_chat_generator.h"
-#endif
 #include "inferencing/generative/chat/onnx_chat_generator.h"
 #include "inferencing/generative/chat/reasoning_stream_splitter.h"
 #include "inferencing/generative/genai_model_instance.h"
@@ -56,12 +54,7 @@ std::unique_ptr<ChatGenerator> CreateTextChatGenerator(const std::vector<Message
                                                        GenAIModelInstance& model,
                                                        const ToolCallContext& tool_ctx) {
   if (model.GetGenAIConfig().GetChatBackendKind() != ChatBackendKind::kGenerator) {
-#ifdef FOUNDRY_LOCAL_HAS_OGA_ENGINE
     return OnnxEngineChatGenerator::Create(messages, options, model, tool_ctx);
-#else
-    FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL,
-             "model requires the ORT GenAI Engine API, but this build does not provide it");
-#endif
   }
 
   return OnnxChatGenerator::Create(messages, options, model, tool_ctx, /*use_full_context=*/true);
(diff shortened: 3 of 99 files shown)


sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
```diff
@@ -0,0 +1,351 @@
...
+    fail(error);
+    return;
+  }
+  command_cv_.notify_one();
+}
+
+void OnnxChatEngine::WorkerLoop(std::promise<void> initialized) {
+  try {
+    engine_ = OgaEngine::Create(model_.GetOgaModel());
+    event_buffer_ = engine_->CreateEventBuffer(model_.GetGenAIConfig().EngineMaxBatchSize().value_or(1) * 2);
```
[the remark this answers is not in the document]
bmehta001: The release pin intentionally remains 0.15.2 because 0.15.3 is not published: the NuGet gallery currently lists 0.15.2 as latest, and the configured AIFoundryLocal public feed returns HTTP 404 for 0.15.3. Non-release CI builds this PR against the selected Engine-capable `0.16.0-dev1001400138` nightly (`.pipelines/foundry-local-packaging.yml:64-66`), while the stable pin will move when an Engine-capable release is available. No dependency-pin change is possible in this PR without referencing a nonexistent package.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_engine_chat_generator.cc
```diff
@@ -0,0 +1,175 @@
...
+#include <algorithm>
+
+namespace fl {
+
+namespace {
+
+std::optional<flFinishReason> MapFinishReason(OgaFinishReason reason) {
+  switch (reason) {
+    case OgaFinishReason_Eos:
+    case OgaFinishReason_StopSequence:
```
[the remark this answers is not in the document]
bmehta001: The exact non-release CI dependency (`0.16.0-dev1001400138`) defines `OgaFinishReason_StopSequence`, and the branch now compiles successfully against that package. Commit dd13bfe4 adds a preprocessor compatibility guard for both `StopSequence` and the later upstream `StopString` spelling, so either Engine-capable API variant maps to `FOUNDRY_LOCAL_FINISH_STOP` without referencing an undefined symbol.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    auto& conversation = it->second->state;
+    {
+      std::lock_guard<std::mutex> lock(conversation->mutex);
+      if ((flags & OgaEngineEventFlag_Token) != 0) {
+        conversation->tokens.push_back(event->Token());
+        ++conversation->sequence_length;
+      }
+      if ((flags & OgaEngineEventFlag_TurnFinished) != 0) {
```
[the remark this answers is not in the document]
bmehta001: Handled in 0d100b75. Request-less capacity events now evict a completed retained conversation and close its native request, freeing a dynamic batch slot. A later turn on that session receives a typed eviction signal and rebuilds from committed ChatSession history. Retryable events are also bounded so the worker cannot spin indefinitely.

sdk_v2/cpp/test/internal_api/chat/chat_session_test.cc
```diff
@@ -18,15 +18,83 @@
...
+namespace {
+
+class EngineModelStaging {
+ public:
+  EngineModelStaging(const std::filesystem::path& source, ChatBackendKind backend)
+      : path_(source.parent_path() / ("engine-chat-test-" + BackendName(backend) + "-" +
+                                      std::to_string(fl::test::CurrentPid()))) {
+    std::error_code ec;
+    std::filesystem::remove_all(path_, ec);
+    std::filesystem::create_directories(path_);
```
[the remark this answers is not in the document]
bmehta001: Handled in 0d100b75. EngineModelStaging now uses the existing TempPath helper under the writable system temp directory rather than writing beside FOUNDRY_TEST_DATA_DIR. Copied files are made owner-writable and the rewritten config stream is validated.

sdk_v2/cpp/test/internal_api/chat/chat_session_test.cc
```diff
@@ -18,15 +18,83 @@
...
+        }
+      }
+
+      const auto config_path = path_ / "genai_config.json";
+      std::ifstream input(config_path);
+      auto config = nlohmann::json::parse(input);
+      if (backend == ChatBackendKind::kDynamicEngine) {
+        config["engine"] = {
+            {"dynamic_batching", {{"max_batch_size", 2}, {"max_scheduled_tokens", 2048}}},
+        };
```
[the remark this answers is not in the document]
bmehta001: Handled in 0d100b75. The generic CPU fixture is no longer mislabeled as dynamic: it is staged only with static_batching, which is compatible with its shared-buffer KV layout. The concurrency, cancellation, and multi-turn reconstruction tests now explicitly assert the static Engine backend. CI does not currently provision a paged-KV model, so this avoids claiming unsupported dynamic coverage.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
```diff
@@ -0,0 +1,386 @@
...
+}
+
+bool OnnxChatEngine::EvictDormantConversation() {
+  for (auto it = conversations_.begin(); it != conversations_.end(); ++it) {
+    auto conversation = it->second->state;
+    {
+      std::lock_guard<std::mutex> lock(conversation->mutex);
+      if (!conversation->turn_finished) {
+        continue;
+      }
```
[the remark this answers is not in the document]
bmehta001: Handled in c3c80c94. Eviction now requires both turn_finished and a nonzero turn_id, so a newly created request cannot be selected before its first BeginTurn. Only conversations that have actually completed at least one turn are capacity candidates.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
```diff
@@ -0,0 +1,386 @@
...
+  for (size_t i = 0; i < event_buffer_->Count(); ++i) {
+    const auto* event = event_buffer_->Get(i);
+    const auto flags = event->Flags();
+    const auto request = event->Request();
+    if (!request) {
+      if ((flags & OgaEngineEventFlag_Failed) != 0) {
+        throw std::runtime_error("ORT GenAI Engine failed with error code " +
+                                 std::to_string(event->ErrorCode()));
+      }
+      if ((flags & OgaEngineEventFlag_CapacityBlocked) != 0 && EvictDormantConversation()) {
```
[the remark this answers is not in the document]
bmehta001: The requested dynamic integration test cannot be added to this CI suite with the models it provisions. `.pipelines/templates/fetch-models-from-artifacts-feed.yml:12-22` supplies only `qwen2.5-0.5b-instruct-generic-cpu-4` for this fixture, and its `genai_config.json` uses fixed `past_key_values.%d.*` / `present.%d.*` shared-buffer tensors rather than the paged-KV inputs required by dynamic batching. That is why 0d100b75 removed the invalid dynamic mutation and now explicitly asserts static Engine. Re-labeling this model dynamic would reproduce the prior invalid test, while adding a new model download violates the existing no-download fixture policy and requires a separately published paged-KV artifact. The capacity path is implemented against the documented Engine contract, but deterministic end-to-end coverage must follow when CI provisions a real paged-KV model. No safe test change is possible in this PR with the available fixture.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_engine_chat_generator.cc
```diff
@@ -0,0 +1,153 @@
...
+  FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_USAGE,
+           "Engine request rewind is unavailable; recreate the request from retained conversation history");
+}
+
+std::optional<ChatTurnUsage> OnnxEngineChatGenerator::GetTurnUsage() const {
+  const auto result = engine_.GetTurnResult(conversation_);
+  return ChatTurnUsage{
+      static_cast<int>(result.prompt_tokens + result.cached_prompt_tokens),
+      static_cast<int>(result.generated_tokens),
+  };
```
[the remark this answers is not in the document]
bmehta001: Addressed in f3469fb8. OnnxEngineChatGenerator::GetTurnUsage now carries TurnResult.finish_reason through MapFinishReason, including both generated-token and session-token limits as FOUNDRY_LOCAL_FINISH_LENGTH, and ChatSession::ProcessGeneratedOutput consumes that backend reason. This comment is resolved.

sdk_v2/cpp/src/inferencing/generative/genai_model_instance.cc
```diff
@@ -68,11 +69,25 @@ GenAIModelInstance::GenAIModelInstance(std::string model_id,
     FL_LOG_AND_THROW(logger, FOUNDRY_LOCAL_ERROR_INTERNAL,
                      "failed to create preprocessor for model ", model_id_, ": ", e.what());
   }
+
+  if (IsMultiModal() && genai_config_.GetChatBackendKind() != ChatBackendKind::kGenerator) {
+    FL_LOG_AND_THROW(logger, FOUNDRY_LOCAL_ERROR_INTERNAL,
+                     "model ", model_id_, " declares an Engine backend, but Engine is not supported for multimodal models");
```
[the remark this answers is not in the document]
bmehta001: Addressed in f3469fb8. The Engine validation diagnostic is wrapped within the repository's 120-character C++ line limit. This stale comment is resolved.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_engine_chat_generator.cc
```diff
@@ -0,0 +1,175 @@
...
+      conversation_(std::move(conversation)),
+      stream_(std::move(stream)),
+      stream_with_special_(std::move(stream_with_special)),
+      model_(model),
+      prompt_token_count_(prompt_token_count) {}
+
+OnnxEngineChatGenerator::~OnnxEngineChatGenerator() {
+  try {
+    engine_.Close(conversation_);
+  } catch (...) {
```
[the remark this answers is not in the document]
bmehta001: Addressed in 9c0575b5. ChatSession::~ChatSession explicitly resets cached_generator_ before model_.ReleaseSession(), so OnnxEngineChatGenerator closes its request while the model-owned Engine is still alive. This comment is resolved.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -48,6 +49,17 @@ void ApplyToolChoiceToContext(std::optional<flToolChoice> tool_choice, ToolCallC
...
 
+std::unique_ptr<ChatGenerator> CreateTextChatGenerator(const std::vector<MessageItem>& messages,
+                                                       const SearchOptions& options,
+                                                       GenAIModelInstance& model,
+                                                       const ToolCallContext& tool_ctx) {
+  if (model.GetGenAIConfig().GetChatBackendKind() != ChatBackendKind::kGenerator) {
+    return OnnxEngineChatGenerator::Create(messages, options, model, tool_ctx);
+  }
+
+  return OnnxChatGenerator::Create(messages, options, model, tool_ctx, /*use_full_context=*/true);
```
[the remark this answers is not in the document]
bmehta001: Addressed in 9c0575b5. Generator creation now takes an explicit use_full_context choice: retained session generation passes true, while the one-shot OpenAI JSON path passes false so its configured token limit remains bounded. This stale comment is resolved.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -474,6 +491,12 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
   int prompt_tokens = 0;
   int pre_turn_token_count = 0;
 
+  if (cached_generator_ &&
+      !cached_search_options_.HasSameRetainedGenerationSettings(effective_options)) {
+    cached_generator_.reset();
+    cached_tool_ctx_ = {};
+  }
```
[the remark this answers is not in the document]
bmehta001: Addressed in 9c0575b5. A rebuild from full history sets can_rewind_to_pre_turn=false unless history is empty; UndoTurns therefore discards that generator instead of rewinding it to token zero and losing prior context. This stale comment is resolved.

sdk_v2/cpp/test/internal_api/chat/chat_session_test.cc
```diff
@@ -144,6 +145,28 @@ TEST_F(ChatSessionTest, RunBasic) {
   EXPECT_EQ(session.GetHistory()[1].GetSimpleText(), text);
 }
 
+TEST_F(ChatSessionTest, ConcurrentIndependentSessions) {
```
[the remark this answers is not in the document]
bmehta001: Addressed in fbac2e5f and 0d100b75. The staged fixture explicitly selects and asserts the static Engine backend. ConcurrentIndependentSessions exercises two Engine requests, RunStreamingCancellation covers cancellation, and StaticEngineReconstructsMultiTurnHistory verifies retained-history reconstruction. This comment is resolved.

sdk_v2/cpp/test/test_main.cc
```diff
@@ -1,7 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-
 #include <gtest/gtest.h>
+#if __has_include(<ort_genai.h>)
+#include <ort_genai.h>
+#define FOUNDRY_LOCAL_TEST_HAS_OGA 1
+#endif
```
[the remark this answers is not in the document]
bmehta001: This target explicitly requires C++17 (`sdk_v2/cpp/test/CMakeLists.txt:175-176`), and `__has_include` is a standard preprocessor operator in C++17. The repository also requires C++20 globally (`sdk_v2/cpp/CMakeLists.txt:25-27`). A compiler that does not recognize `__has_include` is therefore below the supported language contract rather than a supported portability case. No change needed here.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -298,9 +298,10 @@ typedef enum flTensorDataType {
 #define FOUNDRY_LOCAL_PARAM_FREQUENCY_PENALTY "frequency_penalty"
 /// Float presence penalty. Currently only the neutral value 0 is supported.
 #define FOUNDRY_LOCAL_PARAM_PRESENCE_PENALTY "presence_penalty"
-#define FOUNDRY_LOCAL_PARAM_SEED "seed"                            ///< int. for reproducible outputs
-#define FOUNDRY_LOCAL_PARAM_EARLY_STOPPING "early_stopping"        ///< bool. whether to stop on stop sequence or only at max tokens
-#define FOUNDRY_LOCAL_PARAM_DO_SAMPLE "do_sample"                  ///< bool. whether to sample (false = greedy decoding)
+#define FOUNDRY_LOCAL_PARAM_SEED "seed"  ///< int. for reproducible outputs
```
baijumeswani: We should avoid calling the Generator path as legacy. It is not legacy as of now.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
```diff
@@ -0,0 +1,449 @@
...
+          }
+          existing_tokens = conversation->sequence_length;
+          conversation->tokens.clear();
+          conversation->error = nullptr;
+          conversation->result = {};
+          conversation->turn_finished = false;
+        }
+
+        auto plan = BuildEngineTurnOptionsPlan(options, tool_ctx, model_.GetGenAIConfig().GetChatBackendKind(),
+                                               kDefaultTextMaxOutputTokens);
```
baijumeswani: Does this limit the output tokens? Any reason to do that?

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
```diff
@@ -0,0 +1,449 @@
...
+    fail(error);
+    return;
+  }
+  command_cv_.notify_one();
+}
+
+void OnnxChatEngine::WorkerLoop(std::promise<void> initialized) {
+  try {
+    engine_ = OgaEngine::Create(model_.GetOgaModel());
+    event_buffer_ = engine_->CreateEventBuffer(model_.GetGenAIConfig().EngineMaxBatchSize().value_or(1) * 2);
```
baijumeswani: Can you explain here why the event buffer needs to be 2 times the max batch size?

sdk_v2/cpp/src/inferencing/generative/chat/onnx_engine_chat_generator.cc
```diff
@@ -0,0 +1,200 @@
...
+namespace {
+
+std::optional<flFinishReason> MapFinishReason(OgaFinishReason reason) {
+  switch (reason) {
+    case OgaFinishReason_Eos:
+#if defined(OgaFinishReason_StopSequence)
+    case OgaFinishReason_StopSequence:
+#elif defined(OgaFinishReason_StopString)
+    case OgaFinishReason_StopString:
+#endif
```
baijumeswani: This is not needed. We only release version with stopstring. No stop sequence was ever implemented.

sdk_v2/cpp/src/inferencing/generative/genai_config.h
```diff
@@ -3,12 +3,19 @@
...
+#include <cstddef>
 #include <optional>
 #include <string>
 #include <vector>
 
 namespace fl {
 
+enum class ChatBackendKind {
+  kGenerator,
+  kStaticEngine,
```
baijumeswani: you can assume that we do not support static engine batching for now. So, we can consider only 2 workloads, kGenerator and kEngine (meaning dynamic continuous batching).

sdk_v2/cpp/src/inferencing/generative/chat/chat_template.cc
```diff
@@ -63,6 +63,36 @@ std::string BuildChatPrompt(const std::vector<MessageItem>& messages,
   return model.GetPreprocessor().ApplyChatTemplate(messages_str.c_str(), tools_ptr, /*add_generation_prompt=*/true);
 }
 
+std::string BuildChatContinuationPrompt(const std::vector<MessageItem>& messages,
+                                        GenAIModelInstance& model,
+                                        const std::string& tools_json) {
+  const auto rendered_inputs = BuildChatPrompt(messages, model, tools_json);
+  constexpr std::string_view kMarkerPrefix = "__foundry_engine_assistant_boundary_";
```
baijumeswani: Could you help me understand what this is doing?

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
```diff
@@ -0,0 +1,449 @@
...
+
+size_t OnnxChatEngine::SequenceLength(const std::shared_ptr<Conversation>& conversation) const {
+  std::lock_guard<std::mutex> lock(conversation->mutex);
+  return conversation->sequence_length;
+}
+
+void OnnxChatEngine::Cancel(const std::shared_ptr<Conversation>& conversation) {
+  Enqueue(
+      [this, conversation]() {
+        auto& native = FindNative(conversation);
```
baijumeswani: Could we make a late cancel a no-op after the turn has finished? A cancellation can race with natural completion. If this dormant conversation was already evicted, `FindNative` throws and the fail callback below overwrites the successful result with an error. Please read `turn_finished` and `turn_id` under the conversation lock before `FindNative`, and return when the turn is already finished or was never started. A cancel-after-completion-and-eviction test would help.

baijumeswani (commented):
Thanks for the work here. The model-owned Engine and owner-thread design look like the right foundation for shared paged-KV coding workloads. I found a few items worth addressing before approval.

The main blocker is release packaging. The release path still pins ORT GenAI 0.15.2, while these Engine files use APIs that are only in the nightly today. NuGet does not publish 0.15.3 yet. I recommend either holding the merge until the stable package is available, or adding a compile-time Engine capability gate so 0.15.2 release builds remain Generator-only and Engine-configured models fail with a clear unsupported-runtime error. Please keep stable-0.15.2 and nightly-Engine CI lanes until activation.

For the coding scenario, the dynamic path also needs a real CI gate. The checked-in real-model tests use static Engine at two concurrent sessions. The manual A100 B4 result is useful, but I suggest adding a paged dynamic model test covering B1/B4/B8, retained continuation, capacity pressure, cancellation/recovery, output limits, and unload.

The packaging matrix is still running, so I would wait for it before approval.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
```diff
@@ -0,0 +1,449 @@
...
+        throw std::runtime_error("ORT GenAI Engine failed with error code " +
+                                 std::to_string(event->ErrorCode()));
+      }
+      if ((flags & OgaEngineEventFlag_CapacityBlocked) != 0 && EvictDormantConversation()) {
+        consecutive_retry_events_ = 0;
+        continue;
+      }
+      if ((flags & (OgaEngineEventFlag_CapacityBlocked | OgaEngineEventFlag_Retryable)) != 0) {
+        constexpr size_t kMaxConsecutiveRetries = 100;
+        if (++consecutive_retry_events_ > kMaxConsecutiveRetries) {
```
baijumeswani: This turns normal capacity pressure into a fatal error for the model-owned Engine. If all conversations are active, there is nothing to evict; after 101 events the worker throws and `FailAll` fails every session, including the foreground request. Could we queue or reject only the incoming request with a retryable busy/capacity error instead of stopping the whole Engine? Please add an over-capacity test where all resident requests are active.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
```diff
@@ -0,0 +1,449 @@
...
+                               std::to_string(event->ErrorCode())));
+        conversation->turn_finished = true;
+      }
+    }
+    conversation->cv.notify_all();
+  }
+}
+
+bool OnnxChatEngine::EvictDormantConversation() {
+  for (auto it = conversations_.begin(); it != conversations_.end(); ++it) {
```
baijumeswani, line 442: The first idle entry in an `unordered_map` is an arbitrary eviction victim. For foreground plus subagents, this can evict a recently used foreground conversation while keeping an older idle one. Please consider tracking last activity and evicting the least-recent idle conversation. It would also help to log or count evictions so full-context rebuild latency can be diagnosed.

sdk_v2/cpp/src/inferencing/generative/chat/search_options.cc
```diff
@@ -54,66 +242,49 @@ int ApplySearchOptions(const SearchOptions& options,
...
-  if (options.do_sample.has_value()) {
-    gen_params.SetSearchOptionBool("do_sample", *options.do_sample);
-  } else if (options.temperature.has_value()) {
-    gen_params.SetSearchOptionBool("do_sample", *options.temperature > 0.0f);
-  } else {
-    // Default: enable sampling (matches C# behavior)
-    gen_params.SetSearchOptionBool("do_sample", true);
+  // Preserve the model default when neither do_sample nor temperature was supplied. An explicit temperature retains
+  // the established Foundry behavior of selecting sampling above zero and greedy decoding at zero.
+  if (sampling.do_sample.has_value()) {
```
baijumeswani, line 263: This also changes the existing Generator backend: requests with no `do_sample` or `temperature` now inherit the model default instead of forcing sampling as before. That may be the better policy, but it changes behavior for models that do not use Engine batching. Could we either preserve the old Generator behavior or call out and qualify this fleet-wide behavior change in the PR description and release notes?

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -298,9 +298,10 @@ typedef enum flTensorDataType {
...
+/// Bool. Beam-search early-stopping policy; unsupported by Engine backends.
```
baijumeswani: Avoid references to engine/generator here. The API surface should not be aware of what backend was selected.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -48,6 +53,23 @@ void ApplyToolChoiceToContext(std::optional<flToolChoice> tool_choice, ToolCallC
...
 }
 
+std::unique_ptr<ChatGenerator> CreateTextChatGenerator(const std::vector<MessageItem>& messages,
+                                                       const SearchOptions& options,
+                                                       GenAIModelInstance& model,
+                                                       const ToolCallContext& tool_ctx,
+                                                       bool use_full_context) {
+  if (model.GetGenAIConfig().GetChatBackendKind() != ChatBackendKind::kGenerator) {
+#if FOUNDRY_LOCAL_OGA_HAS_DYNAMIC_ENGINE
+    return OnnxEngineChatGenerator::Create(messages, options, model, tool_ctx);
```
baijumeswani: nit: Let's not call the engine a generator.

baijumeswani (commented):
Thanks for addressing the earlier runtime concerns. Leaving the packaging failures aside, I found a few remaining items that would make this safer for long-running coding sessions.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
```diff
@@ -0,0 +1,459 @@
...
+        throw std::runtime_error("ORT GenAI Engine failed with error code " +
+                                 std::to_string(event->ErrorCode()));
+      }
+      if ((flags & OgaEngineEventFlag_CapacityBlocked) != 0 && EvictDormantConversation()) {
+        continue;
+      }
+      if ((flags & (OgaEngineEventFlag_CapacityBlocked | OgaEngineEventFlag_Retryable)) != 0) {
+        // Capacity pressure is transient while resident requests are active. Leave the request queued in OGA and
+        // return to the dispatcher so cancellation and close commands can still make progress.
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
baijumeswani, line 393: This can now wait forever when the Engine has no free capacity and there is no idle request to remove. The caller gets no timeout or busy error. Could we add a bounded wait or return a retryable capacity error for the waiting request? Please add a test where all Engine slots stay busy.

sdk_v2/cpp/test/CMakeLists.txt
```diff
@@ -87,6 +88,10 @@ if(FOUNDRY_LOCAL_HAS_EP_BOOTSTRAPPERS)
 endif()
 
 target_compile_options(foundry_local_tests PRIVATE ${FOUNDRY_LOCAL_COMPILE_OPTIONS})
+target_compile_definitions(
+    foundry_local_tests
+    PRIVATE FOUNDRY_LOCAL_OGA_HAS_DYNAMIC_ENGINE=$<BOOL:${FOUNDRY_LOCAL_OGA_HAS_DYNAMIC_ENGINE}>
```
baijumeswani, line 99: The Engine code is enabled for tests, but there is no test that creates a real dynamic Engine anymore. The earlier static Engine tests were removed and were not replaced. Could we add a small dynamic model test for basic generation, continuation, two concurrent sessions, capacity, cancellation, and unload?

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -578,60 +696,87 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
...
+  bool discard_generator = false;
 
   if (request.canceled) {
-    // Rewind the generator to undo this turn's input. The generator remains valid
-    // for the next attempt — the caller can re-send the same input.
-    cached_generator_->RewindTo(pre_turn_token_count);
+    if (cached_generator_->CanRewind() && can_rewind_to_pre_turn) {
+      cached_generator_->RewindTo(pre_turn_token_count);
+    } else {
+      discard_generator = true;
```
baijumeswani, line 722: Engine requests cannot rewind, so every canceled turn throws away the saved KV cache. The next request must process the full conversation again. This is safe, but it can be very slow for long coding sessions. Could we document this behavior and add a cancellation recovery performance test?

sdk_v2/cpp/src/inferencing/generative/genai_config.cc
```diff
@@ -113,6 +179,38 @@ GenAIConfig GenAIConfig::LoadFromFile(const std::string& path) {
...
+                            dynamic_batching.max_scheduled_tokens);
+      engine.dynamic_batching = dynamic_batching;
+    }
+
+    if (je.contains("static_batching") && !je["static_batching"].is_null()) {
+      FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL,
+               "genai_config.json engine.static_batching is not supported");
+    }
+
+    config.engine = std::move(engine);
```
baijumeswani, line 212: An explicit empty or misspelled `engine` object still falls back to the normal Generator path. That can silently disable batching. Could we reject an `engine` object that does not contain `dynamic_batching`, or at least log a clear warning?

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
bmehta001: Agreed that an enabled Engine backend must not wait indefinitely under capacity pressure. The shipping/package build is now pinned to OGA 0.15.2, where the dynamic Engine API is unavailable, and the CMake probe excludes this implementation entirely. I am deferring the bounded-capacity behavior and its all-slots-busy test to the dependency bump that enables this backend, so we do not add an unexecutable test against the supported runtime.

sdk_v2/cpp/test/CMakeLists.txt
bmehta001: The package graph is now consistently pinned to OGA 0.15.2. Its headers do not expose the dynamic Engine API, so CMake excludes the Engine sources and the Engine integration suite cannot be built or run in supported CI. The requested generation/continuation/concurrency/capacity/cancellation/unload coverage should land with the OGA version bump that enables the backend rather than being compiled only against an unpublished nightly.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
bmehta001: Documenting and measuring cancellation recovery makes sense when this backend is enabled. With this update all shipping builds use OGA 0.15.2, and the CMake API probe excludes the dynamic Engine implementation because that runtime does not support it. I am deferring the performance contract/test to the dependency bump so it runs against the runtime we actually ship.

sdk_v2/cpp/src/inferencing/generative/genai_config.cc
bmehta001: Fixed in bf4d249c. An engine object must now contain an object-valued dynamic_batching entry; empty, misspelled, null, scalar, and array forms are rejected instead of silently selecting the Generator path. The targeted GenAIConfig tests cover these cases.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
bmehta001: I was adding 1 event for a terminal engine failure, but that should be `max_batch_size` + 1, not *2. But I will keep it as `max_batch_size` to keep it simple -> we can fetch any remaining events as part of a future batch.

bmehta001: Updated in 7a0b2547. Token and terminal payloads can share one event (`Token | TurnFinished`), so they do not require two slots per request. OGA retains speculative or fatal overflow and drains it on later `Run` calls. The buffer is now `max_batch_size`, which matches upstream usage and drains an ordinary full decode batch in one call.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
bmehta001: Agreed. Renamed the concrete adapter and files to `OnnxEngineChatStream` in af16203a. It still implements the shared `ChatGenerator` interface consumed by `ChatSession`, but the concrete type no longer calls the Engine itself a generator.

sdk_v2/cpp/src/inferencing/generative/chat/chat_template.cc
bmehta001: This is a hacky workaround to try to get the assistant end delimiter since the engine does not retain it, so we add it ourselves before adding a new turn to the request. - Maybe this should be moved to GenAI

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
bmehta001: Addressed in db73ecff. The public C API now describes early_stopping only by its supported value (false), without calling either inference path legacy or naming a backend. See sdk_v2/cpp/include/foundry_local/foundry_local_c.h:301.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
bmehta001: Addressed in db73ecff. EngineTurnOptionsPlan::max_generated_tokens is now optional, and SetMaxGeneratedTokens is called only when the user explicitly supplies max_output_tokens. Omitted limits defer to OGA's Request/session context ceiling; explicit nonpositive values remain invalid. See search_options.cc:166-172 and onnx_chat_engine.cc:20-24.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_engine_chat_generator.cc
bmehta001: Addressed on the current branch. The finish-reason mapping now uses only OgaFinishReason_StopString, matching the Engine API compiled in the 0.16 nightly lane. See onnx_engine_chat_stream.cc:15-28.

sdk_v2/cpp/src/inferencing/generative/genai_config.h
bmehta001: Addressed on the current branch. ChatBackendKind now has only kGenerator and kEngine, and genai_config.json explicitly rejects engine.static_batching. Engine means dynamic batching. See genai_config.h:12-15 and genai_config.cc:177-195.

sdk_v2/cpp/src/inferencing/generative/chat/chat_template.cc
bmehta001: Documented in db73ecff. ApplyChatTemplate only renders complete conversations, so a collision-free synthetic assistant marker creates a template-aware cut point. Everything through the marker is removed before tokenization; the retained suffix includes the assistant boundary and new turn without repeating BOS/default-system framing. See chat_template.h:42-49 and chat_template.cc:70-84.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
bmehta001: Addressed and covered in db73ecff. Cancel now reads turn_finished and turn_id under the conversation lock before native lookup and returns for completed or never-started turns. DynamicEngineChatTest.LateCancelAfterCompletedConversationRemovalIsNoOp verifies the successful result survives close/removal plus late cancellation.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
bmehta001: Addressed in db73ecff. CapacityBlocked no longer becomes a model-wide fatal error. The dispatcher continues servicing commands, waits up to the configured deadline, then cancels/closes only the newest waiting no-progress conversation; active peers remain usable. DynamicEngineChatTest.CapacityTimeoutFailsOnlyNewestWaitingConversation covers all slots active.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
bmehta001: Addressed in db73ecff. Dormant eviction now scans last_activity and selects the least-recently-active completed conversation instead of relying on unordered_map iteration order; turn activity is updated on admission and every routed event. See onnx_chat_engine.cc:425-451. No arbitrary victim remains.

sdk_v2/cpp/src/inferencing/generative/chat/search_options.cc
bmehta001: Preserved on the current branch. ApplySearchOptions still explicitly sets do_sample=true when Generator callers omit a sampling policy, while Engine turn options forward only caller-supplied values. See search_options.cc:252-259 and ApplyEngineTurnOptions in onnx_chat_engine.cc:20-42.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
bmehta001: Addressed in db73ecff. Public early_stopping documentation is backend-neutral and states only the observable contract: only false is currently supported. See foundry_local_c.h:301-302 and foundry_local_cpp.h:949.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
bmehta001: Addressed in db73ecff. Capacity waiting is bounded by a configurable timeout (30 seconds by default); expiration fails and closes only the newest waiting no-progress turn, without poisoning the shared Engine or active conversations. The dynamic Engine test uses a 100 ms deadline to verify this deterministically.

sdk_v2/cpp/test/CMakeLists.txt
bmehta001: Addressed in db73ecff. DynamicEngineChatTest is compiled only when the Engine capability probe succeeds and uses a real pre-staged paged/dynamic model supplied by FOUNDRY_LOCAL_DYNAMIC_ENGINE_TEST_MODEL_PATH. It covers output limits and continuation, two concurrent sessions, capacity, late cancellation, recovery, and unload. The CI policy now requires Engine-enabled lanes to provide the fixture so skips cannot hide regressions.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
bmehta001: Addressed in db73ecff. ChatSession documentation now explains that cancellation reuses pre-turn state when safely restorable and otherwise discards/rebuilds from committed history. DynamicEngineChatTest.CancellationRebuildsCommittedHistoryWithinBudget verifies history correctness and a bounded recovery time after cancellation.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_engine.cc
```diff
@@ -0,0 +1,347 @@
...
+
+    auto it = std::find_if(conversations_.begin(), conversations_.end(), [&](const auto& entry) {
+      return entry.second->request.get() == &request->get();
+    });
+    if (it == conversations_.end()) {
+      continue;
+    }
+
+    auto& conversation = it->second->state;
+    const auto flags = event->Flags();
```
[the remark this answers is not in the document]
bmehta001: Addressed on the current branch. RouteEvents now treats a request-less Failed event as fatal and throws with the Engine error code; WorkerLoop catches it and calls FailAll so every waiter is released. See onnx_chat_engine.cc:360-370 and 342-351.

sdk_v2/cpp/test/test_main.cc
```diff
@@ -1,7 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-
 #include <gtest/gtest.h>
+#include <gtest/gtest.h>
+#include <gtest/gtest.h>
```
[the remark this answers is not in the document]
bmehta001: Addressed on the current branch. sdk_v2/cpp/test/test_main.cc now contains one gtest/gtest.h include, followed only by the conditional ort_genai.h include.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -517,18 +543,18 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
...
     }
     prompt_tokens = generator->PromptTokenCount();
 
     cached_generator_ = std::move(generator);
     cached_tool_ctx_ = std::move(tool_ctx);
+    cached_search_options_ = effective_options;
   }
 
-  int max_output = effective_options.max_output_tokens.value_or(0);
+  const int max_output = ResolveMaxOutputTokens(effective_options);
```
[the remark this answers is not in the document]
bmehta001: Addressed on the current branch. Generator-backed text still uses the established resolved host limit, but reaching it sets host_output_limit_reached and ResolveGeneratedFinishReason maps that condition to LENGTH. Engine text does not enforce that host default; db73ecff forwards a limit only when the caller supplied max_output_tokens.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -96,9 +112,113 @@ void AppendGeneratedSegment(std::vector<GeneratedOutputEvent>& destination, std:
...
+    }
+
+    return false;
+  }
+
+  auto filtered = stop_filter->Push(fragment);
+  if (!filtered.empty()) {
+    // Filtering may combine or shorten decoded token fragments, so the current token ID is no longer aligned.
+    process_segments(splitter.Push(filtered));
+  }
```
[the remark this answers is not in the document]
bmehta001: Addressed on the current branch. PushDecodedFragment now uses Push(token_id, text) whenever stop filtering returns token-aligned output, and falls back to text-only Push only when buffering/truncation breaks alignment. See chat_session.h:56-63; the decoded-stream tests cover both aligned and buffered paths.

sdk_v2/cpp/src/inferencing/generative/genai_config.cc
```diff
@@ -113,6 +166,46 @@ GenAIConfig GenAIConfig::LoadFromFile(const std::string& path) {
     config.search = std::move(search);
   }
 
+  if (j.contains("engine") && j["engine"].is_object()) {
+    const auto& je = j["engine"];
```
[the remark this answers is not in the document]
bmehta001: Addressed on the current branch. GenAIConfig::FromJson now rejects any non-object engine value and requires object-valued engine.dynamic_batching instead of silently selecting Generator. See genai_config.cc:170-198.

sdk_v2/cpp/src/inferencing/generative/genai_config.cc
```diff
@@ -4,9 +4,33 @@
...
+  if (!value.is_number_integer()) {
+    FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL,
+             std::string("genai_config.json engine.") + name + " must be a positive integer");
+  }
+
+  const auto parsed = value.get<int64_t>();
+  if (parsed <= 0 || static_cast<uint64_t>(parsed) > std::numeric_limits<size_t>::max()) {
+    FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL,
+             std::string("genai_config.json engine.") + name + " must be a positive integer");
+  }
```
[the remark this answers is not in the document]
bmehta001: Addressed on the current branch. ParsePositiveSize now receives the full object path, and dynamic fields report genai_config.json engine.dynamic_batching.<field>. See genai_config.cc:35-50 and 195-202.

sdk_v2/cpp/src/inferencing/generative/chat/stop_strings.cc
```diff
@@ -0,0 +1,280 @@
...
+                 std::to_string(kMaxSerializedStopStringsBytes) + " serialized bytes");
+  }
+
+  nlohmann::json stop_json;
+  try {
+    stop_json = nlohmann::json::parse(serialized_view);
+  } catch (const nlohmann::json::parse_error& e) {
+    FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL,
+             std::string("internal stop strings option is not valid JSON: ") + e.what());
+  }
```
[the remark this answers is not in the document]
bmehta001: Addressed on the current branch. LoadStopStringsOption maps nlohmann parse_error to FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT with an actionable invalid-JSON message. See stop_strings.cc:166-174.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -426,51 +548,67 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
...
+        cached_tool_ctx_.guidance_type != turn_tool_ctx.guidance_type ||
+        cached_tool_ctx_.guidance_data != turn_tool_ctx.guidance_data;
+
+    // User guidance is a finite grammar even when tool output is not required. Rebuild after such a turn because
+    // completion latches IsDone(), and rebuild whenever the explicit schema or any retained search setting changes.
+    if (chat_session_internal::ShouldRebuildRetainedGeneratorBeforeAppend(
+            backend_kind,
+            prev_needs_guidance != curr_needs_guidance || prev_has_user_guidance,
+            guidance_payload_changed,
+            !cached_search_options_.HasSameRetainedGenerationSettings(effective_options))) {
```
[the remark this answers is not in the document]
bmehta001: Addressed on the current branch. HasSameRetainedGenerationSettings returns true immediately for kEngine because Engine-supported settings, including seed, are applied per BeginTurn; seed comparison remains only for Generator state. See search_options.cc:371-382.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -426,51 +563,69 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
...
     } else {
       // Continuous decoding: append only the new messages to the existing generator.
       pre_turn_token_count = cached_generator_->TokenCount();
-      prompt_tokens = cached_generator_->AppendMessages(new_messages, Model(), cached_tool_ctx_.tools_json);
-
-      // Refresh per-turn fields (tool_choice, guidance) while keeping session-level definitions stable.
-      UpdateToolContextForTurn(request, cached_tool_ctx_);
+      try {
+        prompt_tokens = cached_generator_->AppendMessages(new_messages, Model(), turn_tool_ctx, effective_options);
+
```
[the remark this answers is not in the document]
bmehta001: Addressed on the current branch. The retained append path records prompt_tokens as pre_turn_token_count + appended_prompt_tokens, so Generator total minus prompt counts only this turn output. Engine usage replaces both prompt and total from the native TurnResult. See chat_session.cc:557-562 and 727-732.

.pipelines/foundry-local-packaging.yml
```diff
@@ -53,17 +53,13 @@ parameters:
...
 - name: cppOrtVersion
   value: '1.28.0'
-- ${{ if eq(parameters.isRelease, true) }}:
-  - name: cppGenaiVersion
-    value: '0.15.2'
-- ${{ else }}:
-  - name: cppGenaiVersion
-    value: '0.16.0-dev1001400138'
+- name: cppGenaiVersion
+  value: '0.15.2'
```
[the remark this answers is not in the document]
bmehta001: Fixed in the PR description. The release and ordinary CI path intentionally remains pinned to 0.15.2; CMake capability-gates Engine sources/tests there. The implementation was separately compiled against cached 0.16.0-dev1001407373, and the description now states that accurately instead of claiming a pipeline nightly override.

sdk_v2/cpp/src/inferencing/generative/genai_config.cc
```diff
@@ -113,6 +179,39 @@ GenAIConfig GenAIConfig::LoadFromFile(const std::string& path) {
...
+
+  if (j.contains("engine")) {
+    const auto& je = j["engine"];
+    Engine engine;
+
+    if (je.contains("static_batching") && !je["static_batching"].is_null()) {
+      FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL,
+               "genai_config.json engine.static_batching is not supported");
+    }
```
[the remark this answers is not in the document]
bmehta001: Fixed in the PR description. The supported contract is Generator or dynamic Engine only; engine.static_batching is explicitly rejected. The backend table and summary no longer claim static batching support.

sdk_v2/cpp/test/internal_api/chat/chat_template_test.cc
```diff
@@ -122,6 +122,31 @@ TEST_F(ChatTemplateTest, PromptEndsWithAssistantPrefix) {
       << "Prompt should end with assistant prefix for generation. Got: " << prompt;
 }
 
+TEST_F(ChatTemplateTest, EngineContinuationIncludesAssistantTurnBoundary) {
+  std::vector<MessageItem> messages = {{FOUNDRY_LOCAL_ROLE_USER, "What is the codeword?"}};
+
+  std::string prompt = BuildChatContinuationPrompt(messages, GetModel());
+
+  EXPECT_EQ(prompt.find("__foundry_engine_assistant_boundary__"), std::string::npos);
```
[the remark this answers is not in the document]
bmehta001: Fixed in 2d18ca27. The assertion now searches for the generated marker prefix __foundry_engine_assistant_boundary_, so every numeric suffix is covered. Both continuation-marker tests pass on the shipping-runtime Debug build.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -578,60 +678,87 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
...
+  if (request.canceled) {
+    cached_generator_->Cancel();
+  } else if ((stop_sequence_matched || host_output_limit_reached) && !cached_generator_->IsDone()) {
+    cached_generator_->Cancel();
+  }
+
   int total_tokens = cached_generator_->TokenCount();
+  std::optional<flFinishReason> backend_finish_reason;
+  if (const auto turn_usage = cached_generator_->GetTurnUsage()) {
+    prompt_tokens = turn_usage->prompt_tokens;
```
baijumeswani, line 712: Engine usage still replaces the complete logical prompt count with the current `BeginTurn` suffix count. `OnnxEngineChatStream::GetTurnUsage()` maps OGA `prompt_tokens + cached_prompt_tokens`, but under the current OGA contract `prompt_tokens` is only the IDs admitted by this turn and `cached_prompt_tokens` is currently zero. A warm continuation therefore reports fewer prompt tokens than an equivalent fresh full-history request. Please preserve/map the pre-turn resident length when producing the public usage, and add a warm-versus-fresh assertion that `prompt_tokens` matches while `completion_tokens` contains only the current turn output.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.h
```diff
@@ -1,24 +1,120 @@
...
+    return false;
+  }
+
+  auto filtered = stop_filter->PushWithTokenAlignment(fragment);
+  if (!filtered.text.empty()) {
+    if (filtered.token_aligned && token_id.has_value()) {
+      process_segments(splitter.Push(*token_id, std::move(filtered.text)));
+    } else {
+      // Filtering combined or shortened decoded token fragments, so the current token ID is no longer aligned.
+      process_segments(splitter.Push(filtered.text));
```
baijumeswani, line 62: This fallback preserves text correctness but loses token-count provenance. If stop filtering buffers a partial stop prefix and later releases text spanning multiple generated tokens, `Push(filtered.text)` creates one splitter text-token record, so `ReasoningTokenCount()` undercounts the actual reasoning tokens. Please carry the contributing token count/IDs through `StopStringFilter` (or otherwise account for them separately) and add a partial-stop-prefix-inside-reasoning test that asserts exact reasoning-token usage.

baijumeswani (commented):
Correctness review of the current Engine integration. I found two usage-accounting issues that should be addressed before the Engine path is enabled for release; omitting the separate CI-lane concern as requested.

baijumeswani (approved):

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
baijumeswani: let's address in a follow up pr.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.h
baijumeswani: let's address in a follow up pr.

PR #1083, opened 2026-09-08
Preserve tool calls and reasoning boundaries across chat turns

baijumeswani:
## Summary

Preserve tool calls, tool results, visible text, and reasoning as structured conversation history.

This makes a continued conversation independent of whether Foundry Local still has a warm session in memory or has to rebuild the conversation from stored Responses data.

The change also makes failed turns transactional, keeps cached prompts aligned with current instructions and tools, preserves long response chains beyond the store's metadata capacity, and fixes the C# tests and sample to send complete self-contained tool conversations.

## Why

Coding agents depend on long, structured conversations.

A typical turn may include:

- visible assistant text;
- private reasoning;
- one or more tool calls;
- results returned in a different order;
- an empty tool result;
- a later continuation after the live session has been evicted.

Foundry Local previously flattened or dropped parts of this history.

That could leave a tool result without the call that created it, remove an empty result, lose a call ID, replay only the newest Responses hop, leak reasoning as visible text, or reuse a cached prompt that no longer matched the recorded conversation.

### Multi-step coding workflows

A coding harness can now preserve a sequence such as:

1. The assistant calls `search_code`.
2. The assistant calls `read_file` twice in parallel.
3. The harness returns both results.
4. The assistant proposes an edit.
5. The harness runs tests and returns an empty successful result.
6. The assistant continues from the complete exchange.

Each result remains associated with the correct call ID.

### Warm and cold continuation

A short conversation may continue from a live `ChatSession` and its KV cache.

A long or inactive conversation may continue after that session has been evicted.

Both paths now rebuild the same logical prompt, including assistant turn boundaries, tool calls, tool results, and visible text.

### Long Responses conversations

A Responses conversation can exceed the default 20 stored response metadata entries and still continue after a session-cache miss.

Evicted metadata is compacted into shared replay history held by retained descendants.

The old response no longer appears in lookup or pagination, but the conversation history needed by a retained response remains reconstructable.

### Reasoning models

Models whose chat template opens a reasoning block in the prompt can return private reasoning without leaking the opening or closing markers into visible output.

Reasoning remains available as typed output but is not injected back into later prompts.

## Structured conversation transcript

The session now stores provider-neutral transcript messages with ordered entries.

An assistant turn can contain:

```text
visible text
reasoning
tool call
another tool call
```

Each tool call retains:

- call ID;
- tool name;
- original argument bytes;
- normalized object arguments used by the model template.

Tool results retain the call ID they answer.

Foundry Local no longer reconstructs tool history by manually concatenating model-specific marker strings.

The package chat template receives structured messages and owns the final rendering.

## Tool call and result validation

The transcript tracks issued and outstanding call IDs.

Foundry Local rejects:

- an empty tool call ID;
- a duplicate tool call ID;
- a tool result with no call ID;
- a result for an unknown call;
- a second result for an already answered call;
- a result that appears before its call;
- caller-supplied function arguments that are not a JSON object.

Parallel results may arrive in a different order from their calls because correlation uses the call ID rather than position.

An empty tool result remains a real result and is preserved.

## Preserve assistant turn boundaries

Stored Responses data now keeps one input and output pair per response hop.

Assistant items are merged only inside the same recorded segment.

This preserves important boundaries such as:

- a reasoning-only assistant turn;
- an assistant turn with no visible output;
- parallel calls from one assistant turn;
- a later assistant prefill;
- two adjacent empty turns.

Cold replay cannot accidentally merge two recorded assistant turns or remove an empty boundary.

## Complete Responses replay

On a session-cache miss, Foundry Local walks the complete `previous_response_id` chain and rebuilds every hop from oldest to newest:

```text
oldest input
oldest output
next input
next output
...
current input
```

This keeps a tool call before the result that answers it.

Missing links and cycles fail explicitly instead of silently running inference with a truncated conversation.

The `/input_items` endpoint continues to return only the input stored for that individual response.

## Long-chain compaction

The response store still limits normal response metadata to its configured capacity.

When a complete conversation root is evicted, Foundry Local stores its minimal replay input and output in an immutable shared prefix used by retained child responses.

This gives long and branched conversations the following behavior:

- a retained tip can reconstruct more than 20 hops;
- branches share common compacted history instead of copying it;
- repeated compaction preserves order;
- eviction stays within the least recently used conversation;
- malformed cycles cannot monopolize the store;
- replacing a response preserves history only when its parent is unchanged;
- reparenting cannot leak history from the old conversation.

The metadata capacity bounds response lookup and listing, not the replay history required by retained conversations.

Replay history remains resident while a retained descendant needs it and stops growing when the rebuilt conversation can no longer complete a successful model turn.

## Delete behavior

Deleting a response also removes every retained descendant whose replay depends on it.

This applies when the deleted response still has normal metadata and when it survives only in compacted replay history.

Cached sessions for the deleted response and its descendants are evicted as well.

Deleted content therefore cannot remain available through a warm transcript or a compacted cold replay path.

## Warm and cold prompt parity

The cached and rebuilt paths now follow the same rules.

The cached generator is invalidated when:

- request instructions change;
- tool definitions change or are removed;
- the current input is empty and requires full context;
- the current input carries a tool call or result;
- the prior turn generated a tool call;
- reasoning, grammar, media, or an incomplete generated turn requires reconstruction.

Tool definitions are compared by both rendered schema and authoritative kind.

This prevents cache residency from changing the prompt or the interpretation of a generated call.

## Request-scoped instructions

Responses `instructions` are request-scoped.

They are applied as the current system prefix but are not stored as synthetic conversation messages.

This prevents:

- instructions accumulating once per turn;
- old instructions surviving after they are removed;
- caller-supplied system messages being mistaken for generated instructions;
- warm and cold sessions retaining different instruction history.

Caller-supplied system and developer messages remain part of the conversation and are replayed unchanged.

## Reasoning isolation

Reasoning is typed separately from visible text.

It can be returned to the caller but is never projected back into the model prompt.

This avoids feeding private scratchpad text into later turns.

The prompt-opened reasoning check uses the rendered prompt and marker token information.

This prevents marker-shaped user text or tool output from falsely placing generation inside a reasoning block.

When the prompt opens reasoning:

- generated scratchpad text is classified as reasoning;
- generated reasoning markers are suppressed;
- marker tokens are not counted as reasoning content;
- Chat Completions keeps reasoning out of visible `content`;
- Responses emits reasoning and visible output as separate typed items.

Configured marker overrides and default token IDs are kept from different sources so a text override cannot be paired with an unrelated token ID.

## Generated event ordering

The provider transcript format can represent visible assistant text followed by tool calls.

It cannot faithfully represent new visible text after a tool call in the same assistant turn.

Foundry Local therefore ends the generated turn at the tool call rather than reordering later text before the call.

Whitespace between parallel tool-call blocks is dropped rather than moved.

This prevents a generated sequence from being replayed in an order that never occurred.

## Transactional turns

A turn is staged and validated before it changes committed history.

If generation, parsing, streaming, or response construction fails, the failed turn is not committed.

If an appended turn is cancelled, Foundry Local rewinds to the known pre-turn token boundary.

If the turn created or rebuilt the generator and has no safe rewind point, the generator is discarded.

The next request can never reuse generator state that is ahead of the logical transcript.

## Safe undo

Undo rebuilds outstanding tool-call state from the remaining transcript.

A token rewind point is reused only when all removed turns belong to the current generator history.

If undo crosses a generator rebuild, the generator is discarded and recreated on the next request.

Undo and request processing are serialized so they cannot mutate transcript or generator state concurrently.

## Media consistency

Media is accepted only on a history-free conversation.

Media cannot be combined with tools because a later tool-result continuation cannot replay the original media bytes from stored history.

Direct image-only and audio-only input still count as content.

These rules are applied consistently whether a conversation is warm or rebuilt.

## Pagination

Responses list pagination now derives `has_more` from whether another stored item actually exists.

An exactly full final page reports `has_more: false` instead of advertising an empty next page.

## C# Chat Completions corrections

Chat Completions JSON requests are self-contained.

When a C# caller returns a tool result, the request must replay the assistant message that issued the call and set the matching `tool_call_id` on the result.

The C# integration tests and tool-calling sample now send that complete shape.

Model-free serialization tests verify:

- the assistant tool call is present;
- the call ID is preserved;
- the tool result carries the same ID;
- a normal user, assistant, user transcript remains ordered and complete.

The streaming sample now captures the actual tool-call delta rather than the later finish chunk, which carries no call payload.

## Managed CI correction

The C# test project targets `net8.0` on every platform and also targets `net462` on Windows.

The pipeline previously requested `net9.0`, which the test project does not declare.

`dotnet test --no-build --framework net9.0` exited successfully while running zero tests.

The pipeline now runs:

- `net8.0` on Linux, macOS, and Windows;
- `net462` in addition to `net8.0` on Windows.

It validates each requested framework against the project before running tests so a future mismatch fails clearly instead of producing a false green.

The streaming chat test was also changed to test streaming and multi-turn replay rather than making a second arithmetic answer from a small model a release gate.

## Example

Given:

```json
[
  {
    "role": "user",
    "content": "Read both files"
  },
  {
    "role": "assistant",
    "content": null,
    "tool_calls": [
      {
        "id": "call_alpha",
        "type": "function",
        "function": {
          "name": "read_file",
          "arguments": "{\"path\":\"a.txt\"}"
        }
      },
      {
        "id": "call_beta",
        "type": "function",
        "function": {
          "name": "read_file",
          "arguments": "{\"path\":\"b.txt\"}"
        }
      }
    ]
  },
  {
    "role": "tool",
    "tool_call_id": "call_beta",
    "content": "B"
  },
  {
    "role": "tool",
    "tool_call_id": "call_alpha",
    "content": "A"
  }
]
```

Foundry Local now preserves both calls, accepts the results in their supplied order, associates each result by ID, and passes the complete structured conversation to the package chat template.

The same logical conversation is used when the session is live, when it is rebuilt from stored responses, and when early response metadata has been compacted after a long chain.


sdk_v2/cpp/src/service/responses_handler.cc
```diff
@@ -190,20 +188,42 @@ std::shared_ptr<HttpRequestHandler::OutgoingResponse> ResponsesHandler::handle(
     if (session) {
       ctx_.logger.Log(LogLevel::Information,
                       fmt::format("CreateResponse: reusing cached session from '{}'", *params.previous_response_id));
+
+      // The session already holds the conversation, but keep its stored hops warm: if this session is later evicted,
+      // rebuilding the chain from the store is the only way to continue.
+      if (!ctx_.response_store.TouchChain(*params.previous_response_id)) {
```
bmehta001: **High:** This cached session is selected only by `previous_response_id`, without checking that it belongs to the model resolved for the current request. Continuing a response created with model A while requesting model B will run on A when this cache entry is warm, but rebuild and run on B after eviction; the response is also labeled as B in the warm case. Please validate the cached session model and either rebuild it with the requested model or reject cross-model continuation consistently.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -598,43 +646,57 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
...
   int total_tokens = cached_generator_->TokenCount();
 
-  if (request.canceled) {
-    // Rewind the generator to undo this turn's input. The generator remains valid
-    // for the next attempt — the caller can re-send the same input.
-    cached_generator_->RewindTo(pre_turn_token_count);
+  auto assistant_message = MakeAssistantMessage(generated_events, logger_);
+  const bool generated_tool_calls = assistant_message.HasToolCalls();
+
+  if (!request.canceled) {
```
bmehta001, line 769: **Medium:** `request.canceled` can still change after this check because streaming callbacks are dispatched asynchronously and are not drained until the callback handler is destroyed. A callback that returns nonzero late can therefore cancel the request after `CommitTurn` has already run, violating the guarantee that canceled turns commit nothing. Please drain all queued callbacks before the final cancellation check and commit decision.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -371,127 +410,141 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
...
+  });
 
-    std::vector<MessageItem> all_messages;
-    all_messages.reserve(history_.size() + new_messages.size());
-    all_messages.insert(all_messages.end(), history_.begin(), history_.end());
-    all_messages.insert(all_messages.end(), new_messages.begin(), new_messages.end());
+  if (cached_generator_) {
+    // Continuous decoding: append only the new messages to the existing generator.
+    pre_turn_token_count = cached_generator_->TokenCount();
+    prompt_tokens = cached_generator_->AppendMessages(inputs, Model(), cached_tool_ctx_.tools_json);
```
bmehta001: **Medium:** On this warm-continuation path, `prompt_tokens` contains only the newly appended suffix, while the later `total_tokens` value is cumulative for the generator. `ProcessGeneratedOutput` subtracts the former from the latter, so previous conversation tokens are reported as output tokens for this turn and can incorrectly trigger `finish_reason="length"`. Please calculate generated tokens from `total_tokens - pre_turn_token_count`, while reporting the full logical input token count separately.

sdk_v2/cpp/src/service/responses_handler.cc
```diff
@@ -733,8 +777,8 @@ std::shared_ptr<HttpRequestHandler::OutgoingResponse> DeleteResponseHandler::han
 
   ctx_.logger.Log(LogLevel::Debug, fmt::format("DeleteResponse: responseId={}", std::string(id->c_str())));
 
-  bool deleted = ctx_.response_store.Delete(id->c_str());
-  if (!deleted) {
+  auto deleted_ids = ctx_.response_store.DeleteWithDependents(id->c_str());
```
bmehta001, line 887: **Medium:** Chain deletion races with an in-flight continuation. A continuation can check out/touch parent A, this call can delete A while its session is checked out, and the continuation can then store child B and check its session back in. B retains deleted history and becomes unreconstructable after eviction. Please coordinate continuation admission/storage with deletion (for example, pin or tombstone the chain) so no descendant can be committed or cached after its ancestor is deleted.

bmehta001 (commented):
Found four material correctness issues involving model/session consistency, cancellation commit ordering, warm-session token accounting, and concurrent chain deletion.

sdk_v2/cpp/src/service/responses_handler.cc
baijumeswani: Addressed in dc35ef99. Stored responses are now bound to the resolved catalog model ID, and continuation validates that identity before choosing the warm cached-session or cold replay path. A real two-model HTTP regression verifies cross-model continuation returns 400.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
baijumeswani: Addressed in dc35ef99. ChatSession now drains and joins the asynchronous callback handler before the final cancellation check, response construction, and transcript commit. A delayed-last-callback regression verifies cancellation commits no turn.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
baijumeswani: Addressed in dc35ef99. Warm continuation now records the complete post-append input sequence as prompt usage, so total minus prompt is only the current generation. Tests pin bounded completion usage and a natural warm stop rather than a false length finish.

sdk_v2/cpp/src/service/responses_handler.cc
baijumeswani: Addressed in dc35ef99. Response leases snapshot the chain at admission; deletion invalidates dependent in-flight leases; and metadata plus session-cache admission/removal are coordinated under one store critical section. Threaded, compacted-ancestor, branch, and delete-after-commit regressions cover both race orderings.

sdk_v2/cpp/src/inferencing/generative/openresponses/response_store.cc
```diff
@@ -4,32 +4,173 @@
...
+  }
+
+  const std::string response_id = response.id;
+  StoreLocked(std::move(response));
+
+  // Admit under the same lock that published the metadata. A delete that arrives after this point sees the stored
+  // entry, so it removes the metadata and drops the cached session together; one that arrived earlier invalidated
+  // the lease and never reaches here. There is no ordering in between for a session to be resurrected in.
+  if (admission != nullptr) {
+    admission->Admit(response_id);
```
sayanshaw24, line 250: could Admit() throw? StoreLocked() has already published the response and consumed the lease at this point, so an exception from session-cache insertion would make the handler report failure while leaving the response retrievable. we can maybe consider making Admit() noexcept with a defined fallback, or rolling back the stored entry if admission fails, so metadata and session publication remain transactional.

bmehta001 (dismissed):

sdk_v2/cpp/src/inferencing/generative/openresponses/response_store.cc
baijumeswani: addressed this.

bmehta001 (approved):

PR #1087, opened 2026-09-09
Add text custom tools to Chat Completions and Responses

baijumeswani:
## Summary

Add provider-layer support for text custom tools to the OpenAI-compatible Chat Completions and Responses APIs.

A function tool receives a structured JSON object. A text custom tool receives one free-form string payload. This PR
preserves that distinction across request parsing, prompt construction, generated output, streaming, storage, replay,
and continuation.

### Chat Completions

```json
{
  "type": "custom",
  "custom": {
    "name": "edit",
    "description": "Replace file content",
    "format": {
      "type": "text"
    }
  }
}
```

A generated call is returned as:

```json
{
  "id": "call_abc123",
  "type": "custom",
  "custom": {
    "name": "edit",
    "input": "replacement text"
  }
}
```

### Responses

```json
{
  "type": "custom",
  "name": "edit",
  "description": "Replace file content",
  "format": {
    "type": "text"
  }
}
```

A completed call is returned as:

```json
{
  "id": "ctc_abc123",
  "type": "custom_tool_call",
  "call_id": "call_abc123",
  "name": "edit",
  "input": "replacement text",
  "status": "completed"
}
```

The caller can continue the Responses conversation with a correlated result:

```json
{
  "previous_response_id": "resp_abc123",
  "input": [
    {
      "type": "custom_tool_call_output",
      "call_id": "call_abc123",
      "output": "Edit applied"
    }
  ]
}
```

## Behavior

- Supports mixed function and text custom tools in one request.
- Supports automatic, required, and explicitly forced choices at the provider-contract level.
- Preserves the effective tool name and kind in an immutable request snapshot.
- Uses an internal synthesized `{"input": "<text>"}` schema so existing model templates can represent a text tool.
- Returns the original text payload through the provider's custom-tool shape.
- Preserves newlines, indentation, tabs, trailing spaces, quotes, Unicode, empty input, and JSON-looking text.
- Streams one stable call ID through call creation, payload deltas, completion, storage, and later result correlation.
- Keeps parallel calls on one assistant turn and validates results against their call IDs.
- Rebuilds stored custom calls and results using the same canonical prompt representation as a live continuation.

## Validation

The provider adapters reject:

- unknown tool types;
- missing or empty tool names;
- duplicate names;
- a forced choice for an undeclared tool or the wrong tool kind;
- function-only fields on a custom tool;
- non-text custom formats or unsupported text-format constraints;
- `strict: true` function tools, because strict schema enforcement is not implemented;
- malformed custom calls/results and unmatched call IDs.

Unsupported constraints are rejected rather than silently ignored.

## Scope

This PR implements provider transport and lifecycle support for generic text custom tools. It does **not** implement:

- the stock GitHub Copilot `apply_patch` Lark grammar or marker-delimited raw-envelope recognition; that is #1088;
- model-specific automatic Qwen XML tool-call decoding; that is #1104;
- reordering parallel tool results for positional model templates.

Accordingly, this PR supplies the provider foundation needed by patch-style tools, but does not by itself make the
stock Copilot `apply_patch` declaration executable.


sdk_v2/cpp/src/contracts/responses_json.cc
```diff
@@ -185,52 +203,189 @@ void from_json(const nlohmann::json& j, FunctionCallInputItem& f) {
...
   f.type = j.value("type", "function_call_output");
-  f.call_id = j.at("call_id").get<std::string>();
+  f.call_id = RequiredNonEmptyString(j, "call_id", "function_call_output");
   f.output = j.at("output").get<std::string>();
 }
 
+void from_json(const nlohmann::json& j, CustomToolCallResultInputItem& c) {
+  c.type = j.value("type", "custom_tool_call_output");
+  c.call_id = RequiredNonEmptyString(j, "call_id", "custom_tool_call_output");
+  c.output = j.at("output").get<std::string>();
```
tianleiwu: [P2] Accept text-content arrays in custom tool results

The Responses contract permits `custom_tool_call_output.output` to be either a string or a list of output content ([upstream type](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_custom_tool_call_output_param.py)). Even the text-only form `[{"type":"input_text","text":"OK"}]` throws `json.exception.type_error.302` here, which the HTTP parser returns as 400, so a client using that valid representation cannot continue after a custom call. I reproduced this with the PR's actual parser; the equivalent string output succeeds.

Please accept strings and supported text-content arrays, preserving order and bytes while explicitly rejecting unsupported media parts. The stored replay branch in `AddJsonItemsToRequest` also uses `entry.value("output", "")`; it needs the same conversion because `ToInputItems` keeps the original array. Cover both direct input and replay via `previous_response_id`.

sdk_v2/cpp/src/contracts/tool_definitions.cc
```diff
@@ -0,0 +1,141 @@
...
+  });
+
+  if (forced == definitions.end()) {
+    FL_THROW(FOUNDRY_LOCAL_ERROR_INVALID_ARGUMENT,
+             "tool_choice references an undeclared tool or the wrong tool kind: " + name);
+  }
+
+  ToolDefinition only = std::move(*forced);
+  definitions.clear();
+  definitions.push_back(std::move(only));
```
tianleiwu, line 135: [P2] Validate duplicate names before narrowing the tool inventory

Both provider converters call this helper before registering the returned definitions, but this clears every declaration except the first match. `ToolRegistry::Add` therefore never sees the duplicates it is supposed to reject. With two custom `edit` declarations carrying different descriptions, I reproduced automatic choice rejecting the request while forced custom choice accepts it and silently uses the first description. A same-named function/custom pair can similarly lose its conflict through kind filtering.

Please validate name uniqueness over the full declared inventory before applying forced or allowed-tool filters, then normalize/register only once. Add provider-level duplicate tests with forced and allowed choices; testing `ToolRegistry` alone misses this bypass. This preserves the stated duplicate-name contract independently of selection mode and declaration order.

tianleiwu (changes requested):
The raw-input representation, request-local tool snapshots, and shared streaming conversion are well structured. I found two reproducible provider-contract issues that should be fixed before merging: valid text-content-list tool results are rejected, and tool selection can bypass duplicate-name validation. Details and suggested coverage are inline.

Validation: 121 existing parser/registry tests passed in an isolated binary built from the unchanged PR sources. Separate executable probes confirmed both findings, with controls for accepted string results and automatic-choice duplicate rejection. I did not run the full native SDK, model-backed HTTP, or GPU integration suites locally. Existing review threads were checked; the comments below are new concerns, not repeats of the resolved feedback.

sdk_v2/cpp/src/contracts/responses_json.cc
baijumeswani: Custom tool results now accept both: a string or an ordered array of `input_text` parts.

sdk_v2/cpp/src/contracts/tool_definitions.cc
baijumeswani: Updated so that responses api and chat completions both validate the exact case-sensitive uniqueness across the tool definictions.

bmehta001 (approved):

tianleiwu (approved):
Re-reviewed the follow-up commit 47039148 against f462a0d3. Both earlier findings are addressed: custom tool result text arrays now share the same parser for typed input and stored replay, and both provider adapters validate duplicate names before forced/allowed filtering. The added regression tests cover the relevant adapter and stored-input cases. No new actionable findings in the incremental changes.

Validation: rebuilt and ran 121 existing parser/registry tests against an isolated snapshot of the new head; all passed. The earlier executable probes now pass with the corrected expectations, including mixed-kind duplicate rejection, case-sensitive distinct names, ordered text-array results, whitespace preservation, empty arrays, and malformed-content rejection. Full SDK, model-backed HTTP/streaming, and the complete Responses converter/replay suite were not run locally.

Both earlier threads were already resolved, and I agree with those resolutions. This approval updates my previous request-changes verdict; no duplicate inline comments are being added.

PR #1088, opened 2026-09-09
Support declared raw-envelope output for custom tools

baijumeswani:
## Summary

Add bounded raw-envelope recognition for text custom tools before generated bytes are streamed as assistant text.

The primary compatibility case is GitHub Copilot's `apply_patch` tool. When a request declares the exact GitHub Copilot `apply_patch` grammar, Foundry Local recognizes a complete patch envelope as one custom tool call and preserves the complete envelope as the call input.

Other harnesses can opt a custom tool into the same mechanism by registering a small `tool_output_encoding` descriptor in request metadata.

Foundry Local recognizes and transports the call. It does not parse patch semantics, modify files, or execute tools.

## Why

A coding model may emit a tool invocation as raw text rather than in its normal structured function-call syntax.

For `apply_patch`, the generated output looks like:

```text
*** Begin Patch
*** Update File: src/math.py
@@
-return a - b
+return a + b
*** End Patch
```

If Foundry Local streams these bytes as ordinary assistant text, the harness cannot later retract them and reinterpret them as a tool call. Streaming events, final output, stored responses, transcript replay, call IDs, and tool-result continuation would disagree.

This change performs the classification before output crosses the API boundary.

## Built-in GitHub Copilot `apply_patch` support

Foundry Local recognizes the exact custom-tool grammar sent by GitHub Copilot for `apply_patch`.

The built-in envelope is:

```text
start: *** Begin Patch
end:   *** End Patch
```

The built-in behavior is enabled only when:

- the request declares a custom tool named `apply_patch`;
- its grammar exactly matches the supported GitHub Copilot grammar;
- tool output is enabled;
- the tool remains in the effective tool set after filtering.

It works with omitted or automatic tool choice because that is how stock GitHub Copilot sends the request. Ordinary model output remains ordinary text unless a valid envelope actually opens.

When `apply_patch` is explicitly forced, the declared grammar is applied as generation guidance and the output is classified directly as raw tool content rather than private reasoning.

Arbitrary Lark grammars are not accepted as built-in raw tools.

## Explicit raw-envelope registration

Other harnesses and applications can register a raw-envelope custom tool through request metadata:

```json
{
  "metadata": {
    "tool_output_encoding": "{\"type\":\"raw_envelope\",\"tool_name\":\"submit_change\",\"start_marker\":\"<<<CHANGE\",\"end_marker\":\"CHANGE>>>\"}"
  }
}
```

Metadata values are strings on both provider surfaces, so `tool_output_encoding` contains a JSON-encoded descriptor
string. The descriptor is intentionally small:

- `type` must be `raw_envelope`;
- `tool_name` must identify a declared custom tool;
- start and end markers must be distinct, non-empty, single-line strings;
- unknown fields are rejected.

No version field or caller-configurable memory limit is exposed.

Explicit descriptors can be used with automatic or matching forced tool choice. The descriptor must identify a custom
tool in the effective tool set after request filtering; it never enables an undeclared tool and cannot convert a
function tool into a custom tool. If the request declares the exact stock `apply_patch` grammar, a conflicting explicit
descriptor is rejected.

`tool_output_encoding` is reserved internal request metadata. Responses removes it from echoed, stored, GET, and list
metadata.

## Activation and arbitration

Raw recognition is request-scoped.

The raw detector is composed with the existing structured tool-call accumulator rather than replacing its behavior.

The earliest candidate owns its bytes:

- once a structured tool call opens, a later raw marker cannot supersede it;
- once a raw envelope opens, structured-looking content inside it remains raw payload;
- after the first recognized raw call, no second raw or structured call is inferred from later output.

This preserves existing structured function-tool behavior while preventing the same generated bytes from being interpreted by two parsers.

## Streaming and byte fidelity

Markers may be split across arbitrary token and streaming boundaries.

The emitted custom-tool input contains the envelope from the start marker through the closing marker.

The implementation preserves:

- LF and CRLF inside the envelope;
- tabs and indentation inside the envelope;
- trailing spaces;
- Unicode;
- quotes;
- JSON-looking text;
- exact call IDs across streaming and final output.

A line terminator after the closing marker is not part of the custom-tool input. Because an assistant turn cannot
represent visible text after a tool call without reordering it during replay, trailing whitespace after the recognized
call is discarded by the existing assistant-turn ordering guard.

Raw buffering is bounded to 1 MiB per active request. The limit is implementation-owned and cannot be increased by a client.

A valid envelope at the limit behaves identically whether it arrives in one chunk, byte by byte, or at any intermediate split.

## Safety rules

A raw call is actionable only after a complete valid envelope is recognized.

The following remain non-actionable text:

- Markdown backtick or tilde fenced examples;
- indented start or end markers;
- suffixed or misspelled markers;
- reversed markers;
- incomplete envelopes;
- envelopes exceeding the fixed buffer limit;
- candidates interrupted by reasoning;
- candidates interrupted by cancellation, stop strings, token limits, context limits, or backend failure.

A closing marker without a trailing newline is accepted only when the backend positively reports natural end of sequence. It is not accepted after a stop string, limit, cancellation, or failure.

A lone carriage return is not treated as a valid line ending.

Reasoning content never becomes raw tool input. If a raw candidate is interrupted by reasoning, the candidate is rejected while Markdown fence and visible-line state remain intact.

## Generation behavior

For an explicitly forced `apply_patch` call, the exact supported Lark grammar is used as guidance.

For an explicitly registered raw tool without a grammar:

- private reasoning remains private;
- the model may close reasoning and then emit the registered envelope;
- unrelated structured guidance is not forced onto the raw payload.

For automatic `apply_patch`:

- the model may reason normally;
- a later visible valid patch envelope becomes the custom call;
- an ordinary answer remains text.

Automatic decoding of Qwen's native XML function-call dialect is separate work and is not added here.

### Chat Completions

- Accepts the exact nested GitHub Copilot custom grammar declaration.
- Returns raw calls in the custom tool-call shape introduced by #1087.
- Preserves one stable call ID in streaming and final responses.
- Accepts the correlated tool result for continuation.

### Responses

- Accepts the exact flat GitHub Copilot custom grammar declaration.
- Emits the standard custom-tool call lifecycle and exact input bytes.
- Preserves call IDs through stored responses and continuation.
- Keeps internal raw provenance out of GET and list responses.

Wrong-surface grammar shapes and unsupported grammars return a client error.

## Replay and continuation

Generated raw calls carry one narrow internal provenance value distinguishing raw-envelope output from structured output.

The complete envelope remains the authoritative stored payload.

During replay:

- generated raw calls are rendered as the assistant envelope bytes originally produced by the model;
- structured calls remain structured calls;
- client-supplied custom calls cannot forge generated raw provenance;
- tool results remain correlated by call ID;
- warm and cold replay of server-stored Responses chains produce the same canonical history.

Raw provenance is intentionally internal and is not accepted from client-supplied history. For stateless Chat Completions
or Responses requests that echo their own input, a harness that needs the original prompt shape must replay the model's
raw envelope bytes as assistant content rather than re-submit the returned custom-call item.

Successful raw tool turns invalidate retained generation state and rebuild from canonical history on continuation, matching the existing correctness-first policy.


sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -819,11 +1066,14 @@ void ChatSession::ProcessRequestImpl(const Request& request, Response& response)
         continue;
       }
 
-      emit_tool_output(tool_accumulator.Push(seg.text));
+      emit_tool_output(push_tool_output(seg.text));
```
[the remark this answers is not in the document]
baijumeswani: Good catch. The implementation intentionally drops whitespace after a recognized tool call through `AssistantTurnGuard`, because the transcript cannot represent visible text after a call without replaying it in the wrong order. The detector therefore excludes the closing marker terminator from the custom-tool input, and the guard discards that trailing whitespace on both the stateful/Responses and Chat Completions paths. I updated the PR description to state this behavior precisely rather than claiming the terminator remains trailing assistant text. No code change is needed.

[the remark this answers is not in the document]
baijumeswani: updated PR description.

sdk_v2/cpp/src/inferencing/generative/toolcalling/raw_envelope_detector.h
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    if (count == marker.size()) {
+      return true;
+    }
+
+    const auto suffix = marker.substr(count);
+    if (fence_.has_value()) {
+      return count >= fence_->size() &&
+             suffix.find_first_not_of(" \t") == std::string_view::npos;
```
tianleiwu: [P1] Retain a possible CRLF fence closer across chunks

If a chunk ends after the `\r` in a closing fence's CRLF, this suffix check returns false. `DrainSafeOutsideText` then emits the partial line and sets `outside_line_decided_`; the following `\n` bypasses `UpdateFence`, leaving the fence open. This can make a later fenced example actionable, not just miss a legitimate call.

Reproduced with the current detector, where `patch` is a complete patch envelope ending in LF:
```cpp
const std::string prefix = "~~~\r\nexample\r\n~~~\r";
const std::string suffix = "\nprose\n~~~\n" + patch + "~~~\n";
```
`Push(prefix + suffix)` followed by natural finalization produces zero calls; `Push(prefix); Push(suffix)` followed by the same finalization produces one raw call from the fenced example. Backtick fences reproduce this too.

Please keep a possible CRLF closing line buffered until the next byte establishes its ending, without accepting a lone CR, and add every-split coverage for CRLF closers followed by another fenced example.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -265,6 +321,170 @@ void NormalizeToolOutputBatch(ToolCallStreamAccumulator::Output& output,
...
+    // so caller-visible event ordering remains identical to model output ordering.
+    auto structured_output = structured_accumulator.Flush();
+    structured_call_completed |= ContainsToolCall(structured_output);
+    AppendToolOutput(output, std::move(structured_output));
+
+    auto raw_call = std::move(std::get<ParsedToolCall>(event));
+    if (structured_call_completed) {
+      AppendToolOutput(output, structured_accumulator.Push(raw_call.arguments));
+    } else {
+      output.events.emplace_back(std::move(raw_call));
```
tianleiwu, line 405: [P2] Bypass structured-wrapper extraction for recognized raw calls

Raw calls emitted here subsequently pass through `NormalizeToolOutputBatch` in both generation paths. That helper applies `ExtractCustomToolInput(call->argument_source)` to every custom call without checking `raw_envelope`, so a valid raw payload can lose its envelope before streaming, final output, and replay.

For example, a descriptor with `type=raw_envelope`, `tool_name=edit`, `start_marker={` and `end_marker=}` is accepted for a declared custom tool. The detector correctly recognizes the 30-byte envelope `{\n"input":"replacement text"\n}`, but custom normalization reduces it to the 16-byte string `replacement text`. I reproduced this using the current descriptor validator, detector, and extraction helper. The raw provenance still survives, so replay renders the stripped text as though it were the original envelope.

Please skip `ExtractCustomToolInput` for `call->raw_envelope` while retaining UTF-8/NUL validation, and cover exact streaming/final/replay bytes for a JSON-shaped raw envelope.

tianleiwu (changes requested):
Requesting changes for two reproduced correctness issues in raw-output classification and byte preservation; details and reproductions are inline. The request-scoped activation, explicit terminal causes, and private replay provenance are useful safeguards.

Validation: all 23 existing raw-envelope detector/activation tests passed in a standalone harness against the current sources. Separate probes reproduced both findings using the current detector, descriptor validation, and custom-input extraction code. No full SDK build or model-backed/HTTP integration suite was run locally. The existing resolved trailing-whitespace thread was reviewed and is not repeated.

sdk_v2/cpp/src/inferencing/generative/toolcalling/raw_envelope_detector.h
baijumeswani: Updated `CouldStillAffectFence()` to keeps a possible closing fence buffered when the current suffix ends in `\r` .

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
baijumeswani: `NormalizeToolOutputBatch()` now applies `ExtractCustomToolInput()` only to structured custom calls

tianleiwu (commented):
Re-reviewed at `97cffc7e`. Both findings from my previous round are fixed and those threads are resolved:

- `CouldStillAffectFence()` now strips a trailing `\r` before the whitespace-only suffix check, and `CrLfFenceCloserIsIndependentOfEveryByteSplit` exercises every byte split for both fence characters.
- `NormalizeToolOutputBatch()` now guards `ExtractCustomToolInput()` with `&& !call->raw_envelope`, covered by `CustomBatchNormalizationPreservesRawEnvelopeBytes`.

### What holds up well

The detector is genuinely split-independent: `pending_` is retained only while the bytes could still be a start-marker prefix (including the CRLF variant) or could still open/close a Markdown fence, and everything else is flushed eagerly with `outside_line_decided_` so the rest of the line is never re-interpreted. Modelling rejected bytes as a distinct `RejectedCandidate` alternative is what keeps them from being re-offered to the structured parser, and `OverLimitRawCandidateCannotPromoteNestedStructuredCall` proves it. Buffer accounting cannot underflow — the `buffered_pending_size > kMaxBufferedBytes` operand short-circuits before `kMaxBufferedBytes - buffered_pending_size` is evaluated.

Composition is also right: `PushToolOutput()` bypasses the detector entirely once `structured_accumulator.InsideToolCall()`, and `RouteRawToolOutput()` flushes the structured prefix before admitting a raw call, so neither parser can observe the other's bytes and caller-visible ordering matches model output ordering.

`IsNaturalToolOutputEnd()` fails closed (`nullopt` cause is not natural), so a generator that does not report a termination cause can never promote a newline-less closing marker. Replacing the terminal `Drain()` with `DrainPending()` in the Chat Completions path is a real fix beyond this feature: the old code marked the queue finished and joined the worker *before* pushing the final streaming chunk.

And `chat_template.cc` appending the envelope to `content` is safe rather than lucky, because `AssistantTurnGuard` treats post-call whitespace as `kDropped` and non-whitespace as `kEndTurn`, so visible text can never follow the envelope in a committed turn.

### Main open question

Raw provenance only survives the Responses store. Details inline on `response_converter.cc` — the short version is that the same call renders as assistant `content` in-session but as a structured `tool_calls` entry on stateless continuation, and Chat Completions (the stated GitHub Copilot `apply_patch` surface) is stateless by definition.

### Testing

Coverage is unusually thorough — exhaustive byte-split loops over fence handling and the exact-limit closing marker, both provider surfaces exercised for streaming and non-streaming, and rejection paths asserted on `rejected_text` rather than only on the absence of a call. The one gap I would add: a test asserting what the *second* turn's prompt looks like after a recognized raw call is echoed back by a stateless client.

Remaining inline comments are nitpicks and do not block.

sdk_v2/cpp/src/inferencing/generative/openresponses/response_converter.cc
```diff
@@ -330,7 +336,7 @@ static void AddJsonItemsToRequest(Request& request, const nlohmann::json& items)
     }
 
     if (type == "custom_tool_call") {
-      request.AddOwnedItem(MakeReplayedCustomToolCall(entry));
+      request.AddOwnedItem(MakeReplayedCustomToolCall(entry, {}));
```
tianleiwu, line 339: [P2] Raw provenance is lost whenever replay does not come from the Responses store

This call site passes an empty `raw_envelope_call_ids` set, so a `custom_tool_call` arriving in the request's own `input` array always replays as `GeneratedCallEncoding::kStructured`. Only `AddHopOutputToRequest()` receives the real set, i.e. only a stored chain walked via `previous_response_id`.

That makes the same call render two different prompts:

- in-session, `BuildMessageJson()` appends the envelope to assistant `content` and omits `tool_calls`;
- on continuation, `MakeSuppliedToolCall()` -> `NormalizeToolCallArguments(kCustom, ...)` -> `BuildToolCallsJson()` renders `tool_calls[0] = {type: "function", function: {name: "apply_patch", arguments: {"input": "*** Begin Patch\n..."}}}`.

Chat Completions is stateless by definition and has no carrier for this bit at all, and `store: false` plus client-echoed input hits the same path on Responses. Since GitHub Copilot's `apply_patch` is the motivating case and it is a Chat Completions flow, the divergence applies to the primary scenario rather than an edge case.

I understand the security framing — "client-supplied custom calls cannot forge generated raw provenance" is a deliberate and correct constraint, and I am not suggesting a client-settable field. But the PR description's "warm and cold replay produce the same canonical history" only holds for stored Responses chains, and it would help to say so explicitly, or to describe how a stateless harness is expected to keep the model's own bytes in the history it replays.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -248,6 +261,49 @@ void PopulateToolDefinitions(const std::vector<ToolDefinition>& definitions, Too
...
+  const auto grammar = context.custom_lark_grammars.find(raw->tool_name);
+  if (grammar != context.custom_lark_grammars.end()) {
+    context.guidance_type = "lark_grammar";
+    context.guidance_data = grammar->second;
+    return;
+  }
+
+  context.guidance_type.clear();
+  context.guidance_data.clear();
+  context.guidance_disabled = true;
```
tianleiwu, line 301: [P3] Dropping caller-supplied guidance here is undiagnosable

When a forced raw tool has no registered grammar, this clears the caller's `guidance_type` / `guidance_data` and sets `guidance_disabled`, which makes `ResolveTurnGuidanceOptions()` return `nullopt`. The behavior is intentional and documented ("unrelated structured guidance is not forced onto the raw payload"), but from outside the process a request that explicitly asked for guidance silently gets none.

A debug-level log naming the tool and the discarded guidance type would make this supportable without changing the semantics.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_generator.cc
```diff
@@ -218,18 +285,29 @@ int OnnxChatGenerator::AppendMessages(const std::vector<TranscriptMessage>& new_
...
+    ResetTurnState();
   } catch (const std::runtime_error& e) {
     FL_THROW(FOUNDRY_LOCAL_ERROR_INTERNAL, std::string("failed to rewind generator: ") + e.what());
   }
 }
 
+void OnnxChatGenerator::ResetTurnState() {
+  current_token_.reset();
+  last_generated_token_.reset();
+  cancelled_ = false;
```
tianleiwu: [P3] Clearing `cancelled_` is either dead or incomplete

`cancelled_` is a `std::atomic<bool>` written by the cancelling thread, and `ProcessRequestImpl()` documents "Cancel is permanent for classic generators" and discards every canceled generator through the scope guard — so today this reset should be unreachable.

If it ever becomes reachable, it is incomplete: `Cancel()` also sets the `terminate_session` runtime option on the OGA generator, and `ResetTurnState()` does not undo that. `IsDone()` would then return true via `IsSessionTerminated()`, and the new `ClassifyTurnTermination()` would classify the turn as `kFailure`, which in turn makes `IsNaturalToolOutputEnd()` reject an otherwise valid envelope.

Either drop the `cancelled_` line and keep the invariant in one place, or reset the runtime option alongside it.

sdk_v2/cpp/src/inferencing/generative/toolcalling/raw_envelope_detector.h
```diff
@@ -0,0 +1,412 @@
...
+      }
+
+      return count >= fence_->size() && suffix.find_first_not_of(" \t") == std::string_view::npos;
+    }
+
+    if (count < 3) {
+      return false;
+    }
+
+    fence_ = std::string(marker.substr(0, count));
```
tianleiwu, line 359: Nit: `CouldStillAffectFence()` reads as a predicate but assigns `fence_` here, so a partial line can open a fence before the line is complete. The behavior is consistent with `UpdateFence()` — neither rejects a backtick inside a backtick fence's info string — so a line like ``` ```x` ``` opens a fence CommonMark would not, and recognition stays suppressed until a closer appears. That bias is toward *not* recognizing an envelope, so it is safe; only the name is misleading. Something like `RetainOrOpenFence()` would match what it does.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
```diff
@@ -265,6 +321,170 @@ void NormalizeToolOutputBatch(ToolCallStreamAccumulator::Output& output,
...
+    const auto part = text.substr(position, end - position);
+
+    if (structured_accumulator.InsideToolCall()) {
+      auto structured_output = structured_accumulator.Push(part);
+      const bool call_completed = ContainsToolCall(structured_output);
+      AppendToolOutput(output, std::move(structured_output));
+      if (call_completed) {
+        auto disabled = raw_detector->DisableRecognition();
+        for (auto& event : disabled.events) {
+          output.events.emplace_back(std::move(std::get<std::string>(event)));
```
tianleiwu, line 436: Nit: `std::get<std::string>(event)` is correct today because `DisableRecognition()` only emits text, but `AbortRawToolOutput()` a few lines below handles both alternatives with `get_if`. Using the same shape here keeps a future change to `DisableRecognition()` from becoming a runtime `std::bad_variant_access` instead of a compile error.

sdk_v2/cpp/src/inferencing/generative/openresponses/response_converter.cc
baijumeswani: Updated the PR description to clarify warn/cold replays apply to server-stored Responses chains.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
baijumeswani: Added a debug level log statement that mentions that the requested guidance is being discarded.

sdk_v2/cpp/src/inferencing/generative/chat/onnx_chat_generator.cc
baijumeswani: Removed the cancelled_ = false reset.

sdk_v2/cpp/src/inferencing/generative/toolcalling/raw_envelope_detector.h
baijumeswani: Renamed to `RetainOrOpenFence()`

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
tianleiwu: Thanks — the hardening landed in `RouteRawToolOutput()`'s `push_structured`, but the identical pattern in `PushToolOutput()` was missed. At `8e3e99e6` it is still `output.events.emplace_back(std::move(std::get<std::string>(event)));` on line 436, reached when a structured call completes while the accumulator is already inside a tool call.

Same reasoning as before: unreachable today, but a future change to `DisableRecognition()` turns it into a runtime `std::bad_variant_access` instead of a compile error. Worth making the two sites match.

One small note on the new branch in `push_structured`: routing a `RejectedCandidate` into `structured_accumulator.Push()` is the one thing the detector's contract says must not happen ("Rejected bytes are tagged so composition never feeds them to the structured parser"). It is dead code, so this is not a behavior concern — but since the `else` already throws on the impossible `ParsedToolCall` case, throwing on `RejectedCandidate` too would keep the invariant stated in exactly one place.

tianleiwu (approved):
Re-reviewed at `8e3e99e6`. Clearing my earlier changes-requested state — every finding I raised is now either fixed or explicitly scoped in the PR description.

| Round | Finding | Resolution |
|---|---|---|
| `a77eec6d` | CRLF fence closer lost across chunks | Fixed; exhaustive byte-split test added |
| `a77eec6d` | `ExtractCustomToolInput()` applied to raw calls | Fixed via `&& !call->raw_envelope` |
| `97cffc7e` | Raw provenance lost on stateless replay | PR description now scopes "warm and cold replay" to server-stored Responses chains and states what a stateless harness must replay instead |
| `97cffc7e` | Caller guidance silently discarded | Debug log added, asserted by test |
| `97cffc7e` | `ResetTurnState()` clearing atomic `cancelled_` | Line removed, keeping "cancel is permanent for classic generators" in one place |
| `97cffc7e` | `CouldStillAffectFence()` naming | Renamed to `RetainOrOpenFence()` |

The documentation fix on the provenance question is the right call. Foundry Local cannot accept provenance from client-supplied history without making it forgeable, so scoping the guarantee and telling stateless harnesses to replay the envelope bytes as assistant content is the correct resolution rather than a code change.

One non-blocking follow-up left inline: the unchecked-variant hardening landed in `RouteRawToolOutput()` but the identical site in `PushToolOutput()` (line 436) was missed. Fine to fold into this PR or a follow-up.

sdk_v2/cpp/src/inferencing/generative/chat/chat_session.cc
baijumeswani: I'll address this in the next PR in the stack.

PR #1093, opened 2026-09-09
Use model token IDs for guidance markers

baijumeswani:
## What this fixes

Foundry Local builds an llguidance grammar when a request requires a tool call.

Previously, Foundry inserted marker text directly into the grammar:

```lark
toolcall: <tool_call> functioncall </tool_call>
```

In llguidance, `<tool_call>` means a named tokenizer special token. This fails for models that use the same marker as
an ordinary token or ordinary text.

## What changed

Foundry now uses the boundary token IDs published by ORT GenAI:

- BOT — beginning of tool call
- EOT — end of tool call
- BOR — beginning of reasoning
- EOR — end of reasoning

When a published ID matches the configured marker, Foundry uses llguidance's exact numeric token syntax:

```lark
toolcall: <[248058]> functioncall <[248059]>
```

This works whether the tokenizer marks the token as special or ordinary.

If an ID is unavailable, invalid, or does not match an overridden marker, Foundry uses an escaped quoted string:

```lark
toolcall: "<tool_call>" functioncall "</tool_call>"
```

The quoted form allows llguidance to tokenize the marker as ordinary text, including markers represented by multiple
tokens.

Tool-call and reasoning start/end boundaries are resolved independently. The implementation contains no model names
or hardcoded token IDs.

## Why not mark every marker as special?

Changing tokenizer metadata affects more than grammar construction. ORT GenAI normally decodes with
`skip_special_tokens=true`, so changing an ordinary marker to `special: true` can remove it from decoded output.

Using numeric IDs preserves the model package's tokenizer behavior while still constraining the exact published
token.

## Required tool-call failures

If Foundry generates guidance for a required tool call and the runtime cannot apply it, the request now fails clearly
instead of continuing with unconstrained generation.

Invalid caller-provided guidance remains an invalid-argument error on both Generator and Engine backends.

## Example

Request:

```text
Use the multiply tool to calculate 17 times 23. Do not answer directly.
```

Result:

```json
{
  "finish_reason": "tool_calls",
  "tool_calls": [
    {
      "type": "function",
      "function": {
        "name": "multiply",
        "arguments": "{\"a\":17,\"b\":23}"
      }
    }
  ]
}
```

After the tool returns `391`, the next turn answers `17 × 23 = 391`.

### DIFF
--- a/sdk_v2/cpp/src/inferencing/generative/toolcalling/grammar.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/toolcalling/grammar.cc
@@ -55,7 +55,7 @@ std::string EscapeLarkLiteral(const std::string& text) {
 }
 
 std::string RenderLarkMarker(const std::string& marker_text, std::optional<int32_t> token_id) {
-  if (token_id.has_value()) {
+  if (token_id.has_value() && *token_id >= 0) {
     return "<[" + std::to_string(*token_id) + "]>";
   }
 
--- a/sdk_v2/cpp/src/inferencing/generative/toolcalling/grammar.cc
+++ b/sdk_v2/cpp/src/inferencing/generative/toolcalling/grammar.cc
@@ -9,6 +9,59 @@
 
 namespace fl {
 
+std::string EscapeLarkLiteral(const std::string& text) {
+  std::string out;
+  out.reserve(text.size() + 2);
+  out.push_back('"');
+
+  for (const char c : text) {
+    switch (c) {
+      case '\\':
+        out += "\\\\";
+        break;
+      case '"':
+        out += "\\\"";
+        break;
+      case '\n':
+        out += "\\n";
+        break;
+      case '\r':
+        out += "\\r";
+        break;
+      case '\t':
+        out += "\\t";
+        break;
+      case '\b':
+        out += "\\b";
+        break;
+      case '\f':
+        out += "\\f";
+        break;
+      default:
+        if (const auto byte = static_cast<unsigned char>(c); byte < 0x20) {
+          static constexpr char kHex[] = "0123456789abcdef";
+          out += "\\u00";
+          out.push_back(kHex[byte >> 4]);
+          out.push_back(kHex[byte & 0x0f]);
+        } else {
+          out.push_back(c);
+        }
+        break;
+    }
+  }
+
+  out.push_back('"');
+  return out;
+}
+
+std::string RenderLarkMarker(const std::string& marker_text, std::optional<int32_t> token_id) {
+  if (token_id.has_value()) {
+    return "<[" + std::to_string(*token_id) + "]>";
+  }
+
+  return EscapeLarkLiteral(marker_text);
+}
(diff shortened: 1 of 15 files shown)


sdk_v2/cpp/src/inferencing/generative/toolcalling/grammar.cc
```diff
@@ -9,6 +9,59 @@
...
+        }
+        break;
+    }
+  }
+
+  out.push_back('"');
+  return out;
+}
+
+std::string RenderLarkMarker(const std::string& marker_text, std::optional<int32_t> token_id) {
```
kunal-vaishnavi, line 57: From Copilot:

Could we reject negative token IDs here before using numeric-token syntax? Since any engaged `optional<int32_t>` is currently treated as valid, `RenderLarkMarker(marker, -1)` produces `<[-1]>` instead of the escaped literal fallback. Although `ChatSession::ResolveMarkerTokenId()` filters negatives, this function and `ToolCallContext` can be used directly.

Please require `token_id.has_value() && *token_id >= 0` and add a regression test confirming that `-1` renders the escaped marker.

sdk_v2/cpp/test/internal_api/toolcalling/grammar_test.cc
```diff
@@ -118,12 +118,14 @@ TEST(BuildLarkGrammarTest, ToolOnlyWithMarkersGrammar) {
...
   ctx.tool_call_end = "</tool>";
+  ctx.tool_call_start_token_id = 200;
+  ctx.tool_call_end_token_id = 201;
 
   std::string grammar = BuildLarkGrammar(ctx, R"({"type":"array"})");
   EXPECT_NE(grammar.find("start: toolcall"), std::string::npos);
   EXPECT_NE(grammar.find("functioncall"), std::string::npos);
-  EXPECT_NE(grammar.find("<tool>"), std::string::npos);
-  EXPECT_NE(grammar.find("</tool>"), std::string::npos);
+  EXPECT_NE(grammar.find("<[200]>"), std::string::npos);
```
kunal-vaishnavi, line 127: Can we ensure that our unit tests cover Phi-4 mini reasoning's case as well as the regular cases (e.g. Qwen)?

sdk_v2/cpp/src/inferencing/generative/toolcalling/grammar.cc
baijumeswani: Added the check:

```
 if (token_id.has_value() && *token_id >= 0)
```

sdk_v2/cpp/test/internal_api/toolcalling/grammar_test.cc
baijumeswani: added a test

kunal-vaishnavi (dismissed):

kunal-vaishnavi (dismissed):

kunal-vaishnavi (approved):

PR #1107, opened 2026-09-14
Improve BYOM registration metadata defaults and ownership

selenayang888:
## Summary

Improve BYOM registration metadata handling by applying explicit ownership rules:

- Require application-owned metadata such as `task`.
- Derive good defaults where authoritative information is available.
- Preserve supported application overrides.
- Prevent applications from overriding SDK-owned metadata.

Metadata ownership reference:
[BYOM Metadata Ownership](https://microsoft-my.sharepoint.com/:w:/p/yangselena/cQpxpx9suJQuRqH3BqbJWDuvEgUCVVm3lgWgfbbjU__1gjEwvQ)

## Changes

- Derive `context_length` from `genai_config.json`.
- Default `display_name` to the model name and `publisher` to `local`.
- Derive execution provider and device type when provider configuration is recognized.
- Default input and output modalities from the required task.
- Preserve application-supplied values for overridable metadata.
- Ignore conflicting caller values for SDK-owned identity, type, timestamp, path, and internal metadata, and populate those fields from authoritative SDK sources.
- Fix canonical WebGPU and Ryzen AI provider-name round trips.
- Persist the fully resolved metadata in the local model registration index.


sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -316,22 +401,50 @@ void LocalModelCatalog::UnregisterModel(const std::string& alias_or_model_id) {
...
   resolved.uri.clear();
+  if (!resolved.GetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_DISPLAY_NAME_STR)) {
+    resolved.SetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_DISPLAY_NAME_STR, name);
+  }
   if (!resolved.GetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_PUBLISHER_STR)) {
     resolved.SetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_PUBLISHER_STR, "local");
   }
+  if (!resolved.GetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_EP_STR)) {
+    if (const auto provider = ConfigExecutionProvider(genai_config)) {
+      resolved.SetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_EP_STR, *provider);
```
baijumeswani: This derived provider is later treated as an explicit load override. `Model::Load` converts any stored provider to an override, and `GenAIModelInstance` then clears the providers and options from `genai_config.json`. That can remove required CUDA, QNN, or other provider settings. Please keep the artifact-declared provider separate from a caller override, and leave effective provider selection in `ModelLoadManager`. A load test with non-empty provider options would cover the regression.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -316,22 +401,50 @@ void LocalModelCatalog::UnregisterModel(const std::string& alias_or_model_id) {
 }
 
 ModelInfo LocalModelCatalog::ResolveMetadata(const ModelInfo& metadata, const std::string& model_id,
-                                             const std::string& name, int version) const {
+                                             const std::string& name, int version,
+                                             const GenAIConfig& genai_config) const {
   auto resolved = metadata;
   RemoveLegacyRegistrationProperties(resolved);
+  RemoveSdkOwnedProperties(resolved);
```
baijumeswani: These ownership rules apply only to new registrations. Existing schema-v1 and pre-change schema-v2 entries are loaded without this normalization, and a later save can rewrite them as schema v2 while keeping the old values. Please add an explicit schema migration, or preserve the previous persisted semantics. A migration test should include conflicting type, context length, timestamps, prompt templates, model settings, and internal fields.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -316,22 +401,50 @@ void LocalModelCatalog::UnregisterModel(const std::string& alias_or_model_id) {
...
   resolved.version = version;
   resolved.model_id = model_id;
   resolved.uri.clear();
+  if (!resolved.GetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_DISPLAY_NAME_STR)) {
+    resolved.SetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_DISPLAY_NAME_STR, name);
+  }
   if (!resolved.GetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_PUBLISHER_STR)) {
     resolved.SetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_PUBLISHER_STR, "local");
   }
+  if (!resolved.GetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_EP_STR)) {
```
baijumeswani: A caller-provided execution provider is preserved without validation or canonicalization. Registration can succeed, persist the value, and then every normal `Load()` can fail because the provider is unknown. Please validate supported providers during registration, store the canonical registration name, and reject or normalize provider/device pairs that disagree. DML should remain unsupported.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -50,6 +51,70 @@ struct ParsedModelId {
...
 };
 
+std::optional<std::string> ConfigExecutionProvider(const GenAIConfig& config) {
+  const auto config_provider = config.DefaultProvider();
+  if (config_provider.empty()) {
+    return std::nullopt;
+  }
+
+  auto provider = EPUtils::StringtoEP(config_provider);
+  if (provider == ExecutionProvider::kUnknown) {
```
baijumeswani: DML is unsupported in Foundry Local, but a DML config reaches this branch as an unknown provider and registration still succeeds. The load path also has a stale DML exemption that skips provider checks. Please reject known DML spellings here and remove the DML exemption/allowlist expectations in the load path so users get a clear error before OGA model construction.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -316,22 +401,50 @@ void LocalModelCatalog::UnregisterModel(const std::string& alias_or_model_id) {
...
+      }
+    }
+  }
+  ApplyTaskDefaults(resolved);
   resolved.SetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_MODEL_PROVIDER_STR, "LocalRegistration");
   resolved.SetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_ENTITY_TYPE_STR, "Model");
   resolved.SetPropertyStr(FOUNDRY_LOCAL_MODEL_PROP_MODEL_TYPE_STR, "ONNX");
 
+  if (genai_config.model && genai_config.model->context_length > 0) {
+    resolved.SetPropertyInt(FOUNDRY_LOCAL_MODEL_PROP_CONTEXT_LENGTH_INT, genai_config.model->context_length);
```
baijumeswani: `GenAIConfig` currently accepts any JSON number for `context_length` and converts it to `int`. Fractional or out-of-range values can be truncated, omitted, or fail inconsistently before being published here as authoritative metadata. Please require a positive integral value in range, and add tests for fractional, zero, negative, and values above `INT_MAX`.

sdk_v2/cpp/include/foundry_local/foundry_local_c.h
```diff
@@ -1036,7 +1036,12 @@ struct flCatalogApi {
   /// Register a model in a local catalog without taking ownership of its assets.
   /// `model_path` must identify a model directory containing genai_config.json.
   /// `model_id` must use the canonical `<name>:<version>` format and be unique in the local catalog.
-  /// The metadata is copied; model identity and location are taken only from the explicit arguments.
+  /// `metadata.task` is required. Application-owned properties, including display name, publisher, runtime,
+  /// modalities, and arbitrary string or integer properties, are preserved; missing values receive authoritative
+  /// defaults when available.
+  /// Identity, alias, type, timestamps, context length, and prompt templates are derived by the SDK from the
+  /// arguments, registration time, and genai_config.json. Caller-supplied location and internal catalog metadata are
+  /// ignored.
```
baijumeswani: The C/C++ contract is updated, but the C#, JavaScript, Python, and Rust registration docs still say metadata is copied or do not explain ownership. Please carry this contract through every public SDK: required fields, preserved overrides, derived defaults, ignored SDK-owned fields, and whether runtime values affect loading.

sdk_v2/cpp/src/catalog/local_model_catalog.cc
```diff
@@ -50,6 +51,70 @@ struct ParsedModelId {
...
+      return "CPU";
+    case ExecutionProvider::kCUDA:
+    case ExecutionProvider::kWebGPU:
+    case ExecutionProvider::kTensorRT_RTX:
+      return "GPU";
+    case ExecutionProvider::kVitisAI:
+    case ExecutionProvider::kRyzenAI:
+    case ExecutionProvider::kQNN:
+      return "NPU";
+    default:
```
baijumeswani: OpenVINO is recognized as a provider but cannot be assigned one device class without inspecting its options, so leaving the device unknown is reasonable. However, the current serializer can persist this as `deviceType: "Invalid"`, and the C++ `GetRuntime()` then hides the provider because the device is unset. Please represent an unknown device by omitting it rather than serializing `Invalid`, and allow runtime metadata to expose a provider even when its device class is unknown.

baijumeswani (commented):
Thanks for improving BYOM metadata handling. Centralizing this in the native catalog is the right direction. I found one load regression and several contract and migration gaps. I left inline recommendations for resolving each concern.

PR #1114, opened 2026-09-17
catalog v2 migration + minFLVersion check

prathikr:
- migrates catalog api from `crossRegion/` to `asset-gallery/` based catalog fetch
- introduces `minFLVersion` check based on new catalog api response


sdk_v2/cpp/src/catalog/azure_catalog_models.cc
```diff
@@ -239,52 +216,38 @@ std::optional<ModelInfo> CatalogModelToModelInfo(const CatalogLocalModel& cm) {
...
 
-  if (!cm.properties || !cm.properties->name || cm.properties->name->empty()) {
+  if (!cm.name || cm.name->empty()) {
     return std::nullopt;
   }
 
-  if (!cm.entity_id || cm.entity_id->empty()) {
+  if (!cm.alias || cm.alias->empty()) {
     return std::nullopt;
   }
```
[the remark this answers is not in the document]
prathikr: Removed fixture

sdk_v2/cpp/CMakeLists.txt
```diff
@@ -382,12 +382,10 @@ endfunction()
...
 # --------------------------------------------------------------------------
-# CI overrides via -DFOUNDRY_LOCAL_VERSION_STRING=<full-version-string> so the
-# binary's FoundryLocalGetVersionString() matches the .nupkg version (which
-# typically includes a dev/timestamp suffix). Defaults to PROJECT_VERSION for
-# local developer builds.
+# Keep the runtime catalog-compatibility version aligned with the current SDK
+# version, regardless of a stale CMake cache or CI package-version override.
 set(FOUNDRY_LOCAL_VERSION_STRING "${PROJECT_VERSION}"
-    CACHE STRING "Version string returned by FoundryLocalGetVersionString()")
+    CACHE STRING "Version string returned by FoundryLocalGetVersionString()" FORCE)
```
baijumeswani: `FORCE` overwrites the version passed by the packaging pipeline. This can make a prerelease package report itself as `2.0.1`, and the same wrong value is then used for `minFLVersion` checks. Please remove `FORCE` so a caller-provided version wins. If `2.0.1` is the intended package version, update the package and pipeline version sources together instead. A small configure test with `-DFOUNDRY_LOCAL_VERSION_STRING=<prerelease>` would prevent this from coming back.

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -74,163 +70,125 @@ std::vector<std::string> CreateModelFilter(const std::string& filter_override) {
...
+    return std::nullopt;
   }
-
-  logger.Log(LogLevel::Information, "Detected catalog region: '" + region + "'.");
-  return region;
-}
-
-std::string BuildRegionalUrl(const std::string& url_prefix, const std::string& url_suffix, const std::string& region) {
-  return url_prefix + region + url_suffix + kEntitiesPath;
+  return SemVer{major, minor, patch};
```
baijumeswani: This accepts only `major.minor.patch`, but our packages can use versions such as `2.0.1-dev.123` or `2.0.1-rc.1`. After the CMake override is fixed, those builds would fail to parse their own version and hide every model with a `minFLVersion`. Please use a SemVer parser that handles prerelease and build suffixes, and add tests using the actual pipeline version format.

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -240,152 +198,67 @@ AzureCatalogClient::AzureCatalogClient(const std::string& base_url,
...
-    } else if (regional) {
-      // Subsequent pages are pinned to the region that served page 1 — a filter set
-      // never mixes regions (continuation tokens are region-specific).
-      response = http_post_response_(BuildRegionalUrl(url_prefix_, url_suffix_, *pinned_region), body);
-    } else {
-      // Non-regional / custom URL: single verbatim attempt, no fallback.
-      response = http_post_response_(BuildRequestUrl(base_url_, regional, region_, url_prefix_, url_suffix_), body);
-    }
+    const std::string body = BuildRequestBody(filters, continuation_token);
+    http::HttpResponse response = http_post_response_(base_url_, body);
```
baijumeswani: The old client kept the serving region from the response and had recovery for temporary regional failures. This call drops both behaviors. The download manager still uses `ModelInfo::detected_region`, so catalog discovery can succeed but download starts in an unrelated default region. Please read `azureml-served-by-cluster` and carry that region into each model. Also decide whether 429/5xx/transport failures should retry; if not, document the behavior change and add a test that makes it explicit.

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -240,152 +198,67 @@ AzureCatalogClient::AzureCatalogClient(const std::string& base_url,
...
+      logger_.Log(LogLevel::Warning,
+                 "catalog request to " + base_url_ + " failed: " + http::DescribeFailure(response));
       FL_THROW(FOUNDRY_LOCAL_ERROR_NETWORK,
-               "catalog request to " + url + " failed: " + http::DescribeFailure(response));
+               "catalog request to " + base_url_ + " failed: " + http::DescribeFailure(response));
     }
 
     const auto parsed = nlohmann::json::parse(response.body).get<AzureCatalogResponse>();
-    if (!parsed.index_entities_response) {
+    if (parsed.models.empty()) {
```
baijumeswani: This stops when a page has no models, even if it has a continuation token. The reverse case is also unsafe: if the service repeats a token with a non-empty page, this loop never ends and keeps adding duplicates. Please follow pagination based on the token, not page size, and keep a set of seen tokens so repeated or cyclic tokens fail cleanly. Add tests for an empty page with a next token and for a repeated token.

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -394,36 +267,36 @@ std::vector<ModelInfo> AzureCatalogClient::FetchModelsByIds(
...
+  std::erase_if(model_infos, [&model_ids](const ModelInfo& info) {
+    return std::find(model_ids.begin(), model_ids.end(), info.model_id) == model_ids.end();
+  });
+  return model_infos;
 }
 
 std::vector<ModelInfo> AzureCatalogClient::FetchAllVersionsByAlias(
     const std::string& model_alias,
     const std::string& model_name,
     int /*max_versions*/) {
```
baijumeswani: The method contract says `max_versions` is applied per variant, but this implementation ignores it. It also combines results from overlapping device/provider searches without sorting or removing duplicates. The public catalog currently applies another limit later, but direct client behavior no longer matches its contract. Please sort by version, remove duplicate model IDs, and apply the requested limit per variant here, or change the interface contract if the limit is intentionally owned by the caller.

sdk_v2/cpp/src/catalog/azure_catalog_models.cc
```diff
@@ -141,96 +171,100 @@ void from_json(const nlohmann::json& j, VariantInfo& v) {
...
+    if (!m.supports_reasoning && tags.contains("supportsReasoning") &&
+      tags["supportsReasoning"].is_string()) {
+      const auto value = ToLower(tags["supportsReasoning"].get<std::string>());
+      if (value == "true") {
+        m.supports_reasoning = true;
+      } else if (value == "false") {
+        m.supports_reasoning = false;
+      }
+    }
+    return;
```
baijumeswani, line 227: The full-service response takes this early return before `modelLimits` is read. It also does not copy input/output modalities or the `foundryLocal == "test"` marker. As a result, public metadata such as `MaxOutputTokens()`, modalities, and `IsTestModel()` is lost for the response shape requested by this client. Please map these values from `annotations.systemCatalogData` and `annotations.tags` before returning, then assert the public `ModelInfo` accessors in a full-service fixture.

sdk_v2/cpp/src/catalog/azure_catalog_models.cc
```diff
@@ -141,96 +171,100 @@ void from_json(const nlohmann::json& j, VariantInfo& v) {
...
   }
 }
 
-void from_json(const nlohmann::json& j, IndexEntitiesResponse& r) {
+void from_json(const nlohmann::json& j, AzureCatalogResponse& r) {
   opt_int(j, "totalCount", r.total_count);
-  opt_int(j, "nextSkip", r.next_skip);
   opt_str(j, "continuationToken", r.continuation_token);
 
   if (j.contains("value") && j["value"].is_array()) {
```
baijumeswani, line 273: A 200 response with the wrong shape, such as `{}`, `[]`, an error object, or a string-valued `summaries`, is treated as a valid empty catalog. That marks the live request as successful and prevents fallback to the cached snapshot. Please require a valid response object with a recognized array field. A present empty array can be valid, but a missing or wrong-typed field should raise a stable SDK error so snapshot fallback runs.

sdk_v2/cpp/src/catalog/azure_catalog_models.cc
```diff
@@ -239,52 +273,38 @@ std::optional<ModelInfo> CatalogModelToModelInfo(const CatalogLocalModel& cm) {
...
-  if (!cm.properties->variant_info) {
+  // Entries with no variant information are the abstract parent model, not a
+  // runnable variant — skip them.
+  if (!cm.variant_information) {
     return std::nullopt;
   }
 
-  const auto& props = *cm.properties;
-  int version = static_cast<int>(props.version.value_or(0));
+  const int version = ParseIntString(cm.version).value_or(0);
```
baijumeswani: Missing, malformed, overflowing, or partly numeric versions all become version `0` here; negative values are accepted too. That can create fake IDs such as `model:0` and collapse different bad records during deduplication. Please require a valid supported-range version before building `ModelInfo` and skip or reject invalid records. Add cases for missing, negative, non-numeric, trailing text, and overflow values.

sdk_v2/cpp/test/sdk_api/catalog_live_test.cc
```diff
@@ -30,11 +30,7 @@ namespace {
 
 namespace fs = std::filesystem;
 
-// The live catalog endpoint and an explicit region. Setting the region exercises the public
-// Configuration::SetCatalogRegion() override path; it matches the URL template so routing stays
-// consistent.
-constexpr const char* kLiveCatalogUrl = "https://ai.azure.com/api/centralus/ux/v1.0";
-constexpr const char* kLiveCatalogRegion = "centralus";
+constexpr const char* kLiveCatalogUrl = "https://api.catalog.azureml.ms/asset-gallery/v1.0/models";
```
baijumeswani, line 33: This live test uses the new endpoint, but two other test paths still use the old catalog URL: `test/sdk_api/shared_test_env.h:281` and the disabled test in `test/internal_api/download_test.cc:1338`. The shared environment means the integration suite still depends on `/ux/v1.0`, even though the production default has moved. Please update both remaining URLs to this Asset Gallery endpoint and run the targeted integration tests. That will prove the repository no longer relies on the old service.

baijumeswani (commented):
Thanks for moving the default catalog to Asset Gallery and adding `minFLVersion` filtering. The main direction is right, but I found several behavior and reliability gaps that should be addressed. The inline comments explain each concern and a suggested fix. One important scope note: the production default no longer calls the old endpoint, but the test suite still contains two old `/ux/v1.0` URLs, so the repository is not yet fully independent of the old catalog service.

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -74,318 +73,217 @@ std::vector<std::string> CreateModelFilter(const std::string& filter_override) {
...
+  filters.push_back(MakeFilter("name", names));
   return filters;
 }
 
 }  // namespace
 
+bool IsFoundryLocalVersionCompatible(const std::string& current_version,
+                                     const std::string& minimum_version) {
+  try {
+    return semver::version::parse(current_version) >= semver::version::parse(minimum_version);
```
baijumeswani, line 191: This uses package SemVer ordering as the catalog compatibility rule. CI builds packages such as `2.0.0-dev.<timestamp>`, which SemVer correctly considers lower than `2.0.0`. As a result, models requiring `2.0.0` are hidden from the prerelease package. This is why all five native jobs now fail `ModelFixture.CatalogValidation` with only 4 models. Please keep package version ordering and catalog compatibility separate. A good option is a dedicated catalog compatibility version that describes the supported model contract. Another option is to align the pipeline package version with the compatibility level. Add a test using the exact pipeline version format and a `2.0.0` catalog minimum.

sdk_v2/cpp/src/catalog/azure_catalog_models.cc
```diff
@@ -141,96 +173,115 @@ void from_json(const nlohmann::json& j, VariantInfo& v) {
...
+        m.version = std::to_string(properties["version"].get<int>());
+      }
+    }
+    if (system_data.contains("inferenceTasks") && system_data["inferenceTasks"].is_array()) {
+      m.inference_tasks = system_data["inferenceTasks"].get<std::vector<std::string>>();
+    }
+    if (system_data.contains("modelCapabilities") && system_data["modelCapabilities"].is_array()) {
+      m.model_capabilities = system_data["modelCapabilities"].get<std::vector<std::string>>();
+    }
+    if (system_data.contains("modelLimits") && system_data["modelLimits"].is_object()) {
```
baijumeswani, line 208: The live full-service response does not put these values inside `systemCatalogData.modelLimits`. It returns `maxOutputTokens`, `textContextWindow`, `inputModalities`, and `outputModalities` directly under `systemCatalogData`. The current code still drops those public properties. Please map the direct fields into `ModelLimits`. Keeping support for the nested shape is fine, but the main test fixture should use a reduced copy of the real service response so it protects the actual contract.

sdk_v2/cpp/src/catalog/azure_catalog_models.cc
```diff
@@ -239,52 +290,42 @@ std::optional<ModelInfo> CatalogModelToModelInfo(const CatalogLocalModel& cm) {
     return std::nullopt;
   }
 
-  if (!cm.properties || !cm.properties->name || cm.properties->name->empty()) {
+  if (!cm.name || cm.name->empty()) {
     return std::nullopt;
   }
 
-  if (!cm.entity_id || cm.entity_id->empty()) {
+  if (!cm.alias || cm.alias->empty()) {
```
baijumeswani, line 297: This drops an otherwise valid runnable variant when the service omits the redundant alias field. Main previously derived the alias from the parent model URI and then fell back to the variant name. Please preserve that behavior: use the explicit alias when present, otherwise use the parent model name, and finally use `cm.name`. Add a test with a valid parent and no explicit alias.

sdk_v2/cpp/src/catalog/azure_catalog_client.cc
```diff
@@ -74,318 +73,217 @@ std::vector<std::string> CreateModelFilter(const std::string& filter_override) {
...
-    const std::vector<CatalogFilter>& filters) {
-  const bool regional = UsesRegionalRouting(regional_template_, region_);
-
-  FetchedFilterSet result;
-  result.region = region_;
+http::HttpResponse AzureCatalogClient::PostWithRetry(const std::string& body) {
+  std::optional<http::HttpResponse> successful_response;
+  http::RetryWithBackoff(
+      [&]() -> http::RetryAttempt {
+        auto response = http_post_response_(base_url_, body);
```
baijumeswani, line 223: The retry path handles returned status `0`, but an exception from the HTTP callback escapes before `RetryWithBackoff` can retry it. In the production HTTP path, sending the request is caught, but reading the response body can still throw outside that catch. Please convert expected transport/read exceptions into a transient retry result, preferably in the shared HTTP helper. Add one test where the callback throws once and then succeeds, and another where it keeps throwing until the retry budget is exhausted.

baijumeswani (commented):
Thanks for addressing most of the earlier review comments. The endpoint migration, pagination checks, response validation, region handling, retries, version parsing, and remaining old test URLs are much improved. I still see four concerns. The first two affect the current catalog behavior and should be resolved before merge. The native builds compile, but all five platform jobs fail the same catalog integration test because only 4 models are returned instead of more than 10.

baijumeswani (approved):

PR #1119, opened 2026-09-18
Restore model download cancellation parity

bmehta001:
## Summary
- restore JavaScript model-download cancellation through `AbortSignal`, including progress-callback bridging and late-abort-safe `AbortError` mapping
- restore Python model-download cancellation through `threading.Event` while preserving user progress-callback exceptions
- preserve the existing JavaScript callback-first API while adding AbortSignal as an optional second argument

## Validation
- Windows and WSL/Linux: Node addon build and TypeScript build/type checks
- Windows and WSL/Linux: focused JavaScript model and lifecycle tests (11 passed; 6 real-model tests skipped without `FOUNDRY_TEST_DATA_DIR`)
- Windows and WSL/Linux: Python download-cancellation tests (3 passed)
- Biome lint for changed JavaScript/TypeScript/native-addon files
- Ruff for changed Python source and tests
- `git diff --check`

The full Biome formatter check on the Windows checkout reports pre-existing CRLF normalization differences; focused lint passes without suppressions.

### Real-download cancellation stress
Each API used its own empty cache and the real `qwen2.5-0.5b` registry download. The matrix included two pre-start cancellations and eight active cancellations at thresholds from 0% through 3%, verifying progress occurred and the model remained uncached after every cancellation.

| API | Windows | WSL/Linux |
|---|---:|---:|
| JavaScript `AbortSignal` | 10/10 canceled; recovery download reached 100% | 10/10 canceled |
| Python `threading.Event` | 10/10 canceled; recovery download reached 100% | 10/10 canceled |

The successful recovery downloads reused the same partially downloaded cache after all cancellation attempts, validating cleanup/resume behavior. WSL cancellation used separate empty caches; full recovery was not repeated there to avoid two additional full model downloads.


sdk_v2/js/native/src/model.cc
```diff
@@ -276,34 +279,91 @@ Napi::Value Model::Unload(const Napi::CallbackInfo& info) {
...
+        }
+        if (tsfn_) {
+          ProgressDispatch dispatch(percent);
+          const napi_status status = tsfn_.BlockingCall(
+              &dispatch, [](Napi::Env env, Napi::Function js_cb, ProgressDispatch* pending) {
+                ProgressDispatchCompletion completion(*pending);
+                if (env == nullptr || js_cb.IsEmpty()) {
+                  return;
+                }
+                js_cb.Call({Napi::Number::New(env, static_cast<double>(pending->Percent()))});
```
baijumeswani: If the user's progress callback throws, that error is not used to reject the download promise, and the download may continue. Please capture the first callback error, return a cancellation result to native code, and reject `download()` with that same error. Add a test where the progress callback throws. This would also match the behavior added on the Python side.

sdk_v2/js/src/model.ts
```diff
@@ -157,8 +172,18 @@ export class Model implements IModel {
     await this.#native.unload();
   }
 
-  async download(progressCallback?: (progress: number) => void): Promise<void> {
-    await this.#native.download(progressCallback);
+  async download(
+    progressCallback?: (progress: number) => void,
+    signal?: AbortSignal,
```
baijumeswani: The legacy JS API supports `model.download(controller.signal)`, but this signature requires `model.download(undefined, controller.signal)`. That makes migration harder and does not provide full API parity. Please consider a signal-only overload while keeping the callback-first forms. If that compatibility tradeoff is intentional, document the required call shape clearly and avoid describing it as full API parity.

sdk_v2/js/native/src/model.cc
```diff
@@ -378,8 +458,40 @@ Napi::Value Model::Download(const Napi::CallbackInfo& info) {
     return env.Undefined();
   }
 
+  std::shared_ptr<std::atomic<bool>> abort_requested;
+  Napi::ObjectReference abort_signal;
+  Napi::FunctionReference abort_listener;
+  if (info.Length() >= 2 && !info[1].IsUndefined() && !info[1].IsNull()) {
```
baijumeswani, line 603: The progress `ThreadSafeFunction` is created before signal setup finishes. If reading the signal or calling `addEventListener` throws, the function returns before the TSFN is handed to the worker and released. Please validate and register the signal before creating the TSFN, or use a small scope guard that always releases it on early exit. A test with a signal-like object whose `addEventListener` throws would cover this path.

sdk_v2/js/native/src/model.cc
```diff
@@ -276,34 +279,91 @@ Napi::Value Model::Unload(const Napi::CallbackInfo& info) {
...
-      auto progress_cb = tsfn_ ? std::function<int(float)>([this](float percent) {
-        // BlockingCall keeps backpressure on the worker thread: if JS is
-        // slow to drain the queue we'll wait rather than dropping reports.
-        // Callback return value is unused on the JS side; we always continue.
-        tsfn_.BlockingCall([percent](Napi::Env env, Napi::Function js_cb) {
-          js_cb.Call({Napi::Number::New(env, static_cast<double>(percent))});
-        });
-        return 0;  // 0 = continue per flProgressCallback contract.
+      const bool has_cancellation = abort_requested_ != nullptr;
+      auto progress_cb = (tsfn_ || has_cancellation) ? std::function<int(float)>([this](float percent) {
```
baijumeswani, line 462: This checks for abort only when native code reports progress. It cannot stop a request while registry resolution, blob listing, or blob property lookup is blocked; cancellation takes effect at the next progress callback. Please document that cancellation is cooperative, and add a timer-driven mid-download test instead of testing only pre-abort or abort from the 0% callback. If immediate cancellation is required, the native download APIs will need a separate cancellation token.

baijumeswani (commented):
The JavaScript cancellation path is real and useful. I left focused comments on callback errors, API compatibility, cleanup, and the limits of cooperative cancellation.

sdk_v2/js/native/src/model.cc
bmehta001: Addressed in d4fd0c68. DownloadWorker now retains the exact JavaScript exception, returns cancellation to the native download callback, and rejects download() with that same value. The new model test asserts identity with rejects.toBe(callbackError).

sdk_v2/js/src/model.ts
bmehta001: Addressed in d4fd0c68. IModel and Model now provide the signal-only overload, so model.download(controller.signal) works alongside the existing callback-first forms. The type test covers both call shapes.

sdk_v2/js/native/src/model.cc
bmehta001: Addressed in d4fd0c68. Signal validation and addEventListener registration now complete before the ThreadSafeFunction is created. A regression test uses a signal-like object whose addEventListener throws and verifies the original error propagates.

sdk_v2/js/native/src/model.cc
bmehta001: Addressed in d4fd0c68. The public interface now documents that cancellation is cooperative and may wait for registry/blob operations to reach a progress checkpoint. The native-progress test now aborts from a timer and uses the new signal-only overload rather than aborting inside the 0% callback.

baijumeswani (approved):

PR #1120, opened 2026-09-19
feat(java): add streaming ASR SDK

jiec-msft:
Adds a Java 17 SDK for Foundry Local streaming ASR through JNA.

- **Streaming API** - Supports reusable model and session ownership, PCM and WAV input, cancellation, final results, timing data, and deterministic cleanup.
- **Packaging** - Produces standard Maven binary and sources JARs while keeping native runtimes and model weights external.
- **Validation** - Adds focused lifecycle, native-contract, audio-validation, and opt-in real-model integration tests.
- **CI** - Adds a five-platform Azure matrix, resolves Java 17 for each target architecture, and bootstraps pinned Maven 3.9.9 on self-hosted agents.

The existing preview release remains unchanged.

Earlier PRs: [#1094](https://github.com/microsoft/foundry-local/pull/1094) and [#1100](https://github.com/microsoft/foundry-local/pull/1100).

Supersedes #1113 to run from a branch in this repository.


baijumeswani (commented):
Thanks for the thorough work on this Java SDK. The current implementation is careful about native ownership, cleanup, cancellation, and cross-platform testing. I reviewed it against the current C API, older released runtimes, existing SDK conventions, and the packaging pipeline.

The main issue to fix before merging is the API version mismatch: the Java code requests API v1, but it uses speech and queue function-table positions that are not compatible with an older released v1 runtime. That can call the wrong native function and crash the process. The safest fix is to require API v2 and fail cleanly on older runtimes.

The other inline comments focus on keeping native CI mandatory, testing the packaged Maven artifact, preserving useful diagnostics, avoiding JVM shutdown hangs, and aligning the public API with the other SDKs. The five-platform native jobs are a strong foundation.

sdk_v2/java/src/main/java/com/microsoft/foundry/local/NativeApi.java
```diff
@@ -0,0 +1,333 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * Narrow binding to the stable API v1 prefix of the current Foundry Local C ABI.
+ * All supported targets use 64-bit pointers/size_t. C bool is one byte, not JNA boolean.
+ */
+final class NativeApi {
+    static final int VERSION = 1;
```
baijumeswani, line 25: This should request API version 2, not version 1. An older released v1 runtime (for example `cli-preview-0.10.1`) has different `flItemApi` positions: slot 23 is `ItemQueue_Release`, while this binding calls slot 23 as `GetQueue`. Because `FoundryLocalGetApi(1)` succeeds, Java can call the wrong native function and corrupt memory. Please set `VERSION = 2`, document the v2 runtime requirement, and add a test that verifies an older v1-only runtime is rejected cleanly.

.pipelines/foundry-local-packaging.yml
```diff
@@ -69,6 +69,11 @@ extends:
       binskim:
         break: false
         scanOutputDirectoryOnly: true
+      spotBugs:
+        enabled: false
```
baijumeswani, line 73: Disabling SpotBugs leaves the new Java/JNA code without Java static analysis. Could we fix or convert the Guardian report instead? If that is not immediately possible, please run a pinned SpotBugs Maven plugin during `verify` and make findings fail the build. Any Guardian exception should be narrow and linked to a tracked follow-up.

sdk_v2/java/src/test/java/com/microsoft/foundry/local/NativeAsrTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+/** Explicit opt-in; never downloads a model. All sessions share the caller's prepared cache. */
+class NativeAsrTest {
+    @TempDir Path temporary;
+
+    @Test void realAsrOwnershipCancellationAndRepeatedSessions() throws Exception {
+        String runtime = setting("foundry.test.runtime", "FOUNDRY_LOCAL_NATIVE_BIN_DIR");
+        String cache = setting("foundry.test.cache", "FOUNDRY_TEST_DATA_DIR");
+        String wav = setting("foundry.test.wav", "FOUNDRY_TEST_WAV");
+        assumeTrue(
```
baijumeswani, line 21: In CI, missing native inputs should fail this test rather than skip it. A renamed or missing environment variable currently makes the job green even though the only real native ABI/model test did not run. Please add a CI-required property/profile and throw a clear assertion when any required input is missing. Local runs can continue to skip when that property is not set.

.pipelines/templates/stages-java.yml
```diff
@@ -0,0 +1,287 @@
...
+          $artifacts = @(
+            @{
+              Source = Join-Path $target "foundry-local-sdk-$version.jar"
+              Name = "foundry-local-sdk-$version.jar"
+            },
+            @{
+              Source = Join-Path $target "foundry-local-sdk-$version-sources.jar"
+              Name = "foundry-local-sdk-$version-sources.jar"
+            },
+            @{ Source = Join-Path $javaRoot '.flattened-pom.xml'; Name = "foundry-local-sdk-$version.pom" }
```
baijumeswani, line 254: Please add a packaged-consumer test after staging these files. Install this staged POM and JAR into an isolated Maven repository, then compile a small clean project that depends only on `com.microsoft.foundry:foundry-local-sdk:<version>`. That will catch a broken flattened POM, missing JNA dependency, wrong version, or incomplete JAR that source-tree tests cannot detect.

sdk_v2/java/src/main/java/com/microsoft/foundry/local/FoundryLocalManager.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public FoundryLocalManager(Configuration configuration) {
+        NativeApi.outsideCallback();
+        synchronized (FoundryLocalManager.class) {
+            if (active) throw new IllegalStateException("Only one FoundryLocalManager may be open per JVM");
+            api = NativeApi.load(configuration.runtimeDirectory());
+            Pointer config = api.create(api.config, NativeApi.ConfigurationApi.CREATE, configuration.appName());
+            try {
+                // Native status errors still propagate; keep the SDK quiet unless callers opt into logging.
+                api.check(api.config.pointer(NativeApi.ConfigurationApi.SET_DEFAULT_LOG_LEVEL, config, 5));
```
baijumeswani, line 30: Please make the log level and telemetry choice caller-configurable. Forcing Fatal-only logging makes native failures difficult to diagnose, and telemetry is also a product policy choice that the other SDKs expose through configuration. A good default is fine, but applications should be able to select a log level and opt in or out of nonessential telemetry.

sdk_v2/java/src/main/java/com/microsoft/foundry/local/Transcription.java
```diff
@@ -0,0 +1,522 @@
...
+    private long bufferedBytes;
+    private final long originNanos = System.nanoTime();
+    private long firstInputNanos, firstNonemptyNanos, inputClosedNanos, submittedBytes;
+
+    Transcription(AudioSession session, byte[] wav, PcmFormat format, Consumer<SpeechEvent> listener) {
+        this.session = session;
+        this.api = session.api;
+        this.format = format;
+        this.listener = listener;
+        worker = new Thread(this::run, "foundry-java-asr");
```
baijumeswani, line 43: This worker is non-daemon and can wait forever when an application forgets to finish or close a PCM stream. The live thread also keeps the transcription, session, model, manager, and native handles reachable, so the JVM cannot exit and GC cannot clean it up. Please make the worker and feeder daemon threads and/or add manager-level shutdown cleanup. A subprocess test should verify that an abandoned stream does not keep the JVM alive.

sdk_v2/java/src/main/java/com/microsoft/foundry/local/Catalog.java
```diff
@@ -0,0 +1,78 @@
...
+    private final Pointer handle;
+
+    Catalog(FoundryLocalManager owner, Pointer handle) { this.owner = owner; this.handle = handle; }
+
+    /**
+     * No alias fallback: the returned native identity must equal the requested name:version.
+     *
+     * @throws ModelNotFoundException if the exact ID is valid but unavailable
+     */
+    public Model getModel(String exactId) {
```
baijumeswani, line 23: This name differs from the other v2 SDKs: `getModel` normally accepts an alias, while exact `name:version` lookup is `getModelVariant`. Publishing the current name will make cross-language examples confusing and renaming it later will be a breaking change. Please align this surface now by adding `getModel(alias)` and `getModelVariant(exactId)`, and keep alias-level and flattened-variant listing methods clearly separate.

sdk_v2/java/pom.xml
```diff
@@ -0,0 +1,118 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) Microsoft Corporation. Licensed under the MIT License. -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.microsoft.foundry</groupId>
+  <artifactId>foundry-local-sdk</artifactId>
```
baijumeswani, line 8: These generic coordinates can look like the full Java equivalent of the other v2 SDKs, while this artifact intentionally supports only streaming ASR. Please make the preview status clear in the artifact/version (or use ASR-specific preview coordinates) until the public surface is ready to carry normal cross-language compatibility expectations.

sdk_v2/java/src/main/java/com/microsoft/foundry/local/NativeApi.java
jiec-msft: Fixed in 6735ff43 and fe0871e0. Java negotiates C API v2 before accessing function tables or version strings; the released 0.10.1 Windows runtime is cleanly rejected in a separate JVM (verified locally on JDK 17).

.pipelines/foundry-local-packaging.yml
jiec-msft: Fixed in 6735ff43. Maven verify runs pinned SpotBugs and fails on findings (zero in the local JDK 17 run). Only the unsupported Guardian SpotBugs report parser remains disabled; restoration is tracked in #1143.

sdk_v2/java/src/test/java/com/microsoft/foundry/local/NativeAsrTest.java
jiec-msft: Fixed in 6735ff43. CI sets foundry.test.native.required so missing runtime, model cache, or WAV fails instead of skipping. I confirmed the required test fails locally when inputs are absent; the remaining platform CI failures are still under investigation.

.pipelines/templates/stages-java.yml
jiec-msft: Fixed in 6735ff43. The package stage installs the staged flattened POM and JAR into an isolated Maven repository and compiles a clean consumer. I repeated that check successfully on JDK 17 locally; the CI packaging stage is still blocked by the failing platform jobs.

sdk_v2/java/src/main/java/com/microsoft/foundry/local/FoundryLocalManager.java
jiec-msft: Fixed in 6735ff43. The six-argument Configuration exposes caller-selected log level and nonessential telemetry, while the existing four-argument constructor retains the original Fatal and telemetry-disabled defaults.

sdk_v2/java/src/main/java/com/microsoft/foundry/local/Transcription.java
jiec-msft: Fixed in 6735ff43. The ASR worker and WAV feeder are daemon threads, manager close cleans up owned sessions, and the native integration test includes an abandoned-stream child-JVM exit check. Three Azure Java platform lanes have passed; the other two platform CI failures remain under investigation.

sdk_v2/java/src/main/java/com/microsoft/foundry/local/Catalog.java
jiec-msft: Fixed in 6735ff43. Catalog now exposes getModel(alias) and getModelVariant(name:version); models() returns aliases while modelVariants() expands variants. The native integration test exercises both lookups.

sdk_v2/java/pom.xml
jiec-msft: Fixed in 6735ff43. The Maven artifact is now foundry-local-asr-sdk and CI versions carry the -asr-preview suffix. The existing Preview 1 release was not changed.

.pipelines/foundry-local-packaging.yml
jiec-msft: Following up to make the current state explicit: this PR implements the fallback option from your comment.

- Maven `verify` runs a pinned SpotBugs Maven plugin (`com.github.spotbugs:spotbugs-maven-plugin:4.10.4.1`, `check` goal) with `failOnError=true`, so any finding fails the Java build on every platform lane. The latest local JDK 17 run reports zero findings.
- The exception is narrow: only the 1ES Guardian SpotBugs integration is disabled. The other SDL tools stay enabled, and the justification in the pipeline links to the tracked follow-up #1143.

Fixing or converting the Guardian report is not possible from this repository, because `gdn.unknownFormatResult` comes from Guardian's SpotBugs result handling. ONNX Runtime's Java package pipeline disables Guardian SpotBugs for the same error: [jar_package_testing.yml#L33-L35](https://github.com/microsoft/onnxruntime/blob/d811aa6bc95ae72e774fdee003aac078a98e737e/tools/ci_build/github/azure-pipelines/jar_package_testing.yml#L33-L35). If 1ES adds supported parsing or a conversion path, #1143 tracks restoring the Guardian scan and removing this exception.

sdk_v2/java/pom.xml
```diff
@@ -0,0 +1,136 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) Microsoft Corporation. Licensed under the MIT License. -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.microsoft.foundry</groupId>
+  <artifactId>foundry-local-asr-sdk</artifactId>
```
baijumeswani: This Maven coordinate makes ASR the permanent identity of the Java SDK. When chat or embeddings are added, users would either keep depending on an inaccurately named ASR artifact or have to migrate to a new coordinate.

Please use a general coordinate such as `com.microsoft.foundry:foundry-local-sdk`, and make the project name and description general as well. The README can clearly say that streaming ASR is the first supported scenario. Please also replace the `-asr-preview` version suffix and the matching hard-coded names in the packaging and consumer scripts with a general preview version.

sdk_v2/java/src/main/java/com/microsoft/foundry/local/Configuration.java
```diff
@@ -0,0 +1,29 @@
...
+import java.nio.file.Path;
+import java.util.Objects;
+
+/**
+ * Explicit locations only. Construction never loads native code or downloads assets.
+ *
+ * <p>The first runtime directory loaded is retained for the JVM lifetime. Recreated managers
+ * must use the same resolved runtime directory.
+ */
+public record Configuration(String appName, Path runtimeDirectory, Path modelCacheDirectory,
```
baijumeswani: Could this be an immutable class with a builder instead of a positional record? The native API and the other SDKs already support more general settings, including log directories, catalog URLs and filters, catalog region, external service settings, and additional key/value options.

Adding those fields to a public record changes its main constructor and makes future growth harder. A builder lets this preview expose the current settings now and add the remaining general settings later without repeatedly changing how callers construct `Configuration`.

sdk_v2/java/src/main/java/com/microsoft/foundry/local/ModelInfo.java
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. Licensed under the MIT License.
+package com.microsoft.foundry.local;
+
+/** Detached immutable metadata, valid even after the manager closes. */
+public record ModelInfo(String id, String alias, String name, int version, String uri,
```
baijumeswani: This record keeps a small, fixed metadata set and selects only two arbitrary properties (`license` and `licenseDescription`). General models also need prompt templates, model settings, device information, integer/string properties, and input/output metadata.

Please consider a library-created immutable class with named accessors and forward-compatible property maps. That avoids growing a public positional constructor every time another model capability is exposed.

sdk_v2/java/src/main/java/com/microsoft/foundry/local/FoundryLocalManager.java
```diff
@@ -0,0 +1,129 @@
...
+ * Owns the native singleton and every session. Catalogs/models are borrowed views.
+ *
+ * <p>Only one manager may be open at a time. After close, another manager can be created in
+ * the same JVM only with the same resolved runtime directory used by the first manager.
+ */
+public final class FoundryLocalManager implements AutoCloseable {
+    private static volatile FoundryLocalManager open;
+    private static boolean exitHookInstalled;
+    final NativeApi api;
+    final Set<AudioSession> sessions = new HashSet<>();
```
baijumeswani: The manager's ownership logic is tied directly to the first inference scenario. Adding chat or embeddings would require changing this set, the shutdown loop, and model-unload checks for every new session type.

Please track a small package-private lifecycle type instead, for example `OwnedSession extends AutoCloseable`. `AudioSession` can be its only implementation today. This keeps the public ASR API unchanged while making cleanup safe to extend.

sdk_v2/java/src/main/java/com/microsoft/foundry/local/FoundryLocalManager.java
```diff
@@ -0,0 +1,129 @@
...
+            manager.close();
+        } catch (RuntimeException | Error ignored) {
+            // Best effort: a close failure must not block process exit.
+        }
+    }
+
+    public String runtimeVersion() { return api.version; }
+    public String nativeTarget() { return NativeApi.target(); }
+
+    public Catalog catalog() {
```
baijumeswani: This exposes only the public catalog even though C API v2 supports public and local catalogs, and the other SDKs expose both. The local catalog is important for cached/local models and future BYOM support.

Please add a `CatalogType` and `catalog(CatalogType type)`, using `Manager_GetCatalogByType`. The current `catalog()` method can remain as a convenient shortcut for the public catalog.

sdk_v2/build_and_test_all.ps1
```diff
@@ -227,6 +229,33 @@ print(sys.executable)
...
     }
+
+    if ('java' -in $targets) {
+        Invoke-Step 'java' {
+            Push-Location $javaDir
+            try {
+                mvn --batch-mode --no-transfer-progress clean verify
+                if ($LASTEXITCODE -ne 0) { throw "mvn clean verify exit $LASTEXITCODE" }
+
+                $mainJar = @(Get-ChildItem target -Filter 'foundry-local-sdk-*.jar' -File |
```
baijumeswani: This documented build path currently fails after Maven succeeds because the POM produces `foundry-local-asr-sdk-*`, while these filters search for `foundry-local-sdk-*`.

Generalizing the Maven artifact name will fix the immediate mismatch. To prevent another mismatch later, please consider reading Maven's effective `project.build.finalName` instead of repeating the artifact name in this script.

.pipelines/templates/fetch-models-from-artifacts-feed.yml
```diff
@@ -8,27 +8,33 @@ parameters:
   type: string
   default: pscore
   values: ['pscore', 'ps']
+- name: modelSet
+  type: string
+  default: all
+  values: ['all', 'nemotron-speech']
```
baijumeswani: Selective model downloads are useful, but this shared pipeline contract is named after one model family. Replacing Nemotron, testing more than one speech model, or adding another Java feature would require changing the shared enum and every caller.

Please make the selection describe a capability, such as `streaming-asr`, or accept a list of required models. The model ID should be defined once and passed to the Java test. Keeping `all` as the default is good because it preserves existing callers.

sdk_v2/java/src/test/java/com/microsoft/foundry/local/NativeAsrTest.java
```diff
@@ -0,0 +1,143 @@
...
+            borrowed.download(completedDownload, value -> {
+                if (value == 100.0) completedDownload.cancel();
+            });
+            assertTrue(completedDownload.isCancelled(), "Cached download must report completion");
+            borrowed.load();
+            System.err.println("native-test: model loaded");
+            byte[] pcm = WavAudio.read(Path.of(wav)).pcm();
+            try (AudioSession session = borrowed.createAudioSession();
+                 Transcription run = session.transcribeWav(Path.of(wav), event -> callbacks.incrementAndGet())) {
+                assertFalse(run.await(Duration.ofSeconds(60)).text().isBlank());
```
baijumeswani: Checking only for nonblank text can pass when the binding returns truncated, stale, or incorrect output. The C++ test for this same recording already checks stable phrases such as `more than one link`, `live concert`, and `album to purchase`.

Please normalize case and punctuation, then check a few known phrases for both the WAV and streamed PCM results. Exact transcript equality is not needed. It would also be useful to add focused tests for the general catalog and model APIs instead of relying only on this ASR path.

baijumeswani (commented):
Thank you for addressing the earlier review comments. After clarifying the intended product direction, we reviewed the full PR again from four independent angles: product architecture, public API design, CI and packaging, and native correctness.

The key point is that this should be introduced as the general Foundry Local Java SDK. Streaming ASR can be the first supported inference scenario, but it should not become the permanent Maven identity or limit the foundation used by future chat, embeddings, image, and other features.

The native ASR implementation itself looks solid. API v2 negotiation, JNA layouts, ownership, callback lifetime, five-platform testing, pinned dependencies, and platform-independent packaging all look appropriate. The inline comments focus on the parts that should be made general or more reliable before publishing the preview.

One small documentation follow-up is also needed: the header in `.pipelines/foundry-local-packaging.yml` should list Java and the `java-sdk-v2` artifact.

